PR #2, opened 2023-02-21
feat(pack) add rust

luxus:
i guess it still needs some changes


luxus:
@uzaaft any changes that needs to be done?

luxus:
tested dap, crates and lsp.

PR #33, opened 2023-02-27
feat(motion): Add Mini.Basics

luxus:


mehalter (changes requested):
Reword commit to say `motion` as the category and not `movement`

mehalter (approved):
LGTM

PR #34, opened 2023-02-27
feat(motion): Add Mini.move

luxus:


mehalter (approved):
LGTM!

kaka-ruto:
Not sure why <A-hjkl> does nothing on my mac(even which-key doesn't trigger). Does this work for you if you're on a mac?

PR #37, opened 2023-02-28
chore(docs): update the README and add CONTRIBUTING guidelines

mehalter:


luxus:
a note that the ~/projects folder can be modified would be great
https://github.com/folke/lazy.nvim#%EF%B8%8F-configuration

luxus (approved):

PR #40, opened 2023-02-28
feat(colorscheme): Add Oxocarbon-nvim

luxus:


lua/astrocommunity/colorscheme/oxocarbon-nvim/README.md
```diff
@@ -0,0 +1,5 @@
+# oxocarbon
```
mehalter, line 1: I've been making this the name of the repo, so `oxocarbon.nvim` I think would be better here.

lua/astrocommunity/colorscheme/oxocarbon-nvim/oxocarbon-nvim.lua
```diff
@@ -0,0 +1,3 @@
+return {
+  "nyoom-engineering/oxocarbon.nvim",
+}
```
mehalter, line 3:
```suggestion
return { "nyoom-engineering/oxocarbon.nvim" }
```

This is a bit of a nitpick, but just to keep things minimal and clean, I think this is better

mehalter (changes requested):
Just a couple nitpicks :D

mehalter (approved):
LGTM!

PR #43, opened 2023-03-03
Added regexplainer

owittek:
Added a minimal configuration of Regexplainer. Unfortunately the Regexplainer Toggle does not work properly at the moment but it's nothing that I can fix on my side.

### DIFF
--- a/lua/astrocommunity/editing-support/nvim-regexplainer/README.md
+++ b/lua/astrocommunity/editing-support/nvim-regexplainer/README.md
@@ -1,0 +1,5 @@
+# nvim-regexplainer
+
+**Repository:** https://github.com/bennypowers/nvim-regexplainer
+
+Describes the regular expression under the cursor. Toggle it with `gR`
--- a/lua/astrocommunity/editing-support/nvim-regexplainer/nvim-regexplainer.lua
+++ b/lua/astrocommunity/editing-support/nvim-regexplainer/nvim-regexplainer.lua
@@ -1,0 +1,17 @@
+return {
+  { "bennypowers/nvim-regexplainer", opts = {}, ft = { "html", "javascript", "typescript" } },
+  -- Regex support
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      -- Ensure that opts.ensure_installed exists and is a table or string "all".
+      if not opts.ensure_installed then
+        opts.ensure_installed = {}
+      elseif opts.ensure_installed == "all" then
+        return
+      end
+      -- Add the "regex" parser to opts.ensure_installed.
+      table.insert(opts.ensure_installed, "regex")
+    end,
+  },
+}


mehalter (changes requested):
Looks great, only comment is the category this should be in is `editing-support` as found here: http://neovimcraft.com/plugin/bennypowers/nvim-regexplainer/index.html

lua/astrocommunity/editing-support/nvim-regexplainer/README.md
```diff
@@ -0,0 +1,5 @@
+# nvim-regexplainer
+
+**Repository:** https://github.com/bennypowers/nvim-regexplainer
+
+Describes the regular expression under the cursor. Enable it with ```<leader>ur```
```
luxus, line 5: please use Toggle instead of Enable

lua/astrocommunity/editing-support/nvim-regexplainer/nvim-regexplainer.lua
```diff
@@ -0,0 +1,9 @@
+return {
+  "bennypowers/nvim-regexplainer",
+  opts = {
+    auto = true,
+  },
+  keys = {
+    { "<leader>ur", "<cmd>RegexplainerToggle<cr>", desc = "Toggle Regexplainer" },
```
luxus, line 7: you could use the `mappings = {}` instead of keys.

lua/astrocommunity/editing-support/nvim-regexplainer/nvim-regexplainer.lua
```diff
@@ -0,0 +1,9 @@
+return {
+  "bennypowers/nvim-regexplainer",
+  opts = {
```
luxus, line 3: does the plugin start at all? 
you can use `ft =` to lazy load it on specific filetypes

luxus (changes requested):

lua/astrocommunity/editing-support/nvim-regexplainer/nvim-regexplainer.lua
mehalter: I think this is wrong @luxus . The lazy.nvim syntax is `keys`

lua/astrocommunity/editing-support/nvim-regexplainer/nvim-regexplainer.lua
mehalter: It starts when the key is pressed  because it's lazy loaded on `keys`

lua/astrocommunity/editing-support/nvim-regexplainer/README.md
mehalter: fixed

lua/astrocommunity/editing-support/nvim-regexplainer/nvim-regexplainer.lua
luxus: imho it should run without activation

luxus (approved):

PR #47, opened 2023-03-04
fix(pack): using table.insert instead of list_extend on python dap

luxus:
if only 1 entry is added table.insert is faster then list_extend


mehalter (approved):
LGTM

utkarshgupta137:
Sorry, I'm new to Lua 😅.

PR #48, opened 2023-03-05
fix: update to new user event based file lazy loading

mehalter:
Updates to the latest changes in AstroNvim v3 from https://github.com/AstroNvim/AstroNvim/commit/41db48682757dc06b7ea07ab4e29664601a8d4e0


A-Lamia (approved):
Everything seems top notch. 👍

PR #56, opened 2023-03-09
feat(pack): add lua language pack

mehalter:


lua/astrocommunity/pack/lua/lua.lua
```diff
@@ -0,0 +1,30 @@
...
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      -- Ensure that opts.ensure_installed exists and is a table or string "all".
+      if not opts.ensure_installed then
+        opts.ensure_installed = {}
+      elseif opts.ensure_installed == "all" then
+        return
+      end
+      table.insert(opts.ensure_installed, "lua")
```
luxus, line 11: Can we add "luap" as well?

luxus (changes requested):

lua/astrocommunity/pack/lua/lua.lua
mehalter: Added in latest commit revision!

luxus (approved):

PR #60, opened 2023-03-09
refactor: add utilities to easily insert/extend lists with unique values

mehalter:
This utility function fixes some issues when it comes to using multiple packs at the same time that might provide the same language servers/linters/formatters/etc.


20k-ultra:
can you update PR description with an example of before and after to show how this works for noobs like myself

mehalter:
> can you update PR description with an example of before and after to show how this works for noobs like myself

This doesn't change anything on the user side :) it's just an internal thing

PR #68, opened 2023-03-11
add more plugins

luxus:


mehalter (approved):
LGTM

lua/astrocommunity/terminal-integration/flatten-nvim/flatten-nvim.lua
```diff
@@ -0,0 +1 @@
+return { "willothy/flatten.nvim", opts = {} }
```
utkarshgupta137, line 1: Shouldn't there be a condition for loading this plugin?

PR #71, opened 2023-03-12
fix(mini-ai): explicitly call setup()

pkazmier:
The mini-ai astrocommunity plug-in was not working because the setup function was not being called.  According to lazy docs, it seems to imply that it should automatically call setup, but this plug-in did not work without explicitly defining our own config that calls setup.


mehalter (approved):
Thanks for catching this!

mehalter:
I made a small revision to this commit after the merge @pkazmier just to simplify the approach a bit. Thanks again for the fix!

PR #73, opened 2023-03-13
fix: sentence structure

owittek:
Just swapped around the words to fix a grammar issue.


mehalter (changes requested):
Could you please reword your commit to add the section, rather than `fix:` it should be `fix(indent):`

owittek:
> Could you please reword your commit to add the section, rather than `fix:` it should be `fix(indent):`

Learning something new every day :)

mehalter (approved):
LGTM!

PR #82, opened 2023-03-15
feat(pack/typescript): handle file change imports

taskylizard:
Handles TypeScript imports when renaming or moving a file using neo-tree.


mehalter (changes requested):
This is an awesome awesome addition!! Could you make sure that you are using `stylua` here to format the code to make the tests pass? The linting also failed but that looks like it's just a github actions problem haha

taskylizard:
Oops, didn't mean that, but should be fixed now.

mehalter (approved):
Perfect! Thanks again!

PR #91, opened 2023-03-20
fix(pack/rust): Ensure that taplo is only initialized once.

### LINKED ISSUE #90: fix(pack/rust): Taplo initialized twice in the rust pack
Hi there,

Making this issue as a reference for a pull request. Taplo is being initialized twice as it is being initialized elsewhere.

This should be added to astronvim.lsp.skip_setup.

sjcobb2022:
Closes #90

Ensure that taplo is only initialized once.

Just added it to astronvim.lsp.skip_setup.

### DIFF
--- a/lua/astrocommunity/pack/rust/rust.lua
+++ b/lua/astrocommunity/pack/rust/rust.lua
@@ -16,7 +16,7 @@ return {
   {
     "simrat39/rust-tools.nvim",
     ft = { "rust" },
-    init = function() utils.list_insert_unique(astronvim.lsp.skip_setup, "rust_analyzer") end,
+    init = function() utils.list_insert_unique(astronvim.lsp.skip_setup, { "rust_analyzer", "taplo" }) end,
     opts = function() return { server = require("astronvim.utils.lsp").config "rust_analyzer" } end,
     dependencies = {
       {


mehalter (changes requested):
@sjcobb2022 I went and fixed the git history on your branch. One thing I should point out is that this is actually not a real "fix". Taplo here is not being initialized twice as a language server. It is having the language server initialize and it is also having the `null-ls` sources initialize through `mason-null-ls`. This is disabling the language server and just letting `null-ls` set it up for linting. This is *not* the ideal behavior as we probably want the reverse of this. Disabling the null-ls and only using the LSP. I'm going to discuss with the `mason-null-ls` dev to see the best way to move forward with this.

sjcobb2022:
> @sjcobb2022 I went and fixed the git history on your branch. One thing I should point out is that this is actually not a real "fix". Taplo here is not being initialized twice as a language server. It is having the language server initialize and it is also having the `null-ls` sources initialize through `mason-null-ls`. This is disabling the language server and just letting `null-ls` set it up for linting. This is _not_ the ideal behavior as we probably want the reverse of this. Disabling the null-ls and only using the LSP. I'm going to discuss with the `mason-null-ls` dev to see the best way to move forward with this.
> 
> _EDIT_: just to clarify, there are no real requests for changes here, just blocking the commit for now until we clear this up

Alright alright, that sounds good. Feel free to close the pull if you want to move in another direction!

mehalter:
@sjcobb2022 actually, I had a different idea of how we could structure these packs in a more semantically meaningful way, check out #94 . This will make it easier for us to do fixes like this and fix it in all the places where `toml` language is used like rust + julia. So I am going to go ahead and close this and we can make a new PR that just fixes the `toml` language pack when that PR comes in. Thanks for catching this, opening an issue, and investigating!!

PR #93, opened 2023-03-22
Add EdgeDB syntax plugin

divan:
This PR adds new syntax support for [EdgeDB](https://www.edgedb.com) syntax files.

I wasn't sure what directory to put it, and went with `pack/edgedb`. Right now it's not really a pack, but just a syntax hightlight support, but that can change in a future.


mehalter (approved):
Thanks! This looks great and I like putting it in the pack folder to make it more extensible in the future!

PR #95, opened 2023-03-22
fix(leap): add vim-repeat to support dot-repeat

utkarshgupta137:
As per the README: https://github.com/ggandor/leap.nvim#dependencies


mehalter (approved):
LGTM! Thanks for catching this!

PR #106, opened 2023-03-28
feat(pack): switch to codelldb for rust debugging

### LINKED ISSUE #84: enhancement:

ZackGlenn:
Sets up rust-tools to debug with codelldb from Mason instead of lldb-vscode from the system.

Closes #84.

### DIFF
--- a/lua/astrocommunity/pack/rust/rust.lua
+++ b/lua/astrocommunity/pack/rust/rust.lua
@@ -18,7 +18,22 @@ return {
     "simrat39/rust-tools.nvim",
     ft = { "rust" },
     init = function() utils.list_insert_unique(astronvim.lsp.skip_setup, "rust_analyzer") end,
-    opts = function() return { server = require("astronvim.utils.lsp").config "rust_analyzer" } end,
+    opts = function()
+      local extension_path = vim.fn.stdpath "data" .. "/mason/packages/codelldb/extension/"
+      local codelldb_path = extension_path .. "adapter/codelldb"
+      local liblldb_path = extension_path .. "lldb/lib/liblldb"
+
+      if vim.loop.os_uname().sysname == "Linux" then
+        liblldb_path = liblldb_path .. ".so"
+      else
+        liblldb_path = liblldb_path .. ".dylib"
+      end
+
+      return {
+        server = require("astronvim.utils.lsp").config "rust_analyzer",
+        dap = { adapter = require("rust-tools.dap").get_codelldb_adapter(codelldb_path, liblldb_path) },
+      }
+    end,
     dependencies = {
       {
         "jay-babu/mason-nvim-dap.nvim",
--- a/lua/astrocommunity/pack/rust/rust.lua
+++ b/lua/astrocommunity/pack/rust/rust.lua
@@ -19,9 +19,9 @@ return {
     ft = { "rust" },
     init = function() utils.list_insert_unique(astronvim.lsp.skip_setup, "rust_analyzer") end,
     opts = function()
-      local extension_path = vim.fn.stdpath "data" .. "/mason/packages/codelldb/extension/"
-      local codelldb_path = extension_path .. "adapter/codelldb"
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/pack/rust/rust.lua
```diff
@@ -18,7 +18,22 @@ return {
     "simrat39/rust-tools.nvim",
     ft = { "rust" },
     init = function() utils.list_insert_unique(astronvim.lsp.skip_setup, "rust_analyzer") end,
-    opts = function() return { server = require("astronvim.utils.lsp").config "rust_analyzer" } end,
+    opts = function()
+      local extension_path = vim.fn.stdpath "data" .. "/mason/packages/codelldb/extension/"
```
mehalter, line 22: This should use the mason API for getting the package installation location rather than hard coding a path.

mehalter (changes requested):
This is awesome! One comment that needs to be fixed first

lua/astrocommunity/pack/rust/rust.lua
mehalter: Here is the mason API documentation: https://github.com/williamboman/mason.nvim/blob/main/doc/reference.md

ZackGlenn: That should be done now

sjcobb2022:
Just out of stylistic preference I would say that

```lua
local package_path = require("mason-registry").get_package("codelldb"):get_install_path()
```
Is "better practice" than accessing the actual index.

Not sure if anyone else has any thoughts on this. (@mehalter ??)

lua/astrocommunity/pack/rust/rust.lua
```diff
@@ -18,7 +18,22 @@ return {
     "simrat39/rust-tools.nvim",
     ft = { "rust" },
     init = function() utils.list_insert_unique(astronvim.lsp.skip_setup, "rust_analyzer") end,
-    opts = function() return { server = require("astronvim.utils.lsp").config "rust_analyzer" } end,
+    opts = function()
+      local package_path = require("mason-registry.index.codelldb"):get_install_path()
```
mehalter, line 22:
```suggestion
      local package_path = require("mason-registry").get_package("codelldb"):get_install_path()
```

mehalter (changes requested):
Small Change regarding the mason API

lua/astrocommunity/pack/rust/rust.lua
```diff
@@ -18,7 +18,22 @@ return {
...
+      local codelldb_path = package_path .. "/codelldb"
+      local liblldb_path = package_path .. "/extension/lldb/lib/liblldb"
+
+      if vim.loop.os_uname().sysname == "Linux" then
+        liblldb_path = liblldb_path .. ".so"
+      else
+        liblldb_path = liblldb_path .. ".dylib"
+      end
```
owittek, line 30: Personally I prefer ternary operations for such a simple if-statement but that might be controversial so feel free to leave it as it is.

```suggestion
      liblldb_path = liblldb_path .. (vim.loop.os_uname().sysname == "Linux" and ".so" or ".dylib"))
```

owittek (approved):
Looks great!

mehalter:
Github actions are currently down so I'm just going to merge this anyway.

dsisnero:
How should this work on windows?  - do we want
 `'extension/lldb/bin/liblldb.dll'  `

or 

`'extension\adapter\codelldb.dll'`

for liblldb_path

ZackGlenn:
> How should this work on windows? - do we want `'extension/lldb/bin/liblldb.dll' `
> 
> or
> 
> `'extension\adapter\codelldb.dll'`
> 
> for liblldb_path

I believe you would want `extension\lldb\bin\liblldb.dll`, but it's a moot point because astronvim disables nvim-dap on windows because mason-nvim-dap doesn't support it.

dsisnero:
it does support windows - I use it on windows

PR #107, opened 2023-03-29
fix(pack): improve compatibility of table.unpack function

abhijeetsrivastava:
- found this fix from https://github.com/hrsh7th/nvim-cmp/issues/1017
- this fix is working fine in my local

### DIFF
--- a/lua/astrocommunity/pack/java/java.lua
+++ b/lua/astrocommunity/pack/java/java.lua
@@ -110,7 +110,7 @@ return {
                 .. "/extension/server/com.microsoft.java.debug.plugin-*.jar"
             ),
             -- unpack remaining bundles
-            table.unpack(
+            (table.unpack or unpack)(
               vim.split(
                 vim.fn.glob(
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/pack/java/java.lua
```diff
@@ -1,5 +1,8 @@
 local utils = require "astrocommunity.utils"
 
+-- 5.1 lua compatibility
+table.unpack = table.unpack or unpack
+
```
mehalter, line 5:
```suggestion
```

This causes Linting issues and we shouldn't override this function globally

mehalter (changes requested):
We shouldn't override this function nlike this, we should change where it is called to just `(table.unpack or unpack)(...)`

lua/astrocommunity/pack/java/java.lua
abhijeetsrivastava: makes sense. I will update

mehalter (approved):
Great! Thanks for this!

PR #110, opened 2023-03-30
feat(motion): add nvim-surround

Cretezy:


lua/astrocommunity/motion/nvim-surround/nvim-surround.lua
```diff
@@ -0,0 +1,8 @@
+return {
+  {
+    "kylechui/nvim-surround",
+    version = "*", -- Use for stability; omit to use `main` branch for the latest features
+    event = "VeryLazy",
+    config = function(_, opts) require("nvim-surround").setup(opts) end,
```
mehalter, line 6: We should probably leverage lazy's syntax here and just to the `opts` notation 

```suggestion
    opts = {}
```

mehalter (changes requested):
Made small comment about syntax

mehalter (approved):

lua/astrocommunity/motion/nvim-surround/nvim-surround.lua
Cretezy: Sounds good. There's many plugins which also use the same `config`, so probably can clean those up too.

PR #113, opened 2023-04-02
feat(markdown-and-latex): add vimtex

bwkam:
disabling `lazy` by default makes it load when editing a .tex file


mehalter (changes requested):
Please use `stylua` to make the style check pass. You can ignore the conventional commit check. We can just use the pull request title and squash/merge

bwkam:
Okay

mehalter (approved):
Thanks! LGTM!

PR #115, opened 2023-04-04
feat(scrolling): add cinnamon.nvim plugin

rozhelluk:


mehalter (approved):
Thanks for this addition! I made a few small changes to the event and setup style, but other than that it looks perfect!

PR #118, opened 2023-04-04
feat(editing-support): add auto-save.nvim plugin

rozhelluk:


lua/astrocommunity/editing-support/auto-save-nvim/auto-save-nvim.lua
```diff
@@ -0,0 +1,5 @@
+return {
+  "Pocco81/auto-save.nvim",
+  lazy = false,
```
luxus, line 3: why did you turn off lazy?

lua/astrocommunity/editing-support/auto-save-nvim/auto-save-nvim.lua
```diff
@@ -0,0 +1,5 @@
...
+  config = function() require("auto-save").setup {} end,
```
luxus, line 4: any reason why you don't use `opts = {}`

luxus (changes requested):

lua/astrocommunity/editing-support/auto-save-nvim/auto-save-nvim.lua
mehalter: Moved to suggestion type comment.

mehalter (changes requested):
Made a few suggestions

mehalter (approved):
Just to expedite this getting merged, I went ahead and applied the changes. Thanks again for these contributions! <3

PR #122, opened 2023-04-06
feat(editing-support): Add text-case.nvim

Uzaaft:
Title pretty much says it all.
Couple of thoughts before I mark this as a non-draft.
Should be bind this to a key?
And, the event, perhaps we instead should load it on CMD instead of on a real file?


mehalter (approved):
Looks great!

PR #124, opened 2023-04-06
feat(color): add twilight.nvim

taskylizard:


lua/astrocommunity/color/twilight-nvim/twilight-nvim.lua
```diff
@@ -0,0 +1,4 @@
+return {
+  { "folke/twilight.nvim", cmd = "Twilight" },
+  { import = "astrocommunity.editing-support.zen-mode-nvim" },
+}
```
mehalter, line 4:
```suggestion
return { "folke/twilight.nvim", cmd = "Twilight" }
```


Does twilight have a strict requirement for `zen-mode`? seems like it just pairs nicely together and people who use twilight don't necessarily want zen-mode

mehalter (changes requested):

lua/astrocommunity/color/twilight-nvim/twilight-nvim.lua
taskylizard: Nope it doesn't, thought it was a nice addition but that's fine too.

mehalter (approved):
LGTM!

PR #125, opened 2023-04-06
ci: adding gh action to semi-automate releases

jay-babu:


mehalter (approved):
This looks great!

mehalter:
I'll fix the permissions tomorrow for the action to make it actually work but this gets us most of the way there!

PR #129, opened 2023-04-08
feat(completion): improve copilot-cmp defaults 

parK-dev:
- Re-enable C-e to turn off cmp window to see overlapping hidden copilot suggestion without losing said suggestion
- Add C-x and C-z mappings to go to next and previous copilot suggestions

This suggestion comes from an issue I have been having where whenever copilot gives a multi-line suggestion the overlapping cmp window hides it. By re-enabling C-e we can quickly shutoff the window without losing the suggestion and then we can go through them using the new mappings I added.

Screenshot 2023-04-09 at 2 02 55


mehalter (approved):
Looks great! Thanks!

PR #132, opened 2023-04-09
feat(code-runner): Add sniprun

Uzaaft:
Added initial version of sniprun.

### DIFF
--- a/lua/astrocommunity/code-runner/sniprun/sniprun.lua
+++ b/lua/astrocommunity/code-runner/sniprun/sniprun.lua
@@ -1,0 +1,9 @@
+return {
+  {
+    "michaelb/sniprun",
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/code-runner/sniprun/sniprun.lua
```diff
@@ -0,0 +1,9 @@
+return {
+  {
+    "michaelb/sniprun",
+    keys = {},
```
luxus, line 4: no keys?

Uzaaft: Not now. Listening to @owittek for now

luxus (approved):

PR #138, opened 2023-04-13
(feat) Add octo

0xricksanchez:
First attempt to add `octo.nvim`. I added some (imo) usable mappings to AstroNvim as well.


Uzaaft:
I believe that we shouldn't provide mappings, that's something the end-user should add. Instead, they can be added to the readme and be added by the end-user if they'd want

0xricksanchez:
> I believe that we shouldn't provide mappings, that's something the end-user should add. Instead, they can be added to the readme and be added by the end-user if they'd want

Fine by me. I’ll get this changed tomorrow

luxus:
isnt the general prefix for git g?
and not sure about the readme and keymap stuff..
in general adding keymaps is fine. there are no rules for that in astrocommunity.

0xricksanchez:
> isnt the general prefix for git g?

That's what I was playing around with first. However, I found some functions are behind a very long series of keystrokes so getting, so getting rid of 1 level already improved things a bit but its far from ideal...

Uzaaft:
> isnt the general prefix for git g? and not sure about the readme and keymap stuff.. in general adding keymaps is fine. there are no rules for that in astrocommunity.

How would `<leader>G` work?

luxus:
not sure about this yet.. does someone else have inputs?

0xricksanchez:
`<leader>G` sounds better suited to me than the previous `<leader>O`. Generally, I'd still love to place the `octo` stuff inside the git functionality but didn't find a way to improve the keybinds properly yet

luxus:
@Uzaaft are you now fine with this

Uzaaft:
@luxus  Would it be possible to add conditional keymappings? I think many of the keymappings that are set here only works for the Ccto filetype.

luxus:
don't think this exists.. maybe something that loads configs in root folders? but i guess that's out of scope

Uzaaft:
Yeah IG. I guess this PR is good enough for an initial version, I'll talk with mehalter about the file spesific things.

mehalter (changes requested):
Please make sure style and linting checks pass.

mehalter (approved):
LGTM

PR #140, opened 2023-04-13
feat: add harpoon

0xricksanchez:
Unsure about how much the terminal navigation is used, I heavily used the file navigation and figured a first PR for this would be fair enough.


luxus (approved):

0xricksanchez:
@luxus fixed :)

luxus (approved):
Lgtm

lua/astrocommunity/motion/harpoon/harpoon.lua
```diff
@@ -0,0 +1,33 @@
+local prefix = "<leader><leader>"
```
jay-babu, line 1: why leader leader?

PR #142, opened 2023-04-15
feat: Add Pylint and make it work in a virtualenv

sgarcialaguna:
Pylint is a very popular linter, unfortunately it doesn't work out of the box if you are using a virtualenv (as you should be doing). This issue in mason.nvim describes why it doesn't work and how to apply a workaround: https://github.com/williamboman/mason.nvim/issues/668

### DIFF
--- a/lua/astrocommunity/pack/python/README.md
+++ b/lua/astrocommunity/pack/python/README.md
@@ -9,5 +9,6 @@ This plugin pack does the following:
 - Adds the following `null-ls` sources:
   - [black](https://pypi.org/project/black/)
   - [isort](https://pypi.org/project/isort/)
+  - [pylint](https://pypi.org/project/pylint/)
 - Adds `debugpy` for debugging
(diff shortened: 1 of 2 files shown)


luxus:
can someone with python skills say something about this?
i don't use it at the moment

Uzaaft:
I'm not sure the path bit is needed? When spawning nvim in a poetry shell, everything works fine with pylint and my virtual env.

Uzaaft:
> Does it recognize imports that are only available in your venv? Does it pick up your .pylintrc? Does it also work if you run `poetry run nvim .` from a regular shell?

Yes to all of them. 
I just run `poetry shell` and then `nvim`

sgarcialaguna-mms:
Hmm, definitely not working for me. I'm not the first one to have encountered that problem either.

mehalter (approved):
LGTM!

PR #147, opened 2023-04-16
feat(terminal-integration): add vim-tmux-yank & vim-tpipeline

jay-babu:


lua/astrocommunity/terminal-integration/vim-tpipeline/vim-tpipeline.lua
```diff
@@ -0,0 +1,4 @@
+return {
+  "vimpostor/vim-tpipeline",
+  lazy = false,
```
jay-babu, line 3: this plugin needs to load at start. can't be lazy

PR #150, opened 2023-04-16
feat(completion): added more mappings for copilot-lua-cmp

Cthutu:
It is very useful to accept Co-pilot suggestions one word at a time or even a whole line.  Copilot offers these features but they have not been exposed.  VS Code, by default uses Ctrl+Right to accept a word.  I've added similar vim-like controls:

* Ctrl+Right/L for accepting a word
* Ctrl+Down/J for accepting a line
* Ctrl+C to dismiss the suggestion

### DIFF
--- a/lua/astrocommunity/completion/copilot-lua-cmp/copilot-lua-cmp.lua
+++ b/lua/astrocommunity/completion/copilot-lua-cmp/copilot-lua-cmp.lua
@@ -38,6 +38,26 @@ return {
       opts.mapping["<C-z>"] = cmp.mapping(function()
         if copilot.is_visible() then copilot.prev() end
       end)
+      
+      opts.mapping["<C-right>"] = cmp.mapping(function()
+        if copilot.is_visible() then copilot.accept_word() end
+      end)
+
+      opts.mapping["<C-l>"] = cmp.mapping(function()
+        if copilot.is_visible() then copilot.accept_word() end
+      end)
+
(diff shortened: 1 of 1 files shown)


luxus:
as many people using copilot i waiting for some comments, i don't use this version i use the normal version because i like the cmp and tab

Cthutu:
I disagree.  The ability to use copilot's complete next word and complete line is vital to partially accepting its suggestions and makes the tool more productive.  The pack adds `ctrl-x/z` which I would argue is less useful in practice, so why not add the more important features.

This, of course, is a judgement call.

luxus:
everyone can add their own preferences. the main reason for this config is that the tab thing is working.
not sure if adding extra keybinding is important.. imho they are optional (i don't use the word or line thing at all, and i use copilot a lot)
so i still hope some people that actually using this variant of copilot can give some input,
if we would mimic vscode this would make more sense for me

mehalter (changes requested):
Please make sure formatting and linting is passing

justyn:
> imho they are optional (i don't use the word or line thing at all, and i use copilot a lot) so i still hope some people that actually using this variant of copilot can give some input, if we would mimic vscode this would make more sense for me

I use the copilot accept_line option a lot, and should probably use accept_word more too.

But I just set them in my own keymap.

mehalter (approved):
Fixed the formatting/linting. LGTM

Issue #66, opened 2023-03-11
More Plugins To Be Added


luxus:
```lua
  {
    "b0o/incline.nvim",
    event = "User AstroFile",
    opts = { hide = { focused_win = true } },
  },
  ```

```lua
  { "lvimuser/lsp-inlayhints.nvim", event = "LspAttach", config = true },
```
```lua
{ "dpayne/CodeGPT.nvim", cmd = "Chat" },
```
```lua
  { "mrjones2014/op.nvim", build = "make install", opts = {} },
```

   ```lua 
    {
    "lukas-reineke/headlines.nvim",
    dependencies = "nvim-treesitter/nvim-treesitter",
    ft = "markdown",
    opts = {},
  },
  ```
  ```lua
  {
    "nvim-zh/colorful-winsep.nvim",
    event = { "WinNew" },
    opts = {},
  }, 
 ```
some plugins maybe someone wanna add 
edit: i removed the plugins that got added

0xricksanchez:
@luxus There's already a minimal `portal` implementation (from yourself in #35 :D) and I added a `grapple` feat in #139, guess that can be deleted from here?

luxus:
Thanks for the reminder :)

luxus:
i guess the plugins left are not that interesting, so i will close this

PR #154, opened 2023-04-18
feat(lsp): add inc-rename

0xricksanchez:
Added this one a while back and saw that its mentioned in #66. 
Not using this plugin much anymore as all my LSPs support renaming nowadays.


luxus:
didn't know that this is covered by lsp by now
maybe this is not necessary at all?
does it rename the stuff live?

0xricksanchez:
For example, this is what rust_analyzer seems to offer me via `<leader>lr` (Rename current symbol).

image

This doesn't seem to work for cases where the attached LS doesn't offer this, so if I e.g. have a `.json` file with all these attached: `jsonls`, `jsonlint`, `jq` ... I don't get the option. 
So, This may still be a nice complementary plugin to have for some cases/languages… And yes, it offers a "live-preview" as in it highlights all occurrences and updates them live when you change the symbol name.

If we wanna add this, maybe changing this to `<leader>lR` might be a good default fit then

Uzaaft:
IncRename works hand in hand with the LSP guys.

Uzaaft:
So imho, if  you're using IncRename, you WANT it to override the <leader>lr binding.

0xricksanchez:
Then I guess the PR does exactly what we want?

Uzaaft:
Yes, LGTM 🚀

Uzaaft (approved):

luxus (approved):

luxus (changes requested):
can u please change the folder to lsp
see http://neovimcraft.com/plugin/smjonas/inc-rename.nvim/index.html

luxus (approved):

PR #161, opened 2023-04-24
feat(colorscheme): add monokai-pro

bernwo:
Adds the [`monokai-pro`](https://github.com/loctvl842/monokai-pro.nvim) colorscheme.

### DIFF
--- a/lua/astrocommunity/colorscheme/monokai-pro/README.md
+++ b/lua/astrocommunity/colorscheme/monokai-pro/README.md
@@ -4,12 +4,12 @@
 
 A professional theme for neovim. The following filters are available:
 
-- `pro`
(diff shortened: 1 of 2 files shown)


bernwo:
This PR is ready for review. @luxus or anyone

Uzaaft:
LGTM. Just a question: Why  not just use the default options for the setup? I.e just doing something like:

```lua
return {
    "loctvl842/monokai-pro.nvim", 
      opts = {}
```

mehalter (changes requested):
Please use the `opts` table rather than `config` function with a `setup` call

mehalter (approved):
moved to the `opts` notation. LGTM

PR #162, opened 2023-04-25
fix(nvim-ts-rainbow2): remove hack

utkarshgupta137:
Recently, I've noticed that my editor slows down over time. Turns out that this hack is causing the problem. Without it, there is no more slowdown.

### DIFF
--- a/lua/astrocommunity/editing-support/nvim-ts-rainbow2/nvim-ts-rainbow2.lua
+++ b/lua/astrocommunity/editing-support/nvim-ts-rainbow2/nvim-ts-rainbow2.lua
@@ -2,15 +2,6 @@ return {
   "nvim-treesitter/nvim-treesitter",
   dependencies = {
     "HiPhish/nvim-ts-rainbow2",
-    config = function()
-      -- HACK: https://github.com/p00f/nvim-ts-rainbow/issues/112#issuecomment-1310835936
-      vim.api.nvim_create_autocmd({ "BufWritePost", "FocusGained" }, {
-        callback = function()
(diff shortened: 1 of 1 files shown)


A-Lamia:
aaaah, so this is what causes the slow down i thought i was going insane.

mehalter:
Yeah the unfortunate thing is without this autocmd, the coloring just gets messed up over time 😂 it seems that rainbow parentheses are just never going to be stable in Neovim 🤔

A-Lamia:
not sure if it's a good idea letting people use this, maybe put a warning on the read me to let people know ?

utkarshgupta137:
> aaaah, so this is what causes the slow down i thought i was going insane.

Me too brother, me too. You've no idea how much time I spent debugging this 😭.

> So then is there still an issue with ranbow-ts that needs the hack ?

Yes.

> if there isn't a fix i might just disabled it.

AFAIK, `nvim-rainbow-ts` doesn't have an issue with this hack. It seems like this fork changed something recently which made this hack slow down the editor.
A better fix is to do it manually once in a while. I've [created](https://github.com/utkarshgupta137/nvim-astro/commit/446b6479c08bab487f21da6b7ae080417e224776) a custom command & a mapping & I run it every few minutes. Doing it manually seems to have better results too, since there is a delay between disable & enable.

A-Lamia:
maybe we should do it on save ?

utkarshgupta137:
@mehalter This issue might be plaguing unsuspecting users, can we release this on priority?

mehalter (approved):
LGTM

Issue #148, opened 2023-04-16
error with astrocommunity.pack.typescript


citosid:
I have this error with the typescript pack:

```text
Error executing luv callback:
.../.local/share/nvim/lazy/nvim-dap/lua/dap/session.lua:1187: adapter.port is required for server adapter
stack traceback:
	[C]: in function 'assert'
	.../.local/share/nvim/lazy/nvim-dap/lua/dap/session.lua:1187: in function <.../.local/share/nvim/lazy/nvim-dap/lua/dap/session.lua:1179>
```

I already deleted ~/.cache/nvim and ~/.local/nvim and performed a very clean installation and the same error happens.

The adapter is being executed:

![image](https://user-images.githubusercontent.com/201430/232329607-1e519c07-4909-4aec-9acc-1f6e17aa2e73.png)

Due to this, if I try again after executing the debug the first time, it fails with another error saying there is already another process running (which makes sense...)

Any ideas what could be wrong?

luxus:
please provide a `:checkhealth` in a gist

citosid:
@luxus  I did an installation from scratch installing only the `example_user` repo and only this pack and is the same error. These are the steps I followed:

```bash
git clone git@github.com:/AstroNvim/AstroNvim ~/.config/Test
git clone https://github.com/AstroNvim/user_example ~/.config/Test/lua/user
# Modify the file ~/.config/Test/lua/user/plugins/community.lua and add the typescript pack
NVIM_APPNAME=Test nvim
```

Open a very basic typescript file, add a break point, start the debugger, and the same error happens

citosid:
@luxus more information:

I decided to try and make work the main `nvim-dap-vscode-js` by itself, and found this: https://github.com/mxsdev/nvim-dap-vscode-js/issues/14#issuecomment-1514656823

Based on that, I think the issue in this pack is the same. It is trying to use `js-debug-adapter` which is the default adapter provided by Microsoft. However, I read somewhere the author of `nvim-dap-vscode-js` found the one provided by Microsoft is not really dap compatible, due to some things having to be asynchronous (I can't find the comment now, maybe @mxsdev can comment?).

So, the solution would be changing the pack to some configuration similar to the one I have in that other issue. However, I'm not sure if I'm the only one with this issue or if you (or @mehalter) as the original creators of the pack have more comments about it...

luxus:
@owittek @Uzaaft help

owittek:
[...]

PR #163, opened 2023-04-26
feat(pack): make dap work in typescript using vscode-js-debug

skriems:
Hi!

I recently moved from my nvim setup to AstroNvim and really appreciate the work! Awesome suite! I installed the typescript pack and couldn't bring DAP to work properly, basically facing the same issues as in #148.

This PR replicates my working DAP setup with vscode-js-debug in the new _lazy_ format. I hope everything is fine, I only had an issue with the `config` field for vscode-js-debug: I used `config = true` but got an error like:

```
lua module not found for vscode-js-debug. use config func instead...
```

Didn't really know what to do with that, so I did what the msg told me to do ;)

I added an `attach to port` config which I usually use while debugging a bunch of docker containers, which expose the inspector thing on different ports. I have not tested the other configs.

Resolves: #148

### DIFF
--- a/lua/astrocommunity/pack/typescript/typescript.lua
+++ b/lua/astrocommunity/pack/typescript/typescript.lua
@@ -70,7 +70,19 @@ return {
     dependencies = {
       {
         "mxsdev/nvim-dap-vscode-js",
-        opts = { debugger_cmd = { "js-debug-adapter" }, adapters = { "pwa-node" } },
+        dependencies = {
+          {
+            "microsoft/vscode-js-debug",
+            -- FIXME: lua module not found for vscode-js-debug. use config func instead...
+            config = function() end,
+            build = "npm install --legacy-peer-deps && npx gulp vsDebugServerBundle && mv dist out",
+          },
+        },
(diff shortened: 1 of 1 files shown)


mehalter:
@owittek and @Uzaaft have been working on revamping the typescript stuff and have done some significant restructuring. They just got these changes merged in and this is not out of date and would need rebasing. They are also planning on getting a DAP PR together in the coming week I believe. They should comment on whether or not we should get this rebased and merged in or if we should wait on what they are doing.

owittek:
I'll take a look at it tomorrow but from the looks of it I wouldn't rebase it in its current form since the installation of `js-debug-adapter` should be done with Mason instead of building it. The only other change to the functionality is the addition of ports which might be relevant in getting it to work.

Uzaaft:
I think we should wait. We really do not want to be dependent on `ts-node` unless we absolutely need to.

mehalter:
Ok I'm going to go ahead and close this @skriems . I'll be sure to ping you when the new PR is opened to get your feedback though!

PR #166, opened 2023-04-26
fix: readd <leader>O as a prefix for octo

0xricksanchez:
Sorry, this got lost somehow along the way. This was discussed in #138 as a better fit than `<leader>G` until a better integration has been found


Uzaaft (approved):
`<leader>O` is not defined by default, so this LGTM

Issue #167, opened 2023-04-27
Rust debugger not working on Windows


Xoffio:
Hi,

I enabled the rust community package `user/plugins/community.lua`:
```lua
return {
  "AstroNvim/astrocommunity",
  { import = "astrocommunity.pack.rust" },
}
```

added `user/plugins/nvim_dap.lua`: 
```lua 
return {
  {
    "mfussenegger/nvim-dap", -- for debbugging
    enabled = true,
  }
}
```

but when I try to run `RustDebuggables` I get the next error "ENOENT: no such file or directory"
![rustdebuggable_error](https://user-images.githubusercontent.com/38369407/234754321-8cddf7fa-8980-4f7a-9418-4267f4395425.jpg)

Something that I noticed is that the rust community package downloads the codelldb in:
```
codelldb_path = "/extension/adapter/codelldb.exe"
liblldb_paht = "/extension/lldb/lib/liblldb.lib"
```

Instead of:
```
codelldb_path = "/codelldb"
liblldb_paht = "/extension/lldb/lib/liblldb"
```

So I manually edited the `rust.lua` file to match the right path in windows. Now nvim is able to run codelldb but then I get this error `ECCONNREFUSED`:
![dap_error_01](https://user-images.githubusercontent.com/38369407/234756039-7c7deb6a-a4f9-45dd-985d-22d935ce8c70.jpg)

Is there anything I can do to make this work?
If so then I can contribute updating the rust pack to check if the system is running windows and if so to point to the right directory.

Thanks in advance!

Uzaaft:
From the docs:
<img width="799" alt="image" src="https://user-images.githubusercontent.com/48220549/235254816-e7f7bb22-e54f-4de0-a25d-407623a2c822.png">
Feel free to add checks to windows if you think they'll work. It might take some time to double check your work.

Uzaaft:
Additional  pics of the docs: 

<img width="865" alt="image" src="https://user-images.githubusercontent.com/48220549/235255817-83d786c9-ec15-4a18-bac7-54b743727aa6.png">

Xoffio:
>  Feel free to add checks to windows if you think they'll work. It might take some time to double check your work.

HI @Uzaaft, can you clarify what you mean by "adding checks" ?   

> Additional pics of the docs:
> 
> <img alt="image" width="865" src="https://user-images.githubusercontent.com/48220549/235255817-83d786c9-ec15-4a18-bac7-54b743727aa6.png">

[...]

PR #169, opened 2023-04-28
fix(pack): make codelldb point to the right path in windows

Xoffio:
Currently, if you are on Windows and you try to use the rust debugger you will get errors and you won't be able to use the debugger. #167 
In this PR I fix that issue by correcting the path when the plugin is run on windows.


mehalter (changes requested):
Please fix style with stylua

mehalter (approved):
LGTM

PR #171, opened 2023-04-28
fix(pack): upstream util function for typescript-deno

owittek:


mehalter:
I'll resolve this on the PR no worries, I'll finish that PR next week

PR #174, opened 2023-05-01
feat(editing-support): add cutlass-nvim

owittek:


lua/astrocommunity/editing-support/cutlass-nvim/cutlass-nvim.lua
```diff
@@ -0,0 +1,12 @@
+local utils = require "astronvim.utils"
+
+return {
+  "gbprod/cutlass.nvim",
+  event = { "User AstroFile" },
+  opts = function(_, opts)
+    if utils.is_available "leap.nvim" then
+      if not opts.exclude then opts.exclude = {} end
+      utils.list_insert_unique(opts.exclude, { "ns", "nS" })
```
mehalter, line 9:
```suggestion
      opts.exclude = utils.list_insert_unique(opts.exclude, { "ns", "nS" })
```

mehalter (changes requested):
new `list_insert_unique` syntax

mehalter (changes requested):
gotta get them style points :eyes:

mehalter (approved):
LGTM

utkarshgupta137:
@owittek, is it working for you? Cutlass only works for me if I set `lazy = false`?

owittek:
> @owittek, is it working for you? Cutlass only works for me if I set `lazy = false`.

Yep it works for me. I've set the "User AstroFile" event to load the plugin which triggers as soon as you enter a buffer.

utkarshgupta137:
Maybe it's because I use `neovim-session-manager` to load currdir session on startup, but it doesn't work for me unless I add `lazy = false`.

sebastianst:
@owittek it doesn't work for me. The `d` commands still save the deleted text to the yank register, as if another plugin is restoring the changes to the `d` bindings by cutlass. I confirmed that the plugin actually loads (with `:Lazy`), and the `cut_key` option also works and binds e.g. `mm` as expected. I don't have much other custom configuration. Any ideas what could be wrong?

sebastianst:
I found the culprit: which-key. Disabling it makes cutlass work again.
See also https://github.com/gbprod/cutlass.nvim/issues/20 and https://github.com/folke/which-key.nvim/issues/513
However, the bug reporter there only has the issue that which-key doesn't pop up when pressing `d`. In my case, which-key does pop up, but cutlass just doesn't work at all. Seems like an ordering issue.

PR #182, opened 2023-05-03
refactor(pack): require neo-tree later in typescript

pipex:
This allows neo-tree to be disabled by overrriding the config and allow the plugin to keep working.

### DIFF
--- a/lua/astrocommunity/pack/typescript/typescript.lua
+++ b/lua/astrocommunity/pack/typescript/typescript.lua
@@ -1,5 +1,4 @@
 local utils = require "astronvim.utils"
-local events = require "neo-tree.events"
 
 local function on_file_remove(args)
   local ts_clients = vim.lsp.get_active_clients { name = "tsserver" }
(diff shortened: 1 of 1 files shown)


mehalter (commented):
Does this actually work the way you think it does? Have you tested it? This won't change the order in which things are loaded with `require` because of the structure of the file the `require` will be loaded immediately no matter the `opts` being overridden or not

pipex:
No, sorry, I have not yet tested this (I am still trying to figure out how to test from a fork). I’ll rework the PR to try to figure out how to do this.

mehalter:
@pipex you should be able to replace the `opts = { ... }` with this:

```lua
  opts = function(_, opts)
    local events = require "neo-tree.events"
    opts.event_handlers = {
      {
        event = events.FILE_MOVED,
        handler = on_file_remove,
      },
      {
        event = events.FILE_RENAMED,
        handler = on_file_remove,
      },
    }
  end,
```

to achieve what you want

mehalter (approved):
LGTM! Also for anyone that comes across this PR @pipex is the real deal. I used to work with this guy and certifiable genius!! ❤️

PR #189, opened 2023-05-05
Added Nord ColorScheme

sarvex:
feat: Added nord colorscheme with sensible defaults


sarvex:
Closed because existing PR exists

PR #190, opened 2023-05-07
fix(completion): use list_insert_unique in copilot-lua

owittek:


mehalter:
This doesn't make any sense. Please read the doc strings for what this function does. This is only relevant for list like tables. For something like this where everything is keyed, the merging of options works just fine.

PR #215, opened 2023-05-23
feat(pack): add typescript.nvim code actions

rafaelderolez:
This PR adds [typescript.nvim](https://github.com/jose-elias-alvarez/typescript.nvim#null-ls)'s code-actions to null-ls.


mehalter (approved):
LGTM

PR #216, opened 2023-05-24
fix(pack): gracefully rust plugin if codelldb is not available

Mic92:
We have this issue on NixOS but potentiall also other operating systems such as *BSDs and non-mainstream cpu architectures are affected by this.


mehalter (approved):
Thanks for the suggestion! I have gone ahead and improved the logic here a bit and fixed the formatting. Thanks for the contribution!

PR #223, opened 2023-05-25
feat(editing-support): add true-zen.nvim

jay-babu:


Uzaaft (approved):
LGTM🫡

lua/astrocommunity/editing-support/true-zen-nvim/true-zen-nvim.lua
```diff
@@ -0,0 +1,34 @@
+local utils = require "astronvim.utils"
+return {
+  "Pocco81/true-zen.nvim",
+  opts = {
+    integrations = {
+      tmux = os.getenv "TMUX" ~= nil, -- hide tmux status bar in (minimalist, ataraxis)
+      twilight = utils.is_available "twilight.nvim", -- enable twilight (ataraxis)
```
mehalter, line 7: I don't think this will work necessarily because `is_available` relies pretty heavily on lazy being done loading. I think this could be resolved by using the function notation for setting `opts`

mehalter (changes requested):
We should wrap the `opts` here in a function to fix race conditions in code

PR #224, opened 2023-05-25
Use yarn instead of npm for x6 installation speed.

Zeioth:
Using npm this plugin lags behind all the others during the installation process. Using yarn this is solved.


Uzaaft:
Please change your PR text to follow the semantic commit spec.

Uzaaft:
Also. I’m not a fan of this PR. Yarn is not included in node, but npm is. We should use the standard tool.

Uzaaft:
Any thoughts @AstroNvim/astrocommunity-maintainers about using yarn instead of npm? It might end up failing on people’s machines.

mehalter:
Yeah I agree that this should be `npm` for general applicability especially since that is what Mason is using already as well

PR #229, opened 2023-05-26
feat(markdown-and-latex): add markmap.nvim

Zeioth:


Uzaaft:
Please make sure that your PR title follows the semantic PR spec

mehalter (approved):
Made a few modifications to make it more generally accessible to AstroNvim users. Thanks for the contribution!

Zeioth:
@mehalter just one thing:

    npm install -g markmap-cli

will fail because -g require sudo permissions.

    yarn global add markmap-cli

Allow to install globally just for the user. That's why I use it. I don't think there is away around it.

mehalter:
Sorry about that! I just fixed it on `main`: https://github.com/AstroNvim/astrocommunity/commit/f001b6ad9606b06fd5a024969434387945cbeb9a

Uzaaft:
@mehalter  @Zeioth That's not correct.
npm install -g markmap-cli would fail only if you've setup node and npm uncorrectly. I think we should revert f001b6a

mehalter:
1. That is not necessarily the case @Uzaaft it will work if you install `npm` locally. If you use a system package manager to install it then `-g` will require `sudo` and default to the full system. I have looked into this and there is no better way to do this for the general use case. Maybe what we should do is remove the system installation and just say it requires `markmap` to be in your path like other plugins do. If someone wants to extend it to install with their installer of choice they can. What do you think of that @Zeioth 
2. That commit does more than just changing that line (file/folder renames) so we cannot just simply revert that commit.

Uzaaft:
> * That is not necessarily the case @Uzaaft it will work if you install `npm` locally. If you use a system package manager to install it then `-g` will require `sudo` and default to the full system. I have looked into this and there is no better way to do this for the general use case. Maybe what we should do is remove the system installation and just say it requires `markmap` to be in your path like other plugins do. If someone wants to extend it to install with their installer of choice they can. What do you think of that @Zeioth

I've installed npm with pacman on my arch machine, and there, I've had to setup npm correctly to setup non-sudo usage. See the official docs: https://docs.npmjs.com/resolving-eacces-permissions-errors-when-installing-packages-globally


Either way, npm recommends you to use node version manager in their docs, I think we should stick to npm, since most people are likely to have that.

Zeioth:
Thank you so much for going through the effort of checking that out guys. I've documented it [here](https://github.com/Zeioth/markmap.nvim/blob/main/README.md#troubleshooting). So whatever you decide will be fine.

jay-babu:
Can't mason just be used to manage the installation of this?

PR #231, opened 2023-05-26
fix(typescript): improve eslint and prettier loading

### LINKED ISSUE #225: Typescript pack always to configured to use eslint and prettier
With the current configuration of the typescript pack, every project with a `package.json` in the root, will configure `null-ls` to use `prettier` and `eslint`.
This will show an error to the user for every project that doesn't have one of the two installed and configured.

A better approach would be to check inside the `package.json` if the keys `eslintConfig` and `prettier` are present. 

https://github.com/AstroNvim/astrocommunity/blob/b88a6e68c68b5fc5b780c767d8c0ba8cd360198e/lua/astrocommunity/pack/typescript/typescript.lua#L41-L49

mehalter:
This seems like a pretty good way to improve the `prettier` and `eslint` loading. This does make it so that `eslint` is preferred over `prettier` if there are no configuration files for both specifically. This could be changed/improved based on suggestions probably from someone who works with these language servers/linters/formatters more than I do

Resolves #225

### DIFF
--- a/lua/astrocommunity/pack/typescript/typescript.lua
+++ b/lua/astrocommunity/pack/typescript/typescript.lua
@@ -36,25 +36,23 @@ return {
 
       if not opts.handlers then opts.handlers = {} end
 
+      local has_prettier = function(util)
+        return util.root_has_file ".prettierrc"
+          or util.root_has_file ".prettierrc.json"
+          or util.root_has_file ".prettierrc.js"
+      end
+      local has_eslint = function(util) return util.root_has_file ".eslintrc.json" or util.root_has_file ".eslintrc.js" end
+
       opts.handlers.prettierd = function()
         local null_ls = require "null-ls"
-        null_ls.register(null_ls.builtins.formatting.prettierd.with {
-          condition = function(util)
-            return util.root_has_file "package.json"
-              or util.root_has_file ".prettierrc"
-              or util.root_has_file ".prettierrc.json"
-              or util.root_has_file ".prettierrc.js"
-          end,
-        })
+        null_ls.register(null_ls.builtins.formatting.prettierd.with { condition = has_prettier })
       end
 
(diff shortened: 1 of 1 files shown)


Uzaaft:
Sensible defaults, prettier is opinionated so having it as opt in makes sense.

Uzaaft:
Just a thing: Potential edge case: `prettier` key in `package.json`. 
Should we try and catch this?

lua/astrocommunity/pack/typescript/typescript.lua
```diff
@@ -36,25 +36,23 @@ return {
 
       if not opts.handlers then opts.handlers = {} end
 
+      local has_prettier = function(util)
+        return util.root_has_file ".prettierrc"
+          or util.root_has_file ".prettierrc.json"
+          or util.root_has_file ".prettierrc.js"
+      end
+      local has_eslint = function(util) return util.root_has_file ".eslintrc.json" or util.root_has_file ".eslintrc.js" end
```
khaneliman, line 44: should probably look for all expected files from their documentation https://eslint.org/docs/latest/use/configure/configuration-files#configuration-file-formats

lua/astrocommunity/pack/typescript/typescript.lua
```diff
@@ -36,25 +36,23 @@ return {
 
       if not opts.handlers then opts.handlers = {} end
 
+      local has_prettier = function(util)
+        return util.root_has_file ".prettierrc"
+          or util.root_has_file ".prettierrc.json"
+          or util.root_has_file ".prettierrc.js"
```
khaneliman, line 42: similar to eslint https://prettier.io/docs/en/configuration.html

khaneliman (approved):

luxus (approved):
lgtm

luxus (approved):
lgtm

PR #238, opened 2023-05-28
feat(colorschemes)!: adding mellow, juliana, neosolarized and removing onigiri

luxus:


Uzaaft:
@luxus  I believe this PR will break some configs, so it should be: 
`add(colorschemes)!:`
cc @mehalter

luxus:
Because it removes onigiri? Dunno... if someone uses it :)

Uzaaft:
Yeah because it removes onigiri, someone might use it, we can't know. Even if it's highly likely that nobody uses it

mehalter:
Why was onigiri removed?

mehalter:
Can't we just avoid that breaking change and not remove it?

Uzaaft:
> Why was onigiri removed?

It's archived by the maintainer.

mehalter:
Ok, but it can still be installed. Doesn't seem too bad to just leave it to avoid a breaking change.

luxus:
i will do the changes later today

mehalter (changes requested):
Don't remove `onigiri` but we could put an archived notice on the `README`

mehalter (approved):
I went ahead and just resolved all of the issues here.

PR #242, opened 2023-05-28
feat(colorscheme): add 'vscode', 'onedarkpro', 'sonokai'

TheSast:


mehalter (changes requested):
it should be `colorscheme/ondedarkpro-nvim` as the folder name and `colorscheme/vscode-nvim` as the folder name. After those changes this looks great!

TheSast:
I saw some other colorscheme modules omitting the suffix so I did't add it (e.g. dracula.nvim, gruvbox.nvim, kanagawa.nvim, tokyonight.nvim, etc.).
If you reconfirm (a github message reaction would be enough) that you want to include the suffix I'll push the rename.

mehalter:
These should be fixed and are bugs that slipped through the cracks. These can be fixed without breaking changes as they are bugs. Could you also please open a PR that resolves this?

mehalter:
These are described in the contribution guidelines so it seems like some people aren't enforcing them @AstroNvim/astrocommunity-maintainers and some people are not reading them :/ Thanks for bringing this to my attention though! Luckily we can resolve this without breaking change even though it will be "breaking" to users since they are currently incorrect and against our methodology

mehalter (approved):
Looks great! If you could make a PR to fix the other inconsistencies with the suffixes that would be greatly appreciated! If not, I can make one in the next few days

TheSast:
ah, I was to slow, a "merged main into X" commit has slipped in there

mehalter:
Haha no worries @TheSast it wouldn't make it into our git log because this repo only does squash and merge

PR #243, opened 2023-05-28
fix(colorscheme): rename colorshemes to guidelines

TheSast:
Suggested in https://github.com/AstroNvim/astrocommunity/pull/242#pullrequestreview-1448288849

---
Unrelated: 
Shouldn't: README.md and CONTRIBUTING.md be in .github 🤔


mehalter:
They can be in either location for GitHub. In this case we put them front and center so people see them 😅

mehalter (approved):
Thanks for this!

nurseiit:
This change should probably be mentioned somewhere in the README. If I didn't run into this PR I wouldn't know I had to change `"astrocommunity.colorscheme.tokyonight"` to `"astrocommunity.colorscheme.tokyonight-nvim"` 🤦

TheSast:
@mehalter if these do get added to a README or a release note, we should probably fix all the plugins not following the convention, not just the colorschemes. 

Also, do you think the version number should be bumped up after everything is renamed?

mehalter:
@TheSast I do think all plugins get fixed. If you know of more please open a PR to fix them :)

mehalter:
@TheSast i updated the commit message to show breaking

benjaminmordaunt:
Awesome user experience! Break everyone's themes from community overnight!

mehalter:
Thanks @benjaminmordaunt ! ❤️

luxus:
> Awesome user experience! Break everyone's themes from community overnight!

I'm not a big fan of the big fan of the change, but nobody thought that astrocommunity gets so big..
the main issue is that keeping the foldername useful (name of the colorscheme) will have conflicts with themes with the same name etc.

mehalter:
Yeah I agree, also for people who rely on stability can track stable releases with semantic versioning to avoid random breaking changes on updates. Maybe we should just do a full refactor out to remove categories and go to a user/repo structure to completely avoid this issue in the future

luxus:
not sure it makes sense.. there are some plugins that are actually 2 and maybe not a pack.. and packs are out of that anyways.. 
and the folders now help to structure the imports in your community.lua
i guess if you would `dogfooding` you would understand :D

mehalter:
Yeah that's true, good ol' Chesterton's fence at play. It's a shame that there are so many conflicts, but the current approach is pretty good/decent. We just have to make sure it is followed to avoid changing our minds down the line like how we got to this point. Definitely tough problem to solve

PR #245, opened 2023-05-31
fix(java): use stable file locations for JARs provided by mason.nvim

williamboman:
Since the introduction of the new registry, packages started linking
more than executables (`bin/`) in Mason's root directory. Examples of
these are platform-agnostic "share" files, for example the JARs used
here.

Reaching into the internal package structures of Mason packages is not
stable and is prone to break across versions.


williamboman:
Note that java-debug-adapter started linking these files only [3 weeks ago](https://github.com/mason-org/mason-registry/pull/1546). Users who installed java-debug-adapter prior to that will not have files at these locations. I'm not sure what the policy is in terms of supported versions, but figured I'd open a PR for awareness anyway

mehalter (approved):
Thanks so much! This is great!

PR #248, opened 2023-05-31
feat(pack): clojure language pack

### LINKED ISSUE #247: feat: add clojure pack
Add a clojure pack to the Astrocommunity

* Treesitter parser: clojure
* LSP: clojure_lsp
* Plugins: "Olical/conjure" and "gpanders/nvim-parinfer"

#### Additional config
Conjure defined keybingings relative to the localleader to be mapped, usually the `,` key
An autocmd is used to set `;;` as the line comment character, to follow the Clojure style guide.

#### Draft clojure pack
A draft of what I think a Clojure pack would look like.  I am new to lua and astronvim, so code enhancements are most welcome.

```lua
-- Clojure support with Conjure plugin
return {
  -- Clojure Language Server
  {
    opts = function(_, opts) opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "clojure_lsp") end,
  },
  -- Clojure parser
  {
    opts = function(_, opts)
        opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "clojure")
      end
    end,
  },
  -- Parinfer parens management for Clojure
  {
    "gpanders/nvim-parinfer",
    ft = { "clojure" },
    config = function()
      vim.g.parinfer_force_balance = true
      vim.g.parinfer_comment_chars = ";;"
    end,
  },
  -- Conjure plugin for Clojure REPL
  {
    "Olical/conjure",
    -- load plugin on filetypes
    ft = { "clojure" },
    config = function()
      vim.g.maplocalleader = ","
      vim.g["conjure#log#hud#width"] = 1
      vim.g["conjure#log#hud#enabled"] = false
      vim.g["conjure#log#hud#anchor"] = "SE"
      vim.g["conjure#log#botright"] = true
      vim.g["conjure#extract#context_header_lines"] = 100
      vim.g["conjure#eval#comment_prefix"] = ";;"
      vim.g["conjure#client#clojure#nrepl#connection#auto_repl#enabled"] = false
      vim.g["conjure#client#clojure#nrepl#connection#auto_repl#hidden"] = true
      vim.g["conjure#client#clojure#nrepl#connection#auto_repl#cmd"] = nil
      vim.g["conjure#client#clojure#nrepl#eval#auto_require"] = false
      vim.g["conjure#client#clojure#nrepl#test#runner"] = "kaocha"

      vim.api.nvim_create_autocmd("FileType", {
        group = vim.api.nvim_create_augroup("comment_config", { clear = true }),
        pattern = { "clojure" },
        callback = function() vim.bo.commentstring = ";; %s" end,
        desc = "Lisp style line comment",
      })
    end,
  },
}
```

#### References
* https://github.com/Olical/conjure
* https://neovimcraft.com/plugin/Olical/conjure/index.html

practicalli-john:
Add a clojure pack to the Astrocommunity

- Treesitter parser: clojure
- LSP: clojure_lsp
- Plugins: "Olical/conjure" and "gpanders/nvim-parinfer"
- Add autocmd to disable lsp diagnostics in Conjure log
- Add autocmd to configure comments to Clojure style guide

#### Additional Config

Add a localleader mapping to the user config (if not defined in AstroNvim) to enable Conjure key mappings, e.g. `, e r` to evaluate top level root

Define a localleader mapping in `options.lua`

```lua
return {
  g = {
    mapleader = " ",                 -- sets vim.g.mapleader
    maplocalleader = ",",            -- Set local leader key binding
    -- additional options...
  },
}
```

An autocmd is used to set ;; as the line comment character, to follow the Clojure style guide.

Resolve #247

### DIFF
--- a/lua/astrocommunity/pack/clojure/clojure.lua
+++ b/lua/astrocommunity/pack/clojure/clojure.lua
@@ -1,0 +1,60 @@
+-- Clojure support with Conjure plugin
+local utils = require "astronvim.utils"
+return {
+  -- Clojure Language Server
+  {
+    "williamboman/mason-lspconfig.nvim",
+    opts = function(_, opts) opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "clojure_lsp") end,
+  },
+  -- Clojure parser
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      if opts.ensure_installed ~= "all" then
+        opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "clojure")
+      end
+    end,
+  },
+  -- Parinfer parens management for Clojure
+  {
+    "gpanders/nvim-parinfer",
+    ft = { "clojure" },
+    init = function()
+      vim.g.parinfer_force_balance = true
+      vim.g.parinfer_comment_chars = ";;"
+    end,
+  },
+  -- Conjure plugin for Clojure REPL
+  {
+    "Olical/conjure",
+    -- load plugin on filetypes
+    ft = { "clojure" },
+    init = function()
+      vim.g["conjure#log#hud#width"] = 1
+      vim.g["conjure#log#hud#enabled"] = false
+      vim.g["conjure#log#hud#anchor"] = "SE"
+      vim.g["conjure#log#botright"] = true
+      vim.g["conjure#extract#context_header_lines"] = 100
+      vim.g["conjure#eval#comment_prefix"] = ";;"
+      vim.g["conjure#client#clojure#nrepl#connection#auto_repl#enabled"] = false
+      vim.g["conjure#client#clojure#nrepl#connection#auto_repl#hidden"] = true
+      vim.g["conjure#client#clojure#nrepl#connection#auto_repl#cmd"] = nil
+      vim.g["conjure#client#clojure#nrepl#eval#auto_require"] = false
+      vim.g["conjure#client#clojure#nrepl#test#runner"] = "kaocha"
+
+      vim.api.nvim_create_autocmd("BufNewFile", {
+        group = vim.api.nvim_create_augroup("conjure_log_disable_lsp", { clear = true }),
+        pattern = { "conjure-log-*" },
+        callback = function() vim.diagnostic.disable(0) end,
+        desc = "Conjure Log disable LSP diagnostics",
+      })
+
+      vim.api.nvim_create_autocmd("FileType", {
+        group = vim.api.nvim_create_augroup("comment_config", { clear = true }),
+        pattern = { "clojure" },
+        callback = function() vim.bo.commentstring = ";; %s" end,
+        desc = "Lisp style line comment",
+      })
+    end,
+  },
+}
--- a/lua/astrocommunity/pack/clojure/README.md
+++ b/lua/astrocommunity/pack/clojure/README.md
@@ -1,0 +1,38 @@
+# Clojure Language Pack
+
+Requires:
+
+- Java Development Kit, e.g. `JDK-17`
+- [Clojure CLI](https://clojure.org/guides/install_clojure) or [Leiningen](https://leiningen.org/)
+
+This plugin pack does the following:
+
+- Adds `clojure` treesitter parsers
+- Adds `clojure_ls` language server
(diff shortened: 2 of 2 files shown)


lua/astrocommunity/pack/clojure/clojure.lua
```diff
@@ -0,0 +1,54 @@
...
+      if opts.ensure_installed ~= "all" then
+        opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "clojure")
+      end
+    end,
+  },
+  -- Parinfer parens management for Clojure
+  {
+    "gpanders/nvim-parinfer",
+    ft = { "clojure" },
+    config = function()
```
mehalter, line 22:
```suggestion
    init = function()
```

This should be an `init` function I'm pretty sure

lua/astrocommunity/pack/clojure/clojure.lua
```diff
@@ -0,0 +1,54 @@
...
+      vim.g.parinfer_force_balance = true
+      vim.g.parinfer_comment_chars = ";;"
+    end,
+  },
+  -- Conjure plugin for Clojure REPL
+  {
+    "Olical/conjure",
+    -- load plugin on filetypes
+    ft = { "clojure" },
+    config = function()
```
mehalter, line 32:
```suggestion
    init = function()
```

I believe this should be an init function

lua/astrocommunity/pack/clojure/clojure.lua
```diff
@@ -0,0 +1,54 @@
...
+      vim.g.maplocalleader = ","
```
mehalter, line 33:
```suggestion
```

We shouldn't be overriding `maplocalleader` in a plugin pack

mehalter (changes requested):
Thanks for this great contribution! Just a few easy edits for some small things I noticed that should get fixed.

lua/astrocommunity/pack/clojure/clojure.lua
practicalli-john: I could try use the Conjure Clojure mapping prefix rather than explicitly using maplocalleader

```
vim.g.conjure#mapping#prefix = ","
```

If that isn't an option, then the pack would not enable the key mappings.  If that were the case I would need to document in the readme how to activate mappings by setting a localeader or using the Clojure mapping prefix

practicalli-john: The conjure mapping prefix approach works, so an update has been pushed

lua/astrocommunity/pack/clojure/clojure.lua
practicalli-john: Updated and pushed

lua/astrocommunity/pack/clojure/clojure.lua
```diff
@@ -0,0 +1,54 @@
...
+      vim.g.parinfer_comment_chars = ";;"
+    end,
+  },
+  -- Conjure plugin for Clojure REPL
+  {
+    "Olical/conjure",
+    -- load plugin on filetypes
+    ft = { "clojure" },
+    init = function()
+      vim.g["conjure#mapping#prefix"] = ","
```
mehalter, line 33:
```suggestion
```

I think actually we should keep this as `localleader` I think the main thing is AstroNvim should provide a default `localleader`. This will make it easier for users to customize things still on their end. I don't mind just merging this in in the meantime but I think the next release of AstroNvim will just have `localleader` defined as probably `,`

mehalter (changes requested):

lua/astrocommunity/pack/clojure/clojure.lua
practicalli-john: Yes, I would prefer a localleader set in AstroNvim and/or via the AstroNvim user_example 

Before creating the Clojure pack I used the `options.lua` configuration file from the AstroNvim user_example template to set the localleader, using `g = { maplocalleader = "," }`

For the moment, I'll remove the conjure mapping prefix line and add a note to the readme stating that localleader should be set in the user config, if not already set in AstroNvim.

mehalter (approved):
This looks great! Thanks so much! Also just to let you know I have added a default `localleader` to AstroNvim nightly and it will come out with the next release. We are just waiting on a couple small things before we do the release.

PR #253, opened 2023-06-02
feat(pack): add selene linter to lua pack

owittek:
Personally I'm a big fan of `selene` so I think it should be part of the pack.


mehalter (approved):
LGTM

PR #255, opened 2023-06-03
fix!: rename plugins to follow convention

TheSast:
@TheSast I do think all plugins get fixed. If you know of more please open a PR to fix them :)

_Originally posted by @mehalter in https://github.com/AstroNvim/astrocommunity/issues/243#issuecomment-1569149228_

---

This should be all of them and hopefully the last breaking change for a bit.


mehalter (approved):
Bless, thank you so much!

PR #256, opened 2023-06-04
chore: update readme according to breaking renames

owittek:
The `README` should reflect the new colorscheme names.


mehalter (approved):
Thanks!

PR #258, opened 2023-06-04
feat(markdown-and-latex): install markmap from mason

jay-babu:


mehalter:
@jay-babu does mason-null-ls super ensuring the installation of this?

jay-babu:
> @jay-babu does mason-null-ls support ensuring the installation of this?

it is trick (or bug depending on your perspective) that I have no plans fixing :). mason-null-ls does not enforce any checks to ensure the requested tool is in `null-ls` registry

mehalter:
Ah I see 😂 I didn't know if it would require you to use a name that would be translated to a mason name. Cool thanks! I'll merge this!

mehalter (approved):
LGTM!

jay-babu:
> Ah I see 😂 I didn't know if it would require you to use a name that would be translated to a mason name. Cool thanks! I'll merge this!

Simply converts - to _ and back. Other characters are unaffected

PR #254, opened 2023-06-02
feat(pack): adds angular pack

axelcalixte:
Hello,

I noticed there wasn't any pack for Angular so I made one for myself.
Maybe it is useful to someone else.

PR #260, opened 2023-06-05
feat(pack): add html-css base pack

owittek:
As suggested for https://github.com/AstroNvim/astrocommunity/pull/254 I'm adding the html-css base pack.


mehalter (approved):
Looks great!

PR #264, opened 2023-06-06
feat(editing-support): add keymaps to refactoring-nvim

jay-babu:


lua/astrocommunity/editing-support/refactoring-nvim/refactoring-nvim.lua
```diff
@@ -3,4 +3,97 @@ return {
...
+      },
+      desc = "Extract Block To File",
+    },
+    {
+      "<leader>ri",
+      [[ <Cmd>lua require('refactoring').refactor('Inline Variable')<CR>]],
+      { silent = true, expr = false },
+      mode = { "n" },
+      desc = "Inline Variable",
+    },
```
mehalter, line 67:
```suggestion
    {
      "<leader>rb",
      function() require('refactoring').refactor('Extract Block') end,
      { silent = true, expr = false },
      mode = {
        "n",
      },
      desc = "Extract Block",
    },
    {
      "<leader>rbf",
      function() require('refactoring').refactor('Extract Block To File') end,
      { silent = true, expr = false },
      mode = {
        "n",
      },
      desc = "Extract Block To File",
    },
    {
      "<leader>ri",
      function() require('refactoring').refactor('Inline Variable') end,
      { silent = true, expr = false },
      mode = { "n" },
      desc = "Inline Variable",
    },
```

These should just be lua functions

mehalter (changes requested):
The visual mode mappings make sense to be `vimscript` strings, but I let a revision for a few that can be moved to just lua functions

lua/astrocommunity/editing-support/refactoring-nvim/refactoring-nvim.lua
jay-babu: Sure, I mostly just copied pasted

mehalter (approved):
I went ahead and fixed the formatting. Thanks!

jay-babu:
> I went ahead and fixed the formatting. Thanks!

Thanks. I can't do anything from my work computer

PR #269, opened 2023-06-08
fix(editing-support): return opts

jay-babu:


mehalter:
Thai isn't actually necessary with lazy

PR #270, opened 2023-06-09
feat(pack): add luacheck linter to lua pack

RayJameson:
Since `selene` got removed from lua pack in 4caa6da and pretty much every plugin developer using `luacheck`, I think it would be a good idea to add it as linter, especially because Astronvim uses it and `.luacheckrc` is already configured there, so it's not going to be a wall of warnings like `accessing undefined variable 'vim'`


luxus:
@owittek ?

owittek (approved):
Sounds like a good argument for it. LGTM

PR #274, opened 2023-06-13
fix(typescript): fix null_ls eslint_d registration(s)

rafaelderolez:
- Removed `or not has_prettier` from null_ls eslint_d diagnostics registration. Eslint should only be registered if `has_eslint` returns true. It should not be registered just because a project does not have a  prettier config.
- Added `null_ls.builtins.formatting.eslint_d`
- Added `null_ls.builtins.code_actions.eslint_d`
- Fixed `prettier.config.(c)js` filenames


Uzaaft (approved):
LGTM. Thanks for double checking my work! 🤣

PR #275, opened 2023-06-14
feat(editing-support): Add nvim-ts-autotag

Uzaaft:
Notes:
This plugin seems to always load no matter the filetype.


mehalter:
Do we need to add plugins that are included in core AstroNvim?

 https://github.com/AstroNvim/AstroNvim/blob/main/lua/plugins/treesitter.lua#L3

Uzaaft:
> Do we need to add plugins that are included in core AstroNvim?

Good point. 🤣

PR #281, opened 2023-06-17
feat(motion)!: separate flit.nvim from leap.nvim

owittek:
This change has the benefit that only one of those plugins is active instead of both reserving keybinds

### DIFF
--- a/lua/astrocommunity/motion/flit-nvim/README.md
+++ b/lua/astrocommunity/motion/flit-nvim/README.md
@@ -1,0 +1,5 @@
+# flit.nvim
+
+**Repository:** https://github.com/ggandor/flit.nvim
+
+f/F/t/T motions on steroids, building on the Leap interface.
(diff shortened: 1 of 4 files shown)


Uzaaft:
Sensible changes. But wouldn't this be a breaking change? @owittek

owittek:
> Sensible changes. But wouldn't this be a breaking change? @owittek

Yes that was my initial thought too but considering that `leap.nvim` is now actually just `leap.nvim` I think it might be fine to not actually declare this as breaking.

Uzaaft:
Hmm but it changes the previously expected behaviour, no?

owittek:
I guess so, let me change it to breaking then

luxus:
i guess we need a good commit message, dunno if its understandable that people need to add flit.nvim separately now

owittek:
> i guess we need a good commit message, dunno if its understandable that people need to add flit.nvim separately now 

I would just hope for people to look into the commit if a breaking change message pops up. Not sure how to make the message clearer without writing multiple sentences

mehalter:
I think the message is fine

TheSast:
Wouldn't it be preferrable if the flit spec depended on the astrocommunity's spec for leap instead of directly on leap? That would be more similar to how it was before.

luxus (approved):

PR #283, opened 2023-06-18
fix(ccc-nvim.lua): highlighter attached buffer error

khaneliman:
Fix attached buffer error and utilize the auto loading instead of calling it directly since we're passing in the auto_enable option. https://github.com/uga-rosa/ccc.nvim/commit/5a9e394e61925d71f9c1c0cd1fd30ce463d8be46

image

Did nothing to affect code style so not sure why the action is failing...

### DIFF
--- a/lua/astrocommunity/color/ccc-nvim/ccc-nvim.lua
+++ b/lua/astrocommunity/color/ccc-nvim/ccc-nvim.lua
@@ -5,9 +5,5 @@ return {
     event = "User AstroFile",
     keys = { { "<leader>uC", "<cmd>CccPick<cr>", desc = "Toggle colorizer" } },
     opts = { highlighter = { auto_enable = true } },
-    config = function(_, opts)
-      require("ccc").setup(opts)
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/color/ccc-nvim/ccc-nvim.lua
```diff
@@ -7,7 +7,6 @@ return {
     opts = { highlighter = { auto_enable = true } },
     config = function(_, opts)
```
luxus, line 8: can u try to remove the whole config part?

luxus (changes requested):

lua/astrocommunity/color/ccc-nvim/ccc-nvim.lua
khaneliman: looks like that worked, assuming just a error for pointless config section?

mehalter (approved):
Looks great!

mehalter:
This caused issues that are resolved in #356

PR #286, opened 2023-06-18
refactor(twilight): update lazy loading and add default keymap

khaneliman:
Opinions on including keymaps in these community plugins?


owittek:
That would have to be something that is being added to every theme (that supports toggling) imo. Not sure if there's a solution where this keybind could be reserved in AstroNvim itself but otherwise I think that this is more of a user-specific customization.

mehalter:
Just a small change that this isn't a breaking change so we shouldn't have a `!`

mehalter (approved):
LGTM

PR #288, opened 2023-06-19
chore(motion): add missing desc to leap binds

TheSast:
explicitly bind default mappings using 'key' for more intuitive overriding.


mehalter (approved):
LGTM

PR #292, opened 2023-06-20
feat(harpoon)!: remove "j" keymap, make "t" keymap dynamic based on $TMUX

RayJameson:
I believe it's better when you have to remember one less keybinding.
Hope nobody gets upset because of the removed "j" keymap.
I may add it for compatibility

### DIFF
--- a/lua/astrocommunity/motion/harpoon/init.lua
+++ b/lua/astrocommunity/motion/harpoon/init.lua
@@ -26,6 +26,7 @@ return {
       prefix .. "t",
       function()
         local num = tonumber(vim.fn.input "GoTo terminal window number: ")
+        if num == nil then return end
         require("harpoon.term").gotoTerminal(num)
       end,
       desc = "Goto to terminal window",
@@ -34,6 +35,7 @@ return {
       prefix .. "j",
       function()
         local num = tonumber(vim.fn.input "GoTo Tmux window number: ")
+        if num == nil then return end
         require("harpoon.tmux").gotoTerminal(num)
       end,
       desc = "Goto to TMUX tmux window",
--- a/lua/astrocommunity/motion/harpoon/init.lua
+++ b/lua/astrocommunity/motion/harpoon/init.lua
@@ -1,4 +1,13 @@
 local prefix = "<leader><leader>"
+local dynamic_tmux_keymap_desc
+local running_tmux_seesion = vim.fn.exists "$TMUX" == 1
+if running_tmux_seesion then
+  dynamic_tmux_keymap_desc = "Goto to TMUX window"
+else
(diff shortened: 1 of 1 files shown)


RayJameson:
randomly fixed the error that occurred if used canceled input when using the "go to terminal" thing, also added a new keymap to jump to the index of the mark

luxus:
anyone else that uses harpoon can confirm that?

RayJameson:
@luxus on second thought, maybe it's better to leave keymaps as they do not contradict each other. You can still make `:term` while using TMUX, maybe someone using it. And in my experience, the TMUX keymap doesn't work as intended, I checked the plugin repo there is at least one open issue:
- thePrimeagen/harpoon#180

I can remove remap part and leave everything else

mehalter (commented):
Hey @RayJameson just a heads up, I went ahead and rebased this on `main` because there was a bit of standardization that was done for filenames

owittek:
> This also fixes which key preview, by adding an empty function. Before it was `<prefix><prefix>`, now it's `󱡀 Harpoon`, however I didn't figure out how to make it blue color as it should be.

Could you please elaborate what the empty function fixes exactly? Having any functionality (like an anonymous function) leads to the which-key entry not being colored.

lua/astrocommunity/motion/harpoon/init.lua
```diff
@@ -7,27 +20,35 @@ return {
...
+    { "<C-p>", function() require("harpoon.ui").nav_prev() end, desc = "Goto previous mark" },
+    { "<C-n>", function() require("harpoon.ui").nav_next() end, desc = "Goto next mark" },
+    { prefix .. "m", "<cmd>Telescope harpoon marks<CR>", desc = "Show marks in Telescope" },
     {
-      prefix .. "j",
+      prefix .. "t",
       function()
-        local num = tonumber(vim.fn.input "GoTo Tmux window number: ")
-        require("harpoon.tmux").gotoTerminal(num)
+        if running_tmux_seesion then
```
mehalter, line 41:
```suggestion
        if running_tmux_session then
```

Propagate typo fix

mehalter (changes requested):
A few small comments about more easily setting variables and small typo fix

mehalter (approved):
LGTM!

mehalter:
@RayJameson I think the last thing before we merge, is do you think you could come up with a good way to word the fact that there are binding changes from this commit? I think that would take higher precedence than the dynamic changing of the text in the message

RayJameson:
@mehalter done

mehalter:
Thanks for your work on this @RayJameson ! ❤️

PR #295, opened 2023-06-21
feat(pack): add cpp base pack

doupongzeng:
Hello, 

I noticed that there wasn't a package for cpp, so I created one for myself. Perhaps it will be useful to someone else.


Uzaaft:
Looks good to me. 
@doupongzeng  Should we perhaps add stuff like: `p00f/clangd_extensions.nvim` and  `Civitasv/cmake-tools.nvim`?

lua/astrocommunity/pack/cpp/cpp.lua
```diff
@@ -0,0 +1,19 @@
...
+    end,
+  },
+  {
+    "williamboman/mason-lspconfig.nvim",
+    opts = function(_, opts) opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "clangd") end,
+  },
+  {
+    "jay-babu/mason-null-ls.nvim",
+    opts = function(_, opts) opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "clang-format") end,
+  },
```
Uzaaft, line 18:
```suggestion
  },
  {"p00f/clangd_extensions.nvim",
      ft = { "c", "cpp", "objc", "objcpp", "h", "hpp" },
  }, { "Civitasv/cmake-tools.nvim",
      config = function()
        require("user.cle").cmake_config()
      end,
      ft = { "c", "cpp", "objc", "objcpp", "h", "hpp" },}
```

Uzaaft: I think the default is that the server is not setup. So in theory we wouldn't need that, no?

mehalter: yeah and

```lua
init = function() table.insert(astronvim.lsp.skip_setup, "tsserver") end
```

Uzaaft: Ah got it. I just stole from my old config 😥

owittek: > yeah and
> 
> 
> 
> ```lua
> 
> init = function() table.insert(astronvim.lsp.skip_setup, "clangd") end
> 
> ```

why `table.insert` over `list_insert_unique`?

mehalter: This is a good point. I changed it

mehalter (commented):
I went ahead and added `clangd_extensions`, `cmake-tools.nvim` and also standardized the file name

Uzaaft (approved):
Works like a charm☘️

PR #298, opened 2023-06-21
feat(code-runner): add compiler.nvim

Zeioth:
More info about the project can be found here: https://github.com/Zeioth/Compiler.nvim

### DIFF
--- a/lua/astrocommunity/code-runner/compiler-nvim/README.md
+++ b/lua/astrocommunity/code-runner/compiler-nvim/README.md
@@ -1,0 +1,5 @@
+# compiler.nvim
+
+**Repository:** https://github.com/Zeioth/Compiler.nvim
+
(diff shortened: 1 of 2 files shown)


luxus:
so its depends on overseer?

Uzaaft:
Can you fix the way you import your repo? Other repositories doesn’t seem to need the whole git url

Zeioth:
@luxus yes, we use it to run the calls because it fits perfectly to our need. Compiler.nvim is to overseer what friendly snippets is to luasnip, you could say.

@Uza thank you for that, missed it. I've cleaned up that, and the comments too.

Uzaaft:
Can you fix the Invalid PR title, and we'll be ready to merge 🫡

mehalter:
I went ahead and cleaned this up and merged it in. Thanks for the contribution @Zeioth !

PR #303, opened 2023-06-22
fix(harpoon): typo in concatenation

bruunoromero:
Hello, after the changes in #292 it started to fail with:
``` attempt to concatenate local 'running_tmux_session' (a boolean value)```

This PR should fix the issue.


mehalter (approved):
Thanks!

PR #314, opened 2023-06-25
feat(editing-support): add ultimate-autopair.nvim

windowsrefund:


lua/astrocommunity/editing-support/ultimate-autopair-nvim/init.lua
```diff
@@ -0,0 +1,44 @@
+return {
+  {
+    "windwp/nvim-autopairs",
+    enabled = false,
+  },
+  {
+    "altermo/ultimate-autopair.nvim",
+    lazy = false,
```
luxus, line 8: why did you turn of lazy?

Uzaaft:
@windowsrefund I think the opts should be empty, unless there's a reason why you're setting changing them.

mehalter (changes requested):
Oh, this README.md file should be in the `utlimate-autopair-nvim` folder also can you make the format follow the same as other plugins? This looks like it's more following the format of the README for packs

mehalter (approved):
LGTM

PR #315, opened 2023-06-25
feat: add python debug with arguments

chaozwn:
add python debug with arguments


Uzaaft:
Any reason why this would be required?

chaozwn:
> Any reason why this would be required?

with out it, when debug python. i can't add arguments.

image

Uzaaft:
Ah got it, then it makes sense. Thanks! Next time, include info like that in the PR. :)

Uzaaft (approved):

chaozwn:
> Ah got it, then it makes sense. Thanks! Next time, include info like that in the PR. :)

ok, Understood, I'll make sure to do it next time.

PR #319, opened 2023-06-26
feat: add astro filetype to eslint_d handler in typescript pack

rafaelderolez:
According to [this](https://github.com/jose-elias-alvarez/null-ls.nvim/issues/1552) null-ls issue, [Astro.build](https://astro.build/) filetypes needs to be added as `extra_filetypes`.

I thought it made sense to add it in the typescript pack since you'll be writing js/ts with Astro anyway.

The `astro-language-server` uses Prettier [under the hood](https://github.com/withastro/language-tools/tree/main/packages/language-server) so it doesn't have to be added to the `prettierd` handler.


Uzaaft (approved):
Sensible change. Merging. :)

PR #324, opened 2023-06-28
feat: add vim-smoothie plugin

ypli0629:


lua/astrocommunity/scrolling/vim-smoothie/init.lua
```diff
@@ -0,0 +1,5 @@
+return {
+  "psliwka/vim-smoothie",
+  event = "User AstroFile",
+  init = function() vim.g.smoothie_remapped_commands = { "<C-D>", "<C-U>", "<PageDown>", "<PageUp>" } end,
+}
```
Uzaaft, line 5: According to vim-smoothie: 
vim-smoothie aims for sane defaults, and should work out-of-the-box for most users. In some cases, however, you might want to customize its behavior, by adjusting one or more of the following variables in your vimrc:

I think we should drop the remap.

ypli0629: ok

Uzaaft (approved):
LGTM. Thanks 🫡

PR #325, opened 2023-06-28
feat: change olexsmir/gopher.nvim to ray-x/go.nvim

ypli0629:
[`ray-x/go.nvim`](https://github.com/ray-x/go.nvim) provide more powerful functions

### DIFF
--- a/lua/astrocommunity/pack/go/README.md
+++ b/lua/astrocommunity/pack/go/README.md
@@ -13,4 +13,4 @@ This plugin pack does the following:
   - [iferr](https://github.com/koron/iferr)
   - [impl](https://github.com/josharian/impl)
 - Adds [nvim-dap-go](https://github.com/leoluz/nvim-dap-go) for debugging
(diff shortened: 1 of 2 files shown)


Uzaaft:
@mehalter  While I agree, doesn't the change of available commands in Neovim for Go constitute an breaking change?

mehalter:
I think for the case of AstroCommunity the things the plugins provide should be considered outside of the scope of our API

Uzaaft:
I'll take a look at this in a couple of hours just to make sure it works as expected. Sorry about the delay in merging @ypli0629

luxus:
Not sure about x-rayz plugins… 
Most plugins are a bit messy 

didn’t have a go project right now to test it

Uzaaft:
LGTM. go.nvim seems to me more used, so the change makes sense.

Uzaaft (approved):
LGTM

PR #328, opened 2023-06-29
feat(editing-support): add vim-move plugin

ypli0629:
[`vim-move`](https://github.com/matze/vim-move) provides the ability to move blocks of code


luxus:
is mini.move not working for you?

mehalter (approved):
LGTM! @luxus it doesn't matter if mini.move works haha we should have all the plugins

luxus:
mhh i thought vim plugins need lazy disabled?

Issue #279, opened 2023-06-15
Status bar randomly disappearing with andymass/vim-matchup


saumyajyoti:
The status bar gets disabled at some lines when match-up highlights some code lines.

saumyajyoti:
This can be fixed by below addition:

 init = function()
    vim.g.matchup_matchparen_offscreen = { method = "popup" }
  end,

saumyajyoti:
The above change can be added in matchup lua file

Uzaaft:
@saumyajyoti  Do you have some docs to back up your fix?

saumyajyoti:
I think this is expected behavior for matchup. The default settings is replacing the status-bar. Please refer https://github.com/andymass/vim-matchup#c3-display-matches-off-screen

jay-babu:
should this be added to the default?

jay-babu:
https://github.com/AstroNvim/astrocommunity/pull/331

PR #331, opened 2023-06-29
fix(motion): vim-matchup - matchup_matchparen_offscreen method='popup…

jay-babu:
…' #279

### DIFF
--- a/lua/astrocommunity/motion/vim-matchup/init.lua
+++ b/lua/astrocommunity/motion/vim-matchup/init.lua
@@ -1,5 +1,6 @@
 return {
   "nvim-treesitter/nvim-treesitter",
   dependencies = { "andymass/vim-matchup" },
+  init = function() vim.g.matchup_matchparen_offscreen = { method = "popup" } end,
   opts = { matchup = { enable = true } },
 }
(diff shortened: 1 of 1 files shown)


mehalter (changes requested):
Awesome, do you think we should also add

```lua
vim.g.matchup_matchparen_deferred = 1
```

to the default options as well?

lua/astrocommunity/motion/vim-matchup/init.lua
```diff
@@ -1,5 +1,6 @@
 return {
   "nvim-treesitter/nvim-treesitter",
   dependencies = { "andymass/vim-matchup" },
+  init = function() vim.g.matchup_matchparen_offscreen = { method = "popup" } end,
```
jay-babu, line 4:
```suggestion
  init = function() 
    vim.g.matchup_matchparen_offscreen = { method = "popup" }
    vim.g.matchup_matchparen_deferred = 1
  end,
```

jay-babu:
> Awesome, do you think we should also add
> 
> ```lua
> vim.g.matchup_matchparen_deferred = 1
> ```
> 
> to the default options as well?

sure, why not. i just turned the issue into a PR

mehalter (approved):
LGTM!

PR #335, opened 2023-06-30
fix(clojure): add space after clojure comment character

practicalli-john:
Add comment between `;;` comment characters for clojure and text being commented


mehalter:
@practicalli-john for the scope, and I should clarify this in the contribution guidelines, it should be the name of the already existing folder you are modifying. When you are adding a new plugin it is modifying the categories folder. When you are making changes to an already existing plugin the scope is the name of the plugin. This makes a very nice to read changelog.

mehalter (approved):

PR #346, opened 2023-07-01
feat(pack): Adding ansible pack

Sacro:


lua/astrocommunity/pack/ansible/init.lua
```diff
@@ -0,0 +1,24 @@
...
+    "nvim-treesitter/nvim-treesitter",
+    opts = function(_, opts)
+      if opts.ensure_installed ~= "all" then
+        opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "yaml")
+      end
+    end,
+  },
+  {
+    "jay-babu/mason-null-ls.nvim",
+    opts = function(_, opts) opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "ansible-lint") end,
```
jay-babu, line 14: you want the `null-ls` name, so it is automatically registered https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md#ansiblelint

jay-babu: and add a mapping here: https://github.com/jay-babu/mason-null-ls.nvim/blob/73c68abdf65279e41526eb152876511a8ae84ea2/lua/mason-null-ls/mappings/source.lua#L13

Sacro:
Should be better now!

mehalter:
We should probably also add the `filetype.add` to this pack:

```lua
local function yaml_ft(path, bufnr)
    -- get content of buffer as string
    local content = vim.filetype.getlines(bufnr)
    if type(content) == "table" then content = table.concat(content, "\n") end

    -- check if file is in roles, tasks, or handlers folder
    local path_regex = vim.regex "(tasks\\|roles\\|handlers)/"
    if path_regex and path_regex:match_str(path) then return "yaml.ansible" end
    -- check for known ansible playbook text and if found, return yaml.ansible
    local regex = vim.regex "hosts:\\|tasks:"
    if regex and regex:match_str(content) then return "yaml.ansible" end

    -- return yaml if nothing else
    return "yaml"
  end

  vim.filetype.add {
    extension = {
      yml = yaml_ft,
      yaml = yaml_ft,
    },
  }
```

Sacro:
Looks good, do you want me to add that?

mehalter:
Oh wait, does `ansible-vim` already add this? This might not be necessary

Sacro:
It does, but it's not lazy loaded unless you manually set the filetype first. if we can trigger that automagically that would be nice.

mehalter:
@Sacro if we set the filetype and are using language servers and treesitter, does `ansible-vim` even provide anything else? I think it just provides syntax highlighting and filetype detector which are all handled elsewhere

Sacro:
I find I'm getting better code completion with it enabled.

lua/astrocommunity/pack/ansible/init.lua
```diff
@@ -0,0 +1,24 @@
...
+    "jay-babu/mason-null-ls.nvim",
+    opts = function(_, opts) opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "ansiblelint") end,
+  },
+  {
+    "williamboman/mason-lspconfig.nvim",
+    opts = function(_, opts) opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "ansiblels") end,
+  },
+  {
+    "pearofducks/ansible-vim",
+    ft = "yaml.ansible",
```
mehalter, line 22:
```suggestion
    "pearofducks/ansible-vim",
    init = function()
      -- add vim.filetype
    end,
    ft = "yaml.ansible",
```

I think we can put the filetype add to this plugins initialization to get the lazy loading

mehalter (changes requested):

lua/astrocommunity/pack/ansible/init.lua
mehalter: Oh also could you add ansible-vim mention to the readme

mehalter (approved):
This looks awesome thanks!

PR #351, opened 2023-07-02
fix(nvim-scrollbar): Fix issue with is_available calls

Uzaaft:


mehalter (changes requested):
Can you move the utilities import into the opts function

Uzaaft:
> Can you move the utilities import into the opts function

Done

mehalter (approved):
LGTM

PR #352, opened 2023-07-02
feat(editing-support): Add chatGPT.nvim

Uzaaft:


luxus (approved):

kaka-ruto:
Hi @Uzaaft , where do you set the open ai api key? I know it could be anywhere, but do you have a place/file you'd recommend?

PR #353, opened 2023-07-03
fix(ccc-nvim): ccc.nvim can not show colors from neo-tree `o`.

chaozwn:
change event 'User AstroFile' to 'BufEnter' for open ccc.nvim.


mehalter:
Replaced by #356 to avoid removing lazy loading

PR #363, opened 2023-07-03
fix(zen-mode-nvim): disable/enable winbar when entering/exiting zen mode

Subjective:


Uzaaft:
@Subjective  Could you add why one would want this as a default in astrocom?

mehalter:
I think this is a great default. When going into zen mode you want to remove all of the UI elements and just get a focused window so disabling winbar is key

mehalter (approved):
LGTM

PR #364, opened 2023-07-05
docs(ruby): Remove mentions of html and ruby_ls from docs.

natematykiewicz:
The [Ruby pack](https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/pack/ruby/init.lua) does not install `html` or `ruby_ls`.


Uzaaft:
LGTM.

Uzaaft (approved):

PR #367, opened 2023-07-05
fix(python): added debugpy path auto-resolution using mason-registry

BillKills974:
Mason installs debugpy in a venv in $HOME/.local/share/$NVIM_APPNAME/mason/packages/debugpy/ .
This fixes a bug where DAP would not find debugpy (ModuleNotFoundError : no module named "debugpy") and would not start debugging.
The replaced code would only work if you just installed debugpy or running :DapInstall python and use DAP without exiting nvim, or if you had debugpy installed system-wide (via pip) and not using a venv. It would fail upon new nvim runs.

### DIFF
--- a/lua/astrocommunity/pack/python/init.lua
+++ b/lua/astrocommunity/pack/python/init.lua
@@ -33,6 +33,9 @@ return {
     -- NOTE: ft: lazy-load on filetype
     ft = "python",
     event = "User AstroFile",
-    config = function() require("dap-python").setup("python", {}) end,
+    config = function()
+      local path = require("mason-registry").get_package("debugpy"):get_install_path() .. "/venv/bin/python"
+      require("dap-python").setup(path, {})
+    end,
   },
 }
--- a/lua/astrocommunity/pack/python/init.lua
+++ b/lua/astrocommunity/pack/python/init.lua
@@ -33,9 +33,9 @@ return {
     -- NOTE: ft: lazy-load on filetype
     ft = "python",
     event = "User AstroFile",
-    config = function()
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/pack/python/init.lua
```diff
@@ -33,6 +33,9 @@ return {
     -- NOTE: ft: lazy-load on filetype
     ft = "python",
     event = "User AstroFile",
-    config = function() require("dap-python").setup("python", {}) end,
+    config = function()
+      local path = require("mason-registry").get_package("debugpy"):get_install_path() .. "/venv/bin/python"
+      require("dap-python").setup(path, {})
+    end,
```
mehalter, line 39:
```suggestion
    config = function(_, opts)
      local path = require("mason-registry").get_package("debugpy"):get_install_path() .. "/venv/bin/python"
      require("dap-python").setup(path, opts)
    end,
```

mehalter (changes requested):
Thanks for this! I think we can also go ahead and extend the `config` function here to pass the `opts` into the setup call as the second argument. This will allow users to easily use the `opts` key in lazy to override the options in the setup without losing the mason addition

lua/astrocommunity/pack/python/init.lua
BillKills974: Absolutely. That makes it even better :)

mehalter (approved):
Perfect! I will go ahead and merge this :) Thanks so much! I think I also have some other ideas to improve this integration with mason, so I'll make a follow up PR as well :)

PR #368, opened 2023-07-05
feat(pack): add full-dadbod plugin pack

KenroLe:
I dont know what the namespacing guidelines are so I applied intuition there.
This configuration works on my _machine_, thanks mehalter for lazyloading cmp and making the autocmd less funky.


mehalter (changes requested):
This should probably go under `programming-languages-support` rather than `sql`. This is a category in NeovimCraft that seems the most applicable and also a bit more generally accepting than an `sql` category. Also could you please add an appropriate `README` file to that as well :)

lua/astrocommunity/sql/dadbod/init.lua
```diff
@@ -0,0 +1,23 @@
+return {
+  {
+    "kristijanhusak/vim-dadbod-ui",
+    dependencies = "tpope/vim-dadbod",
+    lazy = false,
+  },
+  {
+    "hrsh7th/nvim-cmp",
```
mehalter, line 8:
```suggestion
    "hrsh7th/nvim-cmp",
    optional = true,
```

We should also probably add the optional key just in case someone has disabled `cmp` this will still work well

mehalter:
@KenroLe I went ahead and added the README and further fixed the folder names :)

@Uzaaft do you think this should be a pack since it provides 2 plugins?

KenroLe:
> @KenroLe I went ahead and added the README and further fixed the folder names :)
> 
> @Uzaaft do you think this should be a pack since it provides 2 plugins?

Tyvm, ill get it right next time 👍

mehalter (approved):

Uzaaft:
> > @KenroLe I went ahead and added the README and further fixed the folder names :)
> > @Uzaaft do you think this should be a pack since it provides 2 plugins?
> 
> Tyvm, ill get it right next time 👍

I think it should be a pack since it uses multiple plugins to provide a way to interface with DB’s

mehalter:
What do you think of how I have it set up currently @Uzaaft ? Good to merge? :) It's ready for review

mehalter:
No one has complained, so I'm just going to ship this! Thanks @KenroLe for the contribution!

PR #369, opened 2023-07-05
perf(python): make sure mason-nvim-dap doesn't set up python debuggers

mehalter:


mehalter:
@BillKills974 could you test this out and see if it works? :) This just makes sure that `mason-nvim-dap` doesn't set up any debuggers automatically and lets `nvim-dap-python` set it up

BillKills974 (approved):
It works as advertised in every test I just made.

PR #370, opened 2023-07-06
fix(lua): disable formatting from `lua_ls` by default

mehalter:
This is a bit hacky, but I think this should work in disabling `lua_ls` formatting and just use `stylua`


mehalter:
This is actually a bit hacky... so I think there is a better way to do this with Neoconf

PR #372, opened 2023-07-06
feat(utility): added ProjectMgr plugin

KenroLe:
Heres another one 🧑‍🌾

### DIFF
--- a/lua/astrocommunity/project/projectmgr/init.lua
+++ b/lua/astrocommunity/project/projectmgr/init.lua
@@ -1,0 +1,7 @@
+return {
+  "charludo/projectmgr.nvim",
+  lazy = false, -- important!
+  keys = {
+    { "<leader>P", "<cmd>ProjectMgr<cr>", desc = "Open ProjectMgr panel" },
+  },
+}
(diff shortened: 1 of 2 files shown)


mehalter (changes requested):
This is awesome! Although I do think this will fit better in the `project` category that we have (more evidence for the upcoming refactor we want to do to allow multiple categories for plugins...)

KenroLe:
Oh, i forgot to rename the pr

lua/astrocommunity/project/projectmgr/init.lua
```diff
@@ -0,0 +1,7 @@
+return {
+  "charludo/projectmgr.nvim",
+  lazy = false, -- important!
```
mehalter, line 3: can we use `event = "VeryLazy"` here?

mehalter (approved):
LGTM!

mehalter:
oh shoot, I just realized naming issue. It should be `projectmgr-nvim`. I'll fix the commit

mehalter:
@KenroLe just a heads up, the name of the folder should be the entire repository name with the `.` replaced with `-` so `projectmgr.nvim` would be `projectmgr-nvim`. I revised the commit :)

PR #375, opened 2023-07-06
feat(utility): Add nvim-toggler

Uzaaft:


lua/astrocommunity/utility/nvim-toggler/init.lua
```diff
@@ -0,0 +1 @@
+return { "nguyenvukhang/nvim-toggler", event = "User Astrofile" }
```
mehalter, line 1: Do you need to run `setup`? Passing in `opts = {}`

mehalter, line 1: Also maybe `event = { "User AstroFile", "InsertEnter" }` 

At least a fix for `Astrofile` -> `AstroFile`

Uzaaft: Fixed

Uzaaft: Fixed

mehalter (approved):
LGTM!

PR #376, opened 2023-07-06
fix(harpoon): Make which-key entry color blue, as it should be

RayJameson:
I believe the Harpoon entry in which-key should be blue because it is a nested keymap,  unfortunately when using `keys` it doesn't get highlighted properly, this pr fixes it.

**Before:**
image

**After:**
image


owittek (approved):
LGTM

PR #377, opened 2023-07-07
feat(debugging): add telescope dap extension

chaozwn:
This plugin is also overriding dap internal ui, so running any dap command, which makes use of the internal ui, will result in a telescope prompt.

### DIFF
--- a/lua/astrocommunity/debugging/telescope-dap-nvim/init.lua
+++ b/lua/astrocommunity/debugging/telescope-dap-nvim/init.lua
@@ -2,7 +2,7 @@ return {
   {
     "nvim-telescope/telescope-dap.nvim",
     name = "telescope_dap",
-    event = "VeryLazy",
+    event = "User AstroFile",
     keys = {
       {
         "<leader>fdc",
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/debugging/telescope-dap-nvim/init.lua
```diff
@@ -0,0 +1,60 @@
+return {
+  {
+    "nvim-telescope/telescope-dap.nvim",
+    name = "telescope_dap",
+    event = "VeryLazy",
```
owittek, line 5:
```suggestion
    event = "User AstroFile",
```
There should be a better way to lazy-load this plugin. My initial suggestion would be this unless someone has a better idea.

lua/astrocommunity/debugging/telescope-dap-nvim/init.lua
owittek, line 1: Perhaps adding `local prefix = "<leader>fd"` at the top would be nice so it's easier to modify in case it should change.

owittek (changes requested):

owittek (approved):
LGTM, thank you!

Issue #309, opened 2023-06-23
Configuration for astrocommunity.note-taking.obsidian-nvim


mortang2410:
The current minimal config for obsidian does not truly work for following links to other markdown files.
This is my configuration which works, modified from the default config in the [obsidian.nvim repo](https://github.com/epwalsh/obsidian.nvim).  

```lua
return {
  "epwalsh/obsidian.nvim",
  lazy = true,
  event = { "BufReadPre  */obsidian-vault/*.md" },
  -- If you want to use the home shortcut '~' here you need to call 'vim.fn.expand':
  -- event = { "bufreadpre " .. vim.fn.expand "~" .. "/my-vault/**.md" },
  dependencies = {
    -- Required.
    "nvim-lua/plenary.nvim",

    -- Optional, for completion.
    "hrsh7th/nvim-cmp",

    -- Optional, for search and quick-switch functionality.
    "nvim-telescope/telescope.nvim",

    -- Optional, an alternative to telescope for search and quick-switch functionality.
    -- "ibhagwan/fzf-lua"

    -- Optional, another alternative to telescope for search and quick-switch functionality.
    -- "junegunn/fzf",
    -- "junegunn/fzf.vim"

    -- Optional, alternative to nvim-treesitter for syntax highlighting.
    -- "godlygeek/tabular",
    -- "preservim/vim-markdown",
  },
  opts = {
    dir = "~/obsidian-vault", -- no need to call 'vim.fn.expand' here

    -- -- Optional, if you keep notes in a specific subdirectory of your vault.
    -- notes_subdir = "notes",
    --
    -- daily_notes = {
    --   -- Optional, if you keep daily notes in a separate directory.
    --   folder = "notes/dailies",
    --   -- Optional, if you want to change the date format for daily notes.
    --   date_format = "%Y-%m-%d"
    -- },

    -- Optional, completion.
    completion = {
      nvim_cmp = true, -- if using nvim-cmp, otherwise set to false
    },

    -- -- Optional, customize how names/IDs for new notes are created.
    -- note_id_func = function(title)
    --   -- Create note IDs in a Zettelkasten format with a timestamp and a suffix.
    --   -- In this case a note with the title 'My new note' will given an ID that looks
    --   -- like '1657296016-my-new-note', and therefore the file name '1657296016-my-new-note.md'
    --   local suffix = ""
    --   if title ~= nil then
    --     -- If title is given, transform it into valid file name.
[...]

PR #378, opened 2023-07-07
fix(obsidian-nvim): added obsidian.nvim config and changed the readme

mortang2410:
As requested from #309. We need this config to follow links to other markdown files. Please check if anything is amiss. 

Edit: I see the review checklist asks for "Proper usage of `opts` table rather than setting things up with the `config` function." I'm not sure if that is satisfied here.

### DIFF
--- a/lua/astrocommunity/note-taking/obsidian-nvim/init.lua
+++ b/lua/astrocommunity/note-taking/obsidian-nvim/init.lua
@@ -27,7 +27,7 @@ return {
     -- "preservim/vim-markdown",
   },
   opts = {
-    dir = "~/obsidian-vault", -- specify the vault location. no need to call 'vim.fn.expand' here
+    dir = vim.env.HOME .. "/obsidian-vault", -- specify the vault location. no need to call 'vim.fn.expand' here
 
     -- -- Optional, if you keep notes in a specific subdirectory of your vault.
     -- notes_subdir = "notes",
--- a/lua/astrocommunity/note-taking/obsidian-nvim/init.lua
+++ b/lua/astrocommunity/note-taking/obsidian-nvim/init.lua
@@ -1,17 +1,125 @@
 return {
   "epwalsh/obsidian.nvim",
-  cmd = {
-    "ObsdianBacklinks",
-    "ObsidianToday",
-    "ObsidianYesterday",
-    "ObsidianOpen",
-    "ObsidianNew",
-    "ObsidianSearch",
-    "ObsidianQuickSwitch",
-    "ObsidianLink",
-    "ObsidianLinkNew",
-    "ObsidianFollowLink",
-    "ObsidianTemplate",
+  lazy = true,
+  -- the obsidian vault in this default config  ~/obsidian-vault
+  event = { "BufReadPre  */obsidian-vault/*.md" },
+  -- If you want to use the home shortcut '~' here you need to call 'vim.fn.expand':
+  -- event = { "bufreadpre " .. vim.fn.expand "~" .. "/my-vault/**.md" },
+  dependencies = {
+    -- Required.
+    "nvim-lua/plenary.nvim",
+
+    -- Optional, for completion.
+    "hrsh7th/nvim-cmp",
+
+    -- Optional, for search and quick-switch functionality.
+    "nvim-telescope/telescope.nvim",
+
+    -- Optional, an alternative to telescope for search and quick-switch functionality.
+    -- "ibhagwan/fzf-lua"
+
+    -- Optional, another alternative to telescope for search and quick-switch functionality.
+    -- "junegunn/fzf",
+    -- "junegunn/fzf.vim"
+
+    -- Optional, alternative to nvim-treesitter for syntax highlighting.
+    -- "godlygeek/tabular",
+    -- "preservim/vim-markdown",
   },
-  opts = { completion = { nvim_cmp = true } },
+  opts = {
+    dir = "~/obsidian-vault", -- specify the vault location. no need to call 'vim.fn.expand' here
+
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/note-taking/obsidian-nvim/init.lua
```diff
@@ -1,17 +1,125 @@
...
+    -- "junegunn/fzf",
+    -- "junegunn/fzf.vim"
+
+    -- Optional, alternative to nvim-treesitter for syntax highlighting.
+    -- "godlygeek/tabular",
+    -- "preservim/vim-markdown",
   },
-  opts = { completion = { nvim_cmp = true } },
+  opts = {
+    dir = "~/obsidian-vault", -- specify the vault location. no need to call 'vim.fn.expand' here
```
Uzaaft, line 30: I think it should use `vim.env.HOME `, to make it cross platform. I.e `vim.env.HOME`..`/obsidian-vault`

Uzaaft (dismissed):

owittek:
Wouldn't it be enough to just add your 2 changes to the `init.lua` instead of pasting the entire default config as you're not changing the defaults?

mortang2410:
>  just add your 2 changes

Setting up the `gf` keybinding, for instance, to follow links is done in the config, so we have to copy paste some of it. I also disabled a few things like the `notes`, `daily_notes`, `tabular`, `vim-markdown` (as we already use treesitter), new note ID (it would create new notes upon the slightest mistypes etc.). I think those are sane defaults but please feel free to discuss.

lua/astrocommunity/note-taking/obsidian-nvim/init.lua
mortang2410: Thanks. I made the change.

owittek:
Pardon me then, I came from the issue and the first part as well as the comments seem to be directly taken from the obsidian README.

But my point stands, it's unclear for me what parts deviate from the default config if you also put in all the defaults and leave all the comments in.

Also we'd have to handle line 91-95 in a way that is OS agnostic. For that we'd have to use a function for opts instead.

Lastly we should setup the key bind in a different way than to place it in the `config` function for various reasons.

mortang2410:
> Also we'd have to handle line 91-95 in a way that is OS agnostic. For that we'd have to use a function for opts instead.
> Lastly we should setup the key bind in a different way than to place it in the `config` function for various reasons.

I agree with both of these points. I admit, however, that I am unfamiliar with how astronvim should handle both of these. If you have solutions for them, please share.

owittek:
I can refactor `opts` when I got time and with the key bind part perhaps @RayJameson could help since he did that kind of stuff for harpoon before.

RayJameson:
Shouldn't it work with `keys` property in the plugin setup? I did that thing for Harpoon because there is a top-level mapping that shows in which-key so It would look pretty, otherwise, I would stick with using `keys`

owittek:
Hmm there's still an issue with the setup handler in my code. Testing right now...

owittek (approved):
I like my own changes, soo ...

lua/astrocommunity/note-taking/obsidian-nvim/init.lua
```diff
@@ -1,17 +1,62 @@
...
+
+    -- Optional, by default when you use `:ObsidianFollowLink` on a link to an external
+    -- URL it will be ignored but you can customize this behavior here.
+    follow_url_func = function(url)
+      local this_os = vim.loop.os_uname().sysname
+      -- Open the URL in the default web browser.
+      if this_os == "Darwin" then
+        vim.fn.jobstart { "open", url }
+      elseif this_os == "Linux" then
+        vim.fn.jobstart { "xdg-open", url }
```
Uzaaft, line 58:
```suggestion
        vim.fn.jobstart { "xdg-open", url }
     elseif this_os == "Windows" then
     vim.fn.jobstart { "explorer", url }
     
```

Uzaaft:
@owittek Added a suggestion for windows support as well.

lua/astrocommunity/note-taking/obsidian-nvim/init.lua
```diff
@@ -1,17 +1,65 @@
...
+     elseif this_os == "Windows" then
+     vim.fn.jobstart { "explorer", url }
+
```
owittek, line 61:
```suggestion
```

owittek (commented):
the pipeline does not like us

Uzaaft:
@owittek  Apparently `utils.system_open` should be enough

lua/astrocommunity/note-taking/obsidian-nvim/init.lua
```diff
@@ -1,17 +1,64 @@
...
+      local this_os = vim.loop.os_uname().sysname
+      -- Open the URL in the default web browser.
+      if this_os == "Darwin" then
+        vim.fn.jobstart { "open", url }
+      elseif this_os == "Linux" then
+        vim.fn.jobstart { "xdg-open", url }
+     elseif this_os == "Windows" then
+     vim.fn.jobstart { "explorer", url }
+      end
+    end,
```
mehalter, line 62:
```suggestion
    follow_url_func = vim.ui.open or require("astronvim.utils").system_open,
```

Uzaaft, line 62:
```suggestion
    follow_url_func = require("astronvim.utils").system_open
```

mehalter (changes requested):
@owittek

Uzaaft (approved):

mehalter (approved):
LGTM

PR #379, opened 2023-07-07
feat(editing-support): add multicursors.nvim plugin

fynnfluegge:
Adds multicursors.nvim plugin for multi cursor editing support.


mehalter (approved):
Looks great thanks! I just updated the PR title so that it's valid :)

PR #381, opened 2023-07-07
fix(typescript): lazy load tsc on cmd

Subjective:


owittek:
Hey! Could you please explain why we should load on the command instead of filetype?

Subjective:
From what I can tell, the only time you use TSC is when running the command for it. I don't think it's necessary to load it every time. But supposing we were to load it on filetype, then I think it would make sense to add typescriptreact, javascript, and javascriptreact filetypes.

Subjective:
Previously, TSC was also being loaded upon opening any filetype (from the AstroFile event), which is not ideal.

owittek:
Makes sense!

owittek (approved):

PR #387, opened 2023-07-09
feat(debugging): add nvim-dap-virtual-text

chaozwn:
This plugin adds virtual text support to `nvim-dap`. `nvim-treesitter` is used to find variable definitions.


Uzaaft (approved):
LGTM

PR #394, opened 2023-07-09
feat(vimtex): add which-key mapping descriptions

Subjective:
<details>
<summary>Before</summary>

image

</details>

<details>
<summary>After</summary>

image

</details>

### DIFF
--- a/lua/astrocommunity/markdown-and-latex/vimtex/init.lua
+++ b/lua/astrocommunity/markdown-and-latex/vimtex/init.lua
@@ -1,1 +1,107 @@
-return { "lervag/vimtex", lazy = false }
+return {
+  "lervag/vimtex",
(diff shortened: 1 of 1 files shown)


Uzaaft (approved):
LGTM

Uzaaft:
@Subjective  Why didn't you use the keys table in lazy.nvim to setup the keys?

Subjective:
@Uzaaft VimTex sets up the keybinds by default upon entering a latex buffer, but doesn't add the descriptions necessary for which-key. The main reason for not using the lazy table is that I want the labels to be specific to that buffer.

mehalter (approved):
Cool thanks. Just be sure to track the changes on `vimtex` and come back when these mappings inevitably change 😅

PR #409, opened 2023-07-11
fix(nvim-hlslens): add lazy event

fynnfluegge:
Adds a lazy event to hlslens, since plugin is not loaded when searching buffer.


mehalter (approved):
LGTM!

PR #411, opened 2023-07-11
feat(lsp_lines-nvim): disable virtual text and add `<leader>uD` mapping

TheSast:
I wish I could have implemented the fix and the keybinds better.
`virtual_text` should be off when using `lsp_lines.nvim`

### DIFF
--- a/lua/astrocommunity/diagnostics/lsp_lines-nvim/init.lua
+++ b/lua/astrocommunity/diagnostics/lsp_lines-nvim/init.lua
@@ -1,6 +1,17 @@
 return {
   "https://git.sr.ht/~whynothugo/lsp_lines.nvim",
   event = "LspAttach",
+  keys = {
+    {
+      "<Leader>uD",
+      function()
+        vim.diagnostic.config {
+          virtual_text = not require("lsp_lines").toggle(),
+        }
+      end,
+      desc = "Toggle virtual diagnostic lines",
+    },
+  },
   init = function() require("astronvim.utils.ui").toggle_diagnostics() end,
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/diagnostics/lsp_lines-nvim/init.lua
```diff
@@ -1 +1,17 @@
...
+      function()
+        vim.diagnostic.config {
+          virtual_text = not require("lsp_lines").toggle(),
+        }
+      end,
+      desc = "Toggle virtual diagnostic lines",
+    },
+  },
+  init = function() require("astronvim.utils.ui").toggle_diagnostics() end,
+  opts = {},
```
mehalter, line 16:
```suggestion
  opts = {},
  config = function(_, opts)
    -- disable diagnostic virtual text
    local lsp_utils = require "astronvim.utils.lsp"
    lsp_utils.diagnostics[3].virtual_text = false
    vim.diagnostic.config(lsp_utils.diagnostics[vim.g.diagnostics_mode])
    require("lsp_lines").setup(opts)
  end,
```

mehalter (changes requested):
Could you try out these changes @TheSast ?

lua/astrocommunity/diagnostics/lsp_lines-nvim/init.lua
TheSast: Will test this tomorrow as soon as I can

mehalter (approved):
LGTM

PR #413, opened 2023-07-12
fix(nvim-spectre): Make which-key entry color blue

bruunoromero:
I believe the Spectre entry in which-key should be blue because it is a nested keymap, unfortunately when using keys it doesn't get highlighted properly, this pr fixes it.


owittek (approved):
LGTM!

owittek:
Oh, you'll have to change the scope from `spectre` to `nvim-spectre` before I can merge the changes

PR #414, opened 2023-07-12
fix(grapple-nvim): Make which-key entry color blue, as it should be

TheSast:
This is definetly not a copy of #376


owittek (approved):
I was waiting for people to definitely not copy that :D

PR #415, opened 2023-07-12
feat(pack): add wgsl pack

mlambir:


lua/astrocommunity/pack/wgsl/init.lua
```diff
@@ -0,0 +1,23 @@
...
+    lazy = false,
+    build = "cp -r queries ../nvim-treesitter/queries/wgsl",
+    config = function()
+      vim.filetype.add { extension = { wgsl = "wgsl" } }
+      require("nvim-treesitter.parsers").get_parser_configs().wgsl = {
+        install_info = {
+          url = vim.fn.stdpath "data" .. "/lazy/tree-sitter-wgsl",
+          files = { "src/parser.c", "src/scanner.c" },
+        },
+      }
```
owittek, line 15: This code should be moved into the `opts` function

lua/astrocommunity/pack/wgsl/init.lua
```diff
@@ -0,0 +1,23 @@
+local utils = require "astronvim.utils"
+return {
+  {
+    "szebniok/tree-sitter-wgsl",
+    dependencies = "nvim-treesitter/nvim-treesitter",
+    lazy = false,
```
owittek, line 6:
```suggestion
    event = "User AstroFile",
```

not sure if that works for this particular plugin but couldn't you lazy load when actually opening a file buffer?

owittek (changes requested):

lua/astrocommunity/pack/wgsl/init.lua
mehalter: No this should be in the `config` function in pretty sure

lua/astrocommunity/pack/wgsl/init.lua
mehalter: I think instead of that we should just reverse the dependencies. Make treesitter dependent on this package. Then the treesitter module will trigger this lazy loading.

mehalter: I have a suggestion here

lua/astrocommunity/pack/wgsl/init.lua
```diff
@@ -0,0 +1,23 @@
...
+      vim.filetype.add { extension = { wgsl = "wgsl" } }
+      require("nvim-treesitter.parsers").get_parser_configs().wgsl = {
+        install_info = {
+          url = vim.fn.stdpath "data" .. "/lazy/tree-sitter-wgsl",
+          files = { "src/parser.c", "src/scanner.c" },
+        },
+      }
+    end,
+    opts = function(_, opts) opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "wgsl") end,
+  },
```
mehalter, line 18:
```suggestion
  {
    "nvim-treesitter/nvim-treesitter",
    optional = true,
    dependencies = {
      {
        "szebniok/tree-sitter-wgsl",
        build = "cp -r queries ../nvim-treesitter/queries/wgsl",
        config = function()
          vim.filetype.add { extension = { wgsl = "wgsl" } }
          require("nvim-treesitter.parsers").get_parser_configs().wgsl = {
            install_info = {
              url = vim.fn.stdpath "data" .. "/lazy/tree-sitter-wgsl",
              files = { "src/parser.c", "src/scanner.c" },
            },
          }
        end,
      },
    },
    opts = function(_, opts) opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "wgsl") end,
  },
```

mehalter (changes requested):

PR #418, opened 2023-07-13
feat(grapple-nvim): change grapple icon from arrow to hook to distinguish from harpoon

Subjective:
**Before**:
image
**After**:
image

Definitely not opinionated or anything :)


owittek (approved):
👍

PR #419, opened 2023-07-13
fix(octo-nvim): which-key description and color for nested entries

RayJameson:
Yet another one of like #376 and what after it. Also, remove default `mappings` which the plugin sets, to avoid keymap mess


TheSast:
Did you mean: "Make which-key entries color blue, as they should be" ?

owittek (approved):
LGTM

RayJameson:
> Did you mean: "Make which-key entries color blue, as they should be" ?

yea, my bad :D

PR #420, opened 2023-07-13
feat(remote-development): Add netman.nvim

Uzaaft:


luxus:
does netman make sense without intigrating it into neo-tree?

Uzaaft:
Yes, it's mentioned in the docs.

mehalter (changes requested):
We should also configure the neo-tree integration here as well if neo-tree exists using the `optional` key.

lua/astrocommunity/remote-development/netman-nvim/init.lua
```diff
@@ -0,0 +1,4 @@
+return {
+  "miversen33/netman.nvim",
+  opts = {},
+}
```
mehalter, line 4:
```suggestion
return {
  {
    "miversen33/netman.nvim",
    cmd = {
      "NmloadProvider",
      "Nmlogs",
      "Nmdelete",
      "Nmread",
      "Nmwrite",
    },
    opts = {},
  },
  {
    "nvim-neo-tree/neo-tree.nvim",
    dependencies = { "netman.nvim" },
    optional = true,
    opts = function(_, opts)
      opts.sources = require("astronvim.utils").list_insert_unique(opts.sources, "netman.ui.neo-tree")
    end,
  },
}
```

mehalter (changes requested):

mehalter (approved):
LGTM!

PR #424, opened 2023-07-13
fix: Make which-key entry color blue, as it should be

RayJameson:
This PR fixes remaining plugins with wrong which-key colors


RayJameson:
I'm not sure which scope should it be in PR subject

owittek:
@mehalter will have to merge this due to the PR message not having a scope

mehalter (approved):
LGTM!

PR #439, opened 2023-07-14
fix(zen-mode-nvim): ensure winbar stays disabled and restore old mini.indentscope options

Subjective:
Previously, the winbar would enable itself on various events (switching buffers, writing to file, which-key popups, telescope, etc.) I added an autocmd that will disable the winbar for all these events while ZenMode is enabled. This does feel a bit hacky though, so please let me know if there's a better way to toggle the winbar.

Additionally, this fixes an issue where previous mini.indentscope options would not be restored upon exiting ZenMode.


owittek (approved):
👍

PR #440, opened 2023-07-14
fix(neotest): Fix bug in neotest by checking for existing library key table 

### LINKED ISSUE #406: Add neodev integration for neotest community pack
The [neotest README](https://github.com/nvim-neotest/neotest#configuration) says:

>It is highly recommended to use [neodev.nvim](https://github.com/folke/neodev.nvim) to enable type checking for neotest to get type checking, documentation and autocompletion for all API functions.
>
> ```lua
> require("neodev").setup({
>   library = { plugins = { "neotest" }, types = true },
>   ...
> })
> ```

Could we please add this to the astrocommunity's neotest pack [here](https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/test/neotest/init.lua)?

Uzaaft:
I’d let mehalter double check this one

Closes #406

### DIFF
--- a/lua/astrocommunity/test/neotest/init.lua
+++ b/lua/astrocommunity/test/neotest/init.lua
@@ -8,8 +8,12 @@ return {
     {
       "folke/neodev.nvim",
       opts = function(_, opts)
-        opts.library.plugins = require("astronvim.utils").list_insert_unique(opts.library.plugins, "neotest")
-        opts.library.types = true
+        return require("astronvim.utils").extend_tbl(opts, {
+          library = {
+            plugins = { "neotest" },
+            types = true,
+          },
+        })
       end,
     },
   },
--- a/lua/astrocommunity/test/neotest/init.lua
+++ b/lua/astrocommunity/test/neotest/init.lua
@@ -8,12 +8,9 @@ return {
     {
       "folke/neodev.nvim",
       opts = function(_, opts)
-        return require("astronvim.utils").extend_tbl(opts, {
-          library = {
-            plugins = { "neotest" },
-            types = true,
-          },
-        })
+        opts.library = opts.library or {}
+        opts.library.plugins = require("astronvim.utils").list_insert_unique(opts.library.plugins, "neotest")
+        opts.library.types = true
       end,
     },
   },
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/test/neotest/init.lua
```diff
@@ -8,8 +8,12 @@ return {
...
       "folke/neodev.nvim",
       opts = function(_, opts)
-        opts.library.plugins = require("astronvim.utils").list_insert_unique(opts.library.plugins, "neotest")
-        opts.library.types = true
+        return require("astronvim.utils").extend_tbl(opts, {
+          library = {
+            plugins = { "neotest" },
+            types = true,
+          },
+        })
```
mehalter, line 16:
```suggestion
        opts.library = opts.library or {}
        opts.library.plugins = require("astronvim.utils").list_insert_unique(opts.library.plugins, "neotest")
        opts.library.types = true
```

mehalter (changes requested):
This is the incorrect approach @Uzaaft . The original method was correct, we just forgot to check if the library key was a table or not

mehalter (approved):
LGTM!

lua/astrocommunity/test/neotest/init.lua
```diff
@@ -8,6 +8,7 @@ return {
     {
       "folke/neodev.nvim",
       opts = function(_, opts)
+        opts.library = opts.library or {}
         opts.library.plugins = require("astronvim.utils").list_insert_unique(opts.library.plugins, "neotest")
```
mehalter, line 12:
```suggestion
        if opts.library.plugins ~= true then
          opts.library.plugins = require("astronvim.utils").list_insert_unique(opts.library.plugins, "neotest")
        end
```

mehalter (changes requested):
`library.plugins` can technically be `true` which I believe enables all plugins

lua/astrocommunity/test/neotest/init.lua
mehalter: Docs for reference: https://github.com/folke/neodev.nvim#%EF%B8%8F-configuration

mehalter (approved):
LGTM!

PR #441, opened 2023-07-14
fix(neogit): Need to setup neogit before it runs

dsisnero:
Add config setup to setup neogit so it runs

### DIFF
--- a/lua/astrocommunity/git/neogit/init.lua
+++ b/lua/astrocommunity/git/neogit/init.lua
@@ -10,7 +10,7 @@ return {
       { "nvim-lua/plenary.nvim" },
     },
     event = "User AstroGitFile",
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/git/neogit/init.lua
```diff
@@ -10,6 +10,7 @@ return {
       { "nvim-lua/plenary.nvim" },
     },
     event = "User AstroGitFile",
+    config = function(_, opts) require("neogit").setup(opts) end,
```
mehalter, line 13:
```suggestion
    opts = {},
```

mehalter (approved):
Small copy edit to align with other specs and keep things simple. This does the same thing :)

PR #443, opened 2023-07-15
fix(vim-illuminate)!: rename plugin to follow convention

TheSast:
Should breaking changes only be considered such if they break compared to a previous releases?


Uzaaft (approved):
Sorry about the mistake 😅

Uzaaft:
Makes sense to mark this one as breaking @TheSast . Thanks for catching the mistake!!

PR #445, opened 2023-07-16
feat(motion): add marks.nvim improve nvim mark function

chaozwn:
add marks.nvim improve nvim mark function

### DIFF
--- a/lua/astrocommunity/motion/marks-nvim/init.lua
+++ b/lua/astrocommunity/motion/marks-nvim/init.lua
@@ -1,0 +1,15 @@
+return {
+  "chentoast/marks.nvim",
+  event = "BufEnter",
+  opts = {
+    excluded_filetypes = {
+      "qf",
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/motion/marks-nvim/init.lua
```diff
@@ -0,0 +1,15 @@
+return {
+  "chentoast/marks.nvim",
+  event = "BufEnter",
```
Uzaaft, line 3: Considering the different excluded filetypes, wouldn't `event = "User AstroFile"` make more sense?

chaozwn: ok

mehalter (approved):
LGTM

mehalter:
Not sure where the `editing-support` came from, but I went ahead and fixed the PR title to show the correct category

PR #448, opened 2023-07-16
fix(mini-surround): load on User AstroFile event

pkazmier:


Uzaaft (approved):

lua/astrocommunity/motion/mini-surround/init.lua
```diff
@@ -7,6 +7,7 @@ return {
   { "machakann/vim-sandwich", enabled = false },
   {
     "echasnovski/mini.surround",
+    event = "User AstroFile",
```
mehalter, line 10:
```suggestion
    event = { "User AstroFile", "InsertEnter" }
```

@Uzaaft this would be a good change

PR #449, opened 2023-07-16
fix(mini-surround): use keys to lazy load instead

pkazmier:


lua/astrocommunity/motion/mini-surround/init.lua
```diff
@@ -7,16 +7,29 @@ return {
...
-        highlight = prefix .. "h", -- Highlight surrounding
-        replace = prefix .. "r", -- Replace surrounding
-        update_n_lines = prefix .. "n", -- Update `n_lines`
+        add = "gza", -- Add surrounding in Normal and Visual modes
+        delete = "gzd", -- Delete surrounding
+        find = "gzf", -- Find surrounding (to the right)
+        find_left = "gzF", -- Find surrounding (to the left)
+        highlight = "gzh", -- Highlight surrounding
+        replace = "gzr", -- Replace surrounding
+        update_n_lines = "gzn", -- Update `n_lines`
```
mehalter, line 32: This change should be reverted.

mehalter (changes requested):
Awesome! Just need to revert back the `opts` and it's good to go!

lua/astrocommunity/motion/mini-surround/init.lua
pkazmier: I don't understand. You need those lines in order for mini-surround to map the keys to the actions. No?

mehalter: They were already set previously

mehalter: We were just using a common `prefix` variable rather than hard coding each one with the same prefix. So we should stay with that.

pkazmier: Oh, yes, I totally missed that. Will fix.

mehalter, line 32:
```suggestion
        add = prefix .. "a", -- Add surrounding in Normal and Visual modes
        delete = prefix .. "d", -- Delete surrounding
        find = prefix .. "f", -- Find surrounding (to the right)
        find_left = prefix .. "F", -- Find surrounding (to the left)
        highlight = prefix .. "h", -- Highlight surrounding
        replace = prefix .. "r", -- Replace surrounding
        update_n_lines = prefix .. "n", -- Update `n_lines`
```

mehalter (changes requested):

mehalter (approved):
LGTM!

PR #450, opened 2023-07-17
feat(editing-support): add nvim-neoclip plugin

tku137:
Adding plugin specifications for [nvim-neoclip](https://github.com/AckslD/nvim-neoclip.lua). Telescope integration and single key-binding under the find menu. Lazy loaded to minimise startup-time.


lua/astrocommunity/editing-support/nvim-neoclip/init.lua
```diff
@@ -0,0 +1,14 @@
+return {
+  "AckslD/nvim-neoclip.lua",
+  event = { "User AstroFile" },
```
mehalter, line 3:
```suggestion
  event = { "User AstroFile", "InsertEnter" },
```

This will allow the plugin to also work on an initial new file which would require text to be inserted to have something to yank

lua/astrocommunity/editing-support/nvim-neoclip/init.lua
```diff
@@ -0,0 +1,14 @@
...
+  dependencies = {
+    { "nvim-telescope/telescope.nvim" },
+  },
+  config = function()
+    require("neoclip").setup()
+    require("telescope").load_extension "neoclip"
+  end,
```
mehalter, line 10:
```suggestion
  config = function(_, opts)
    require("neoclip").setup(opts)
    require("telescope").load_extension "neoclip"
  end,
```

mehalter (changes requested):
Just a few suggestions to fix and also the folder name should be `nvim-neoclip-lua` because it shold be the direct repository name with `.` replaced with `-`

lua/astrocommunity/editing-support/nvim-neoclip/init.lua
tku137: good idea, thanks

mehalter (approved):
Thanks for this contribution! Looks awesome :D

mehalter:
@tku137 one more thing, the category here should be `register` and not `editing-support` as on neovim craft: http://neovimcraft.com/plugin/AckslD/nvim-neoclip.lua/index.html

tku137:
Oh snap! Changed that. Sorry.

mehalter (approved):
Awesome, this is perfect. Thanks so much!

tku137:
Thanks! I'm super happy to contribute :)

PR #454, opened 2023-07-18
feat(debugging): add nvim-dap-repl-highlights

tku137:
Add [nvim-dap-repl-highlights](https://github.com/LiadOz/nvim-dap-repl-highlights) plugin. Tree-sitter setup needs to be called after plugin setup or `dap_repl` language won't be found. No `opts` for plugin setup, as far as I can see.


lua/astrocommunity/debugging/nvim-dap-repl-highlights/init.lua
```diff
@@ -0,0 +1,14 @@
...
+  config = function()
+    require("nvim-dap-repl-highlights").setup()
+    require("nvim-treesitter.configs").setup {
+      highlight = {
+        enable = true,
+      },
+      ensure_installed = { "dap_repl" },
+    }
+  end,
+}
```
mehalter, line 14:
```suggestion
local utils = require "astronvim.utils"
return {
  "nvim-treesitter/nvim-treesitter",
  dependencies = {
    "LiadOz/nvim-dap-repl-highlights",
    dependencies = { "mfussenegger/nvim-dap" },
    opts = {},
  },
  opts = function(_, opts)
    if opts.ensure_installed ~= "all" then
      opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "dap_repl")
    end
  end,
}
```

mehalter (changes requested):
A bit of a restructure recommendations

mehalter (approved):
LGTM!

tku137:
Out of curiosity: the restructuring enables the use of `list_insert_unique()` on the `ensure_installed` list of tree sitter? Did I get that right?

mehalter:
Yeah just to make sure it's type safe and also provides the options that are being set in a way that the end user can override them if they need to. This also allows us to leverage the built in support for lazy.nvim to call setup rather than us specifying it manually

PR #457, opened 2023-07-18
feat(catppuccin): enable catppuccin integrations automatically

mehalter:


luxus (approved):
wonderful idea.

mehalter:
Now we just need people to do this for other colorschemes :joy:

PR #459, opened 2023-07-19
fix(catppuccin): enable alpha.nvim by default

Sacro:
We use alpha.nvim for the start screen, might as well enable it.

Also sorting integrations alphabetically makes adding/removing later nicer.


mehalter (approved):
LGTM!

PR #470, opened 2023-07-21
feat(lsp): add ray-x/lsp_signature.nvim

Juniar-Rakhman:
added lsp_signature plugin

### DIFF
--- a/lua/astrocommunity/lsp/lsp-signature-nvim/README.md
+++ b/lua/astrocommunity/lsp/lsp-signature-nvim/README.md
@@ -1,0 +1,5 @@
+# lsp_signature.nvim
+
+**Repository:** <https://github.com/ray-x/lsp_signature.nvim>
+
+Show function signature when you type.
--- a/lua/astrocommunity/lsp/lsp-signature-nvim/README.md
+++ b/lua/astrocommunity/lsp/lsp-signature-nvim/README.md
@@ -3,3 +3,5 @@
 **Repository:** <https://github.com/ray-x/lsp_signature.nvim>
 
 Show function signature when you type.
+
+Additional hint is visually distracting therefore "hint_enable" option is set to false.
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/lsp/lsp-signature-nvim/init.lua
```diff
@@ -0,0 +1,9 @@
+return {
+  {
+    "ray-x/lsp_signature.nvim",
+    event = "BufRead",
+    opts = {
+      hint_enable = false, -- disable hints as it will crash in some terminal
```
Uzaaft, line 6: Do you have docs for this?

lua/astrocommunity/lsp/lsp-signature-nvim/README.md
```diff
@@ -0,0 +1,5 @@
+# lsp_signature.nvim
+
+**Repository:** <https://github.com/ray-x/lsp_signature.nvim>
+
+Show function signature when you type.
```
Uzaaft, line 5: Could you add some info here about hint_enable and why it's disabled?

lua/astrocommunity/lsp/lsp-signature-nvim/init.lua
```diff
@@ -0,0 +1,9 @@
...
+    },
+  },
+}
```
mehalter, line 9:
```suggestion
return {
  "neovim/nvim-lspconfig",
  dependencies = {
    "ray-x/lsp_signature.nvim",
    opts = {
      hint_enable = false, -- disable hints as it will crash in some terminal
    },
  },
}
```

Should we just lazy load this with `lspconfig`? Since you won't be able to use lsp_signature without LSP

mehalter (dismissed):

lua/astrocommunity/lsp/lsp-signature-nvim/README.md
Juniar-Rakhman: done :)

Uzaaft (approved):

PR #471, opened 2023-07-21
fix(catppuccin): add recommended telescope config

Sacro:
Recommended in the FAQs.

### DIFF
--- a/lua/astrocommunity/colorscheme/catppuccin/init.lua
+++ b/lua/astrocommunity/colorscheme/catppuccin/init.lua
@@ -1,20 +1,32 @@
 return {
-  "catppuccin/nvim",
-  name = "catppuccin",
-  opts = {
-    integrations = {
-      alpha = true,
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/colorscheme/catppuccin/init.lua
```diff
@@ -1,20 +1,32 @@
...
+    },
+  },
+  {
+    "nvim-telescope/telescope.nvim",
+    optional = true,
+    opts = {
+      highlight = {
+        enable = true,
+        additional_vim_regex_highlighting = false,
+      },
```
Uzaaft, line 29: Shouldn't these opts me extended? In case of an already existing user-config

mehalter: This does do the extension of the `opts` table here.

mehalter (approved):
LGTM!

PR #482, opened 2023-07-26
chore(main): replace astronvim with astrocore

Uzaaft:


mehalter:
I'm going to go ahead and close this just because v4 is pretty far away so this isn't going to be merged for a while

PR #486, opened 2023-07-29
feat(editing-support): add yanky.nvim

fynnfluegge:
Adds [yanky.nvim](https://github.com/gbprod/yanky.nvim) plugin for improved Improved yank and put functionalities.

### DIFF
--- a/lua/astrocommunity/editing-support/yanky-nvim/init.lua
+++ b/lua/astrocommunity/editing-support/yanky-nvim/init.lua
@@ -1,0 +1,40 @@
+return {
+  "gbprod/yanky.nvim",
+  dependencies = { { "kkharji/sqlite.lua", enabled = not jit.os:find "Windows" } },
+  opts = function()
+    local mapping = require "yanky.telescope.mapping"
+    local mappings = mapping.get_defaults()
+    mappings.i["<c-p>"] = nil
+    return {
+      highlight = { timer = 200 },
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/editing-support/yanky-nvim/init.lua
```diff
@@ -0,0 +1,40 @@
...
+      picker = {
+        telescope = {
+          use_default_mappings = false,
+          mappings = mappings,
+        },
+      },
+    }
+  end,
+  keys = {
+        -- stylua: ignore
```
Uzaaft, line 20: Why do you have this here?

fynnfluegge: Otherwise it will be formatted to multiple lines per keymap. By disabling lua formatting and keep each keymap in one line, the readability is way better imo.

Uzaaft: We have other cases where we let stylua split the line, so I'd say we should keep doing it. Can you remove the ignore line here, and we're GTG 🚀

fynnfluegge: Alright 👌✅

fynnfluegge: Just formats the first line, so nevermind actually 😅

Uzaaft (approved):

PR #489, opened 2023-07-29
feat(editing-support): TreeSJ supports join length of 9999

jay-babu:


Uzaaft:
@jay-babu  Just wondering: Why is this required?

jay-babu:
image

annoying problem to deal with. I asked them why it is 120: https://github.com/Wansmer/treesj/discussions/117

Uzaaft:
Got it. I'm not sure if we should have 9999 in astrocommunity. Perhaps we should wait for an answer from Wansmer, in case having a too big value has some side effects, or performance hit

jay-babu:
> Got it. I'm not sure if we should have 9999 in astrocommunity. Perhaps we should wait for an answer from Wansmer, in case having a too big value has some side effects, or performance hit

they responded. 

```
The length limit for join is needed to prevent unwanted joining when the plugin is not called at the right place (e.g., to avoid joining the whole json root node).
The reason that it is 120 is that it seemed right to me, since there is usually no need to collapse/uncollapse longer nodes.
To disable this behavior, you can simply set this option to any very large number, such as 9999.
```

Uzaaft:
> > Got it. I'm not sure if we should have 9999 in astrocommunity. Perhaps we should wait for an answer from Wansmer, in case having a too big value has some side effects, or performance hit
> 
> they responded.
> 
> ```
> The length limit for join is needed to prevent unwanted joining when the plugin is not called at the right place (e.g., to avoid joining the whole json root node).
> The reason that it is 120 is that it seemed right to me, since there is usually no need to collapse/uncollapse longer nodes.
> To disable this behavior, you can simply set this option to any very large number, such as 9999.
> ```


@jay-babu 
I'm inclined towards keeping the default of the plugin author. Perhaps this should instead be added in the `Readme.md`?

PR #496, opened 2023-08-02
feat(pack): Adding helm pack

Sacro:
Setting up treesitter, language server and vim-helm.

### DIFF
--- a/lua/astrocommunity/pack/helm/README.md
+++ b/lua/astrocommunity/pack/helm/README.md
@@ -1,0 +1,7 @@
+# Helm Language Pack
+
+This plugin pack does the following:
+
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/pack/helm/helmfile.test.yaml
Uzaaft, line 1: Why are you including this file?

Sacro: Because I used it to test and didn't remove it before doing a `git add .`

One moment caller.

Sacro: You saw nothing!

Uzaaft (approved):

PR #510, opened 2023-08-08
chore(bufferline-nvim): add neo-tree offset

fynnfluegge:
## 📑 Description

Since `Neo-tree` is the default file explorer, it would be nice to add the `neo-tree` offset to `bufferline` by default as:

Screenshot 2023-08-08 at 17 06 22

Currently it looks like this:

Screenshot 2023-08-08 at 17 08 38


mehalter (approved):
LGTM!

PR #511, opened 2023-08-08
fix(auto-save-nvim): disable format on save when auto saving

mehalter:
## 📑 Description

Formatting on save with the automatic saving breaks undo because whenever you undo it reformats immediatley. This disables the automatic formatting when buffers are auto saved. It still lets auto formatting work when saving manually with `:w` for example.


owittek (approved):
LGTM

PR #513, opened 2023-08-09
fix(cpp): update clangd_extensions plugin to new setup process

mehalter:
## 📑 Description

`clangd_extensions` recently made some significant breaking changes to change how setup is done. This greatly simplifies the setup process. We just have to make sure the plugin is added before the language servers are set up. This moves it to a dependency of `mason-lspconfig`


mehalter:
@Uzaaft for AstroNvim v4 we should probably just have this fully lazy loaded and then load the plugin with the `clangd` `on_attach` method! 🚀

Uzaaft:
Shouldn't this PR add back clangd, since the plugin itself doesn't do that anymore?

mehalter:
@Uzaaft it does that, it removes the line that skips the setup. `clangd` is already in the mason-lspconfig setup

PR #515, opened 2023-08-10
fix(rust): Lazy load crates.nvim on BufRead Cargo.toml event

Uzaaft:
## 📑 Description

Lazy load `crates.nvim` when Cargo.toml is loaded


owittek (approved):
LGTM

PR #516, opened 2023-08-10
feat(cpp): only load `clangd_extensions` when `clangd` attaches

mehalter:
## 📑 Description

This improves the loading of `clangd_extensions` by making it only load when clangd actually attaches to a buffer rather than just loading it on file types.


A-Lamia (approved):

A-Lamia:
LGTM!

PR #518, opened 2023-08-12
fix noice-nvim warning by disable lsp signature of noice-nvim

t1gu1:
## 📑 Description

I got a warning when I was opening my nvim.
I disable the signature and the warning is gone and I see no difference.

Here the warning
image

### DIFF
--- a/lua/astrocommunity/utility/noice-nvim/init.lua
+++ b/lua/astrocommunity/utility/noice-nvim/init.lua
@@ -22,6 +22,10 @@ return {
           ["vim.lsp.util.stylize_markdown"] = true,
           ["cmp.entry.get_documentation"] = true,
         },
+
+        signature = {
+          enabled = false,
+        },
       },
       presets = {
         bottom_search = true, -- use a classic bottom cmdline for search


Uzaaft:
> ## 📑 Description
> I got a warning when I was opening my nvim. I disable the signature and the warning is gone and I see no difference.
> 
> Here the warning image

This error is most likely due to another plugin.

Uzaaft:
See if you get an error with checkhealth.
But this is most likely a bug with your config that causes this error to happen.

t1gu1:
I do not have error with checkhealth.

Looks like if I disable this astrocommunity plugin: (lsp-signature-nvim)
https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/lsp/lsp-signature-nvim/init.lua
Then I don't get the error with noice-nvim.

But why do you want the lsp enable with noice?

@Uzaaft  What's you proposition to fix this warning between these 2 AstroCommunity plugins?

Uzaaft:
Follow what the error say. Disable one of them or set the variable to false. :)

PR #524, opened 2023-08-17
feat(pack): java. enable niceities with jdtls and downloading source …

jay-babu:
…code from libraries

## 📑 Description

enable niceities with jdtls and downloading source code from libraries

### DIFF
--- a/lua/astrocommunity/pack/java/init.lua
+++ b/lua/astrocommunity/pack/java/init.lua
@@ -81,7 +81,45 @@ return {
         },
         root_dir = root_dir,
         settings = {
-          java = {},
+          java = {
+            eclipse = {
+              downloadSources = true,
+            },
+            configuration = {
+              updateBuildConfiguration = "interactive",
+            },
+            maven = {
+              downloadSources = true,
+            },
+
+            implementationsCodeLens = {
+              enabled = true,
+            },
+            referencesCodeLens = {
+              enabled = true,
+            },
+          },
+          signatureHelp = {
+
+            enabled = true,
+          },
+          completion = {
+            favoriteStaticMembers = {
+              "org.hamcrest.MatcherAssert.assertThat",
+              "org.hamcrest.Matchers.*",
+              "org.hamcrest.CoreMatchers.*",
+              "org.junit.jupiter.api.Assertions.*",
+              "java.util.Objects.requireNonNull",
+              "java.util.Objects.requireNonNullElse",
+              "org.mockito.Mockito.*",
+            },
+          },
+          sources = {
+            organizeImports = {
+              starThreshold = 9999,
+              staticStarThreshold = 9999,
+            },
+          },
         },
         init_options = {
           bundles = {


Uzaaft:
@AstroNvim/astrocommunity-maintainers Anyone of you who uses Java, and thinks this should be added into the pack?
Imo it seems like something that should be enabled by the user, not by the pack.

owittek:
Didn't we have something similar with the rust lsp where we were planning to just add it to the readme? But if you ask me we can just merge it and if anyone complains it can still be removed

Uzaaft:
> Didn't we have something similar with the rust lsp where we were planning to just add it to the readme? But if you ask me we can just merge it and if anyone complains it can still be removed

We added that actually, lsp-settings.
Up to you hehe, if you want to go ahead and merge this one

jay-babu:
@owittek @Uzaaft what will it take to get this merged?

Uzaaft:
> Contributor

My view is that these changes might be considered as **too opinionated** for someone who uses Java, so I'm just letting it stay until one of the more java savy maintainers take a look.

bubuntux:
the star import and static imports seems very opinionated i do agree with it, but i know of a lot of projects/companies that do not agree with it, i would argue that would be better leave it off the pack

jay-babu:
> the star import and static imports seems very opinionated i do agree with it, but i know of a lot of projects/companies that do not agree with it, i would argue that would be better leave it off the pack

Instead of worrying about what other projects/companies would like, can we focus on the astrocommunity and what they want?

PS, I've worked at fairly large companies and they have this rule. I am okay with removing what seems too opinionated, but I ask that this is inline with the astrocommunity wants not what other projects would do. Most of the PR is just enabling features that provide more information. there isn't much else happening. The star imports just ensures [LOD](https://en.wikipedia.org/wiki/Law_of_Demeter), so it shouldn't be an issue

Uzaaft:
> > the star import and static imports seems very opinionated i do agree with it, but i know of a lot of projects/companies that do not agree with it, i would argue that would be better leave it off the pack
> 
> Instead of worrying about what other projects/companies would like, can we focus on the astrocommunity and what they want?
> 
> PS, I've worked at fairly large companies and they have this rule. I am okay with removing what seems too opinionated, but I ask that this is inline with the astrocommunity wants not what other projects would do. Most of the PR is just enabling features that provide more information. there isn't much else happening. The star imports just ensures [LOD](https://en.wikipedia.org/wiki/Law_of_Demeter), so it shouldn't be an issue

Astrocommunity tries to provide a ***sane and sensible*** default. Not sure if this qualifies as that, but I'll let another Java using maintainer be the judge of that.

owittek:
Personally I'm fine with it, people can always check the sources if something seems wrong for them and PR changes or override it in their config

sjcobb2022:
As the person who made this pack, this seems fine if everyone is alright with it. Definitely seem like some sane defaults.

I mean it's all overrideable anyway, as the user can configure it however they like through the opts :)).

owittek:
I'll just go ahead and merge this then, I see the benefit of having sane defaults and there's always the possibility to dispute this decision

owittek (approved):
LGTM!

PR #526, opened 2023-08-20
fix(bufferline-nvim): fix neo-tree offset

fynnfluegge:
## 📑 Description

I noted that the neo-tree offset was not applied after switching to bufferline.nvim from astrocommunity. Seems that I missed the `options` struct when copy paste from my local bufferline config in this PR https://github.com/AstroNvim/astrocommunity/pull/510 😅

### DIFF
--- a/lua/astrocommunity/bars-and-lines/bufferline-nvim/init.lua
+++ b/lua/astrocommunity/bars-and-lines/bufferline-nvim/init.lua
@@ -3,12 +3,14 @@ return {
     "akinsho/bufferline.nvim",
     event = "VeryLazy",
     opts = {
-      offsets = {
-        {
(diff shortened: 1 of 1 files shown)


Uzaaft:
> ## 📑 Description
> I noted that the neo-tree offset was not applied after switching to bufferline.nvim from astrocommunity. Seems that I missed the `options` struct when copy paste from my local bufferline config in this PR #510 😅

Do you have images of before and after?

fynnfluegge:
> Do you have images of before and after?

before:
Screenshot 2023-08-20 at 12 04 56
after:
Screenshot 2023-08-20 at 12 04 42

Uzaaft (approved):

PR #528, opened 2023-08-21
feat(git): add blame.nvim

fynnfluegge:
## 📑 Description

Adds [blame.nvim](https://github.com/FabijanZulj/blame.nvim), a fugitive style git blame visualizer:

Screenshot 2023-08-21 at 12 31 59

### DIFF
--- a/lua/astrocommunity/git/blame-nvim/README.md
+++ b/lua/astrocommunity/git/blame-nvim/README.md
@@ -2,6 +2,6 @@
 
 Neovim fugitive style git blame plugin
 
-**Repository**: https://github.com/FabijanZulj/blame.nvim
+**Repository**: <https://github.com/FabijanZulj/blame.nvim>
 
 Adds git blame visualizer.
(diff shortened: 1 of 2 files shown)


Uzaaft (dismissed):

lua/astrocommunity/git/blame-nvim/README.md
```diff
@@ -0,0 +1,7 @@
+# blame.nvim
+
+Neovim fugitive style git blame plugin
+
+**Repository**: https://github.com/FabijanZulj/blame.nvim
```
Uzaaft, line 5: this url needs to be between < >

luxus:
the benefit of this is compare to the build in one is that we can see all blames and not only for the correct line?

fynnfluegge:
> the benefit of this is compare to the build in one is that we can see all blames and not only for the correct line?

yep, it is just for showing all blames for the whole file in a separate view, toggle by `:ToggleBlame`

PR #535, opened 2023-08-24
perf(cmp-cmdline): lazy load

meijieru:
## 📑 Description

Lazy load `cmp-cmdline`, for faster startup time.

## ℹ Additional Information

Original implementation load `cmp` at startup.

### DIFF
--- a/lua/astrocommunity/completion/cmp-cmdline/init.lua
+++ b/lua/astrocommunity/completion/cmp-cmdline/init.lua
@@ -1,22 +1,28 @@
-local cmp = require "cmp"
-
 return {
   "hrsh7th/cmp-cmdline",
-  lazy = false,
-  opts = {
-    mapping = cmp.mapping.preset.cmdline(),
-    sources = cmp.config.sources({
-      { name = "path" },
-    }, {
-      {
-        name = "cmdline",
-        option = {
-          ignore_cmds = { "Man", "!" },
+  opts = function()
+    local cmp_mapping = require "cmp.config.mapping"
+    local cmp_sources = require "cmp.config.sources"
+
+    return {
+      mapping = cmp_mapping.preset.cmdline(),
+      sources = cmp_sources({
+        { name = "path" },
+      }, {
+        {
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/completion/cmp-cmdline/init.lua
```diff
@@ -1,22 +1,31 @@
...
+    },
+    config = function(_, opts) require("cmp").setup.cmdline(":", opts) end,
+  },
+  {
+    "nvim-cmp",
+    event = { "CmdlineEnter" },
+    dependencies = {
+      "cmp-cmdline",
+    },
   },
```
Uzaaft, line 30: Why do you have this?

meijieru: Otherwise, the `cmp-cmdline` won't get loaded.

Uzaaft: I believe that's the incorrect way to do it.
Instead, you should make `nvim-cmp` a dependency of `cmp-cmdline`, and make `cmp-cmdline` load on CmdLineEnter.

meijieru: fixed.

lua/astrocommunity/completion/cmp-cmdline/init.lua
```diff
@@ -1,22 +1,28 @@
...
+      mapping = cmp_mapping.preset.cmdline(),
+      sources = cmp_sources({
+        { name = "path" },
+      }, {
+        {
+          name = "cmdline",
+          option = {
+            ignore_cmds = { "Man", "!" },
+          },
         },
```
Uzaaft, line 17: Any particular reason why you changed the opts to a function?

meijieru: That's for further lazy loading the `cmp_sources` & `cmp_mappings`.

Uzaaft (approved):

PR #541, opened 2023-08-27
fix(obsidian-nvim): remove annoying notification about "gf" mapping

RayJameson:
## 📑 Description
Remove annoying notification
image

according to the plugin's `help`:

If you want the `gf` passthrough functionality but you’ve already overridden
the `gf` keybinding, just change your `gf` mapping definition to something like
this:

```lua
    vim.keymap.set("n", "gf", function()
      if require("obsidian").util.cursor_on_markdown_link() then
        return "<cmd>ObsidianFollowLink<CR>"
      else
        return "gf"
      end
    end, { noremap = false, expr = true })
```

Then make sure to comment out the `gf` keybinding in your obsidian.nvim config:

```lua
    mappings = {
      -- ["gf"] = require("obsidian.mapping").gf_passthrough(),
    },
```

### DIFF
--- a/lua/astrocommunity/note-taking/obsidian-nvim/README.md
+++ b/lua/astrocommunity/note-taking/obsidian-nvim/README.md
@@ -16,5 +16,8 @@ and
 
 to match your vault location. 
 
+We set `mappings` to an empty table, because `gf` is used for lazy loading in `keys`  
+if you don't want `gf` for lazy loading, you should remove `keys` and `mappings` tables
+
 
 The plugin may also nag and ask you to create a `templates` directory in the vault. You can use `mkdir templates` to create an empty directory.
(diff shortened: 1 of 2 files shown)


Uzaaft:
Could you update the readme file

RayJameson:
Do you want me to add something like "We set `mappings` to an empty table, because `gf` is used for lazy loading in `keys` If you don't want `gf` for lazy loading, you should remove `keys` and `mappings` tables?"

Uzaaft:
Yeah that sounds like a great initial version @RayJameson  ;) 
Appreciate it!

RayJameson:
> Yeah that sounds like a great initial version @RayJameson ;) Appreciate it!

Can you check if it's good?

Uzaaft (approved):

Uzaaft:
> 

I assume you've test ran the plugin code?

RayJameson:
Yes, that's why I made this PR

Uzaaft (approved):

PR #551, opened 2023-09-01
fix(python): Adds Imports Sorting with isort to the Python Pack

WaelKarkoub:
## 📑 Description

It's a common way to sort Python imports on save. `ruff_lsp` could be used to sort the imports the exact same way through code actions, but I'm not sure how to do that on file save.


Uzaaft (changes requested):
Please add isort to the readme as well.

WaelKarkoub:
@Uzaaft just updated the readme :heavy_check_mark:

Uzaaft (approved):

PR #553, opened 2023-09-02
fix(haskell): Update `haskell-tools.nvim` to v2.0.0

Per48edjes:
## 📑 Description

Update the Haskell pack's `init.lua` to reflect the changes made in mrcjkb/haskell-tools.nvim#227. (The changes herein are lifted directly from the plugin creator's suggested `lazy.nvim` config.)

## ℹ Additional Information

- mrcjkb/haskell-tools.nvim#238 prompted an updated (for me) from v1.0.0 to v2.0.0
- This is my first time contributing to this project, so any guidance is most appreciated!
  - *Caveat lector*: As someone who just recently transitioned from `vim` to Neovim, I'm also a newbie when it comes to Lua configs... 😬

### DIFF
--- a/lua/astrocommunity/pack/haskell/init.lua
+++ b/lua/astrocommunity/pack/haskell/init.lua
@@ -21,7 +21,7 @@ return {
       utils.list_insert_unique(astronvim.lsp.skip_setup, "hls")
       vim.g.haskell_tools = {
         hls = {
-          on_attach = function(client, bufnr, ht) 
+          on_attach = function(client, bufnr, ht)
             require("astronvim.utils.lsp").on_attach(client, bufnr)
             ht.dap.discover_configurations(bufnr)
           end,
--- a/lua/astrocommunity/pack/haskell/init.lua
+++ b/lua/astrocommunity/pack/haskell/init.lua
@@ -19,13 +19,8 @@ return {
     branch = "2.x.x", -- Recommended
     init = function() -- Optional, see Advanced configuration
       vim.g.haskell_tools = {
-        tools = {
-          -- ...
-        },
         hls = {
-          ---@param ht is a copy of the haskell-tools module
-          on_attach = function(client, bufnr, ht) end,
-          -- ...
+          on_attach = function(client, bufnr) require("astronvim.utils.lsp").on_attach(client, bufnr) end,
         },
       }
     end,
--- a/lua/astrocommunity/pack/haskell/init.lua
+++ b/lua/astrocommunity/pack/haskell/init.lua
@@ -18,9 +18,13 @@ return {
     },
     branch = "2.x.x", -- Recommended
     init = function() -- Optional, see Advanced configuration
+      utils.list_insert_unique(astronvim.lsp.skip_setup, "hls")
       vim.g.haskell_tools = {
         hls = {
-          on_attach = function(client, bufnr) require("astronvim.utils.lsp").on_attach(client, bufnr) end,
+          on_attach = function(client, bufnr, ht) 
+            require("astronvim.utils.lsp").on_attach(client, bufnr)
+            ht.dap.discover_configurations(bufnr)
+          end,
         },
       }
     end,
--- a/lua/astrocommunity/pack/haskell/init.lua
+++ b/lua/astrocommunity/pack/haskell/init.lua
@@ -12,33 +12,25 @@ return {
   },
   {
     "mrcjkb/haskell-tools.nvim",
-    ft = { "haskell" },
-    branch = "1.x.x", -- recommended by haskell-tools
-    init = function() astronvim.lsp.skip_setup = utils.list_insert_unique(astronvim.lsp.skip_setup, "hls") end,
-    opts = {
-      hls = {
-        on_attach = function(client, bufnr) require("astronvim.utils.lsp").on_attach(client, bufnr) end,
-      },
-    },
-    config = function(_, opts)
-      local tools = require "haskell-tools"
-      vim.api.nvim_create_autocmd("Filetype", {
-        pattern = "haskell", -- autocmd to start haskell-tools
-        callback = function() tools.start_or_attach(opts) end,
-      })
-
-      vim.api.nvim_create_autocmd("LspAttach", {
-        pattern = "*.hs", -- autocmd to start haskell-tools
-        callback = function(args)
-          local client = vim.lsp.get_client_by_id(args.data.client_id)
-          if client.name == "haskell-tools.nvim" then tools.dap.discover_configurations(args.buf) end
-        end,
-      })
-    end,
     dependencies = {
       "nvim-lua/plenary.nvim",
-      "nvim-telescope/telescope.nvim", -- optional
+      "nvim-telescope/telescope.nvim", -- Optional
     },
+    branch = "2.x.x", -- Recommended
+    init = function() -- Optional, see Advanced configuration
+      vim.g.haskell_tools = {
+        tools = {
+          -- ...
+        },
+        hls = {
+          ---@param ht is a copy of the haskell-tools module
+          on_attach = function(client, bufnr, ht) end,
+          -- ...
+        },
+      }
+    end,
+    -- load the plugin when opening one of the following file types
+    ft = { "haskell", "lhaskell", "cabal", "cabalproject" },
   },
   {
     "williamboman/mason-lspconfig.nvim",
--- a/lua/astrocommunity/pack/haskell/init.lua
+++ b/lua/astrocommunity/pack/haskell/init.lua
@@ -14,11 +14,11 @@ return {
     "mrcjkb/haskell-tools.nvim",
     dependencies = {
       "nvim-lua/plenary.nvim",
-      "nvim-telescope/telescope.nvim", -- Optional
+      { "nvim-telescope/telescope.nvim", optional = true }, -- Optional
     },
     branch = "2.x.x", -- Recommended
     init = function() -- Optional, see Advanced configuration
-      utils.list_insert_unique(astronvim.lsp.skip_setup, "hls")
+      astronvim.lsp.skip_setup = utils.list_insert_unique(astronvim.lsp.skip_setup, "hls")
       vim.g.haskell_tools = {
         hls = {
           on_attach = function(client, bufnr, ht)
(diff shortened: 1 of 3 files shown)


lua/astrocommunity/pack/haskell/init.lua
```diff
@@ -12,33 +12,20 @@ return {
...
       "nvim-lua/plenary.nvim",
-      "nvim-telescope/telescope.nvim", -- optional
+      "nvim-telescope/telescope.nvim", -- Optional
     },
+    branch = "2.x.x", -- Recommended
+    init = function() -- Optional, see Advanced configuration
+      vim.g.haskell_tools = {
+        hls = {
+          on_attach = function(client, bufnr) require("astronvim.utils.lsp").on_attach(client, bufnr) end,
+        },
```
mrcjkb, line 24: Since the `LspAttach` autocommand that adds the nvim-dap configurations has been deleted, you could put it in here (the `on_attach` function passes in a `ht` (`HaskellTools` object):

```suggestion
          on_attach = function(client, bufnr, ht) 
            require("astronvim.utils.lsp").on_attach(client, bufnr)
            ht.dap.discover_configurations(bufnr)
          end,
        },
```

Note: If it's possible for users to uninstall nvim-dap, then you might want to check if nvim-dap is installed first:

```lua
on_attach = function(client, bufnr, ht) 
  require("astronvim.utils.lsp").on_attach(client, bufnr)
  local has_dap, _ = pcall(require, 'dap')
  if has_dap then
    ht.dap.discover_configurations(bufnr)
  end
end,
```

lua/astrocommunity/pack/haskell/init.lua
```diff
@@ -12,33 +12,20 @@ return {
...
-    end,
     dependencies = {
       "nvim-lua/plenary.nvim",
-      "nvim-telescope/telescope.nvim", -- optional
+      "nvim-telescope/telescope.nvim", -- Optional
     },
+    branch = "2.x.x", -- Recommended
+    init = function() -- Optional, see Advanced configuration
+      vim.g.haskell_tools = {
+        hls = {
```
mrcjkb, line 22: This looks like it should stay in the `init` function:

```lua
utils.list_insert_unique(astronvim.lsp.skip_setup, "hls")
```

```suggestion
    init = function() -- Optional, see Advanced configuration
      utils.list_insert_unique(astronvim.lsp.skip_setup, "hls")
      vim.g.haskell_tools = {
        hls = {
```

lua/astrocommunity/pack/haskell/init.lua
```diff
@@ -12,33 +12,20 @@ return {
...
-        callback = function(args)
-          local client = vim.lsp.get_client_by_id(args.data.client_id)
-          if client.name == "haskell-tools.nvim" then tools.dap.discover_configurations(args.buf) end
-        end,
-      })
-    end,
     dependencies = {
       "nvim-lua/plenary.nvim",
-      "nvim-telescope/telescope.nvim", -- optional
+      "nvim-telescope/telescope.nvim", -- Optional
```
mrcjkb, line 17: Not directly related to this PR, as it was already there,
but I would suggest not adding the telescope dependency here.
If a user wants telescope uninstalled for some reason, they might be annoyed if it's automatically installed because it's listed as a dependency for haskell-tools.

mrcjkb (changes requested):

mrcjkb (approved):

mrcjkb:
Nit: I would suggest changing the PR title's prefix to `chore(deps):`, as this is not really a bug fix.

Per48edjes:
> Nit: I would suggest changing the PR title's prefix to `chore(deps):`, as this is not really a bug fix.

Fair enough; will update the type to `chore` but I think this project wants the [scope to be the existing enclosing folder](https://github.com/AstroNvim/astrocommunity/blob/177f8f4dd0cf8e0e00c52f651e88dc5b7dcde63c/CONTRIBUTING.md?plain=1#L68).

lua/astrocommunity/pack/haskell/init.lua
Uzaaft:
```lua
{
"nvim-telescope/telescope.nvim",
optional = true
}
```
Solves this issue. :)
Reference: https://github.com/folke/lazy.nvim#-plugin-spec

Uzaaft:
@Per48edjes  Your first PR title was fine. :)

lua/astrocommunity/pack/haskell/init.lua
mrcjkb: Oh, nice. I'll update the snippet in the readme.

lua/astrocommunity/pack/haskell/init.lua
```diff
@@ -12,33 +12,24 @@ return {
...
     },
+    branch = "2.x.x", -- Recommended
+    init = function() -- Optional, see Advanced configuration
+      utils.list_insert_unique(astronvim.lsp.skip_setup, "hls")
```
Uzaaft, line 21: I don't think `utils.list_insert_unique` is in place. See the similar code in typescript pack.

Uzaaft (changes requested):
1. Add optional to telescope.nvim
2. I don't think `utils.list_insert_unique(astronvim.lsp.skip_setup, "hls")` is inplace

lua/astrocommunity/pack/haskell/init.lua
mrcjkb: Yup. This should be

```lua
astronvim.lsp.skip_setup = utils.list_insert_unique(astronvim.lsp.skip_setup, "hls")
```

(Was missing in the snippet in my previous comment for some reason)

lua/astrocommunity/pack/haskell/init.lua
```diff
@@ -12,33 +12,24 @@ return {
...
+      vim.g.haskell_tools = {
```
mrcjkb, line 22: I just had a thought: I don't know if AstroNvim packages are designed to be user configurable.
It seems like they are, as `opts` can be a function that takes an `opts` parameter, which I'm assuming can be influenced by the user? (See for example the [java pack](https://github.com/AstroNvim/astrocommunity/blob/bb528993675119e4392acd1aea2b1691d5791005/lua/astrocommunity/pack/java/init.lua#L36)).

If this is the case, then I would suggest putting this line in the `init` function:

```lua
astronvim.lsp.skip_setup = utils.list_insert_unique(astronvim.lsp.skip_setup, "hls")
```

And setting `vim.g.haskell_tools` in the `config` function (merging the default with the `opts` passed to `config`).
I had recommended to set `vim.g.haskell_tools` in the `init` function, 
because lazy.nvim doesn't use Neovim's built-in initialisation order and I wanted to be on the safe side, setting `vim.g.haskell_tools` as early as possible). But iin principle it can also be set in the `config` function.

Note: As the previous package wasn't user-configurable, please don't see this as a blocker for this PR.

Uzaaft: It can, I don't think it should.

mrcjkb: @Uzaaft I presume you're referring to the fact that `vim.g.haskell_tools` can go in `config`?

What's your reasoning?

Uzaaft: All examples I've seen from folke does `vim.g` stuff in `init`

mrcjkb: That makes sense for plugins that have their Initialization in the `plugin` directory (see `:h runtimepath`), which is the majority of Vimscript plugins.

This is not the case for `haskell-tools`, which has its initialization in `ftplugin`.

mrcjkb, line 22: Another option (to keep user-configurability) would be:

```suggestion
      vim.g.haskell_tools = vim.g.haskell_tools or {
```

Per48edjes:
> 1. Add optional to telescope.nvim
> 2. I don't think `utils.list_insert_unique(astronvim.lsp.skip_setup, "hls")` is inplace. It should be
>    `astronvim.lsp.skip_setup = utils.list_insert_unique(astronvim.lsp.skip_setup, "hls")`

Incorporated these in ead0450! Happy to continue moving this forward -- would need some direction on whether this is good as is or whether we're trying to throw `vim.g` in `config`.

mrcjkb:
> > 1. Add optional to telescope.nvim
> > 2. I don't think `utils.list_insert_unique(astronvim.lsp.skip_setup, "hls")` is inplace. It should be
> >    `astronvim.lsp.skip_setup = utils.list_insert_unique(astronvim.lsp.skip_setup, "hls")`
> 
> Incorporated these in [ead0450](https://github.com/AstroNvim/astrocommunity/commit/ead04502aa20dc7ee35ec45b5bfea4cc0e742c76)! Happy to continue moving this forward -- would need some direction on whether this is good as is or whether we're trying to throw `vim.g` in `config`.

I don't use AstroNvim, so I don't feel strongly about this.
My recommendation would be to either one of

- set `vim.g.haskell_tools` in `config` (by merging a default with the `opts` argument - this is fine for a lazily initialized filetype plugin)
-  or incorporate my second suggestion `vim.g.haskell_tools = vim.g.haskell_tools or { ... }` in `init`, so that users can override it if they want.

On a related note: haskell-tools will soon auto-discover DAP configurations by default if `nvim-dap` is installed and a `haskell-debug-adapter` executable is detected, see: https://github.com/mrcjkb/haskell-tools.nvim/pull/249
So the call to `ht.dap.discover_configurations` will soon be redundant.

Uzaaft:
> 

Appreciate the heads up on the v2.1 launch @mrcjkb

Uzaaft:
I'll leave the last word to @mrcjkb  He knows his plugin well enough. Any thoughts about where the vim.g code should go, etc?

mrcjkb:
> I'll leave the last word to @mrcjkb He knows his plugin well enough. Any thoughts about where the vim.g code should go, etc?

Let's merge this as is for now and I'll open a separate PR after the 2.1 haskell-tools release 😄

Uzaaft:
> > I'll leave the last word to @mrcjkb He knows his plugin well enough. Any thoughts about where the vim.g code should go, etc?
> 
> Let's merge this as is for now and I'll open a separate PR after the 2.1 haskell-tools release 😄

Looking forward to it!

Uzaaft (approved):

Per48edjes:
Thanks for all the help @mrcjkb & @Uzaaft ! 🙏

PR #555, opened 2023-09-03
fix(yanky-nvim): Fix yanky.nvim loading when restoring session

chaozwn:
there can not set event 'User AstroFile', because restore session from resession.nvim will not tigger 'User AstroFile' event.

### DIFF
--- a/lua/astrocommunity/editing-support/yanky-nvim/init.lua
+++ b/lua/astrocommunity/editing-support/yanky-nvim/init.lua
@@ -1,5 +1,6 @@
 return {
   "gbprod/yanky.nvim",
+  event = "UIEnter",
   dependencies = { { "kkharji/sqlite.lua", enabled = not jit.os:find "Windows" } },
   opts = function()
     local mapping = require "yanky.telescope.mapping"


Uzaaft:
Updated the title.
@chaozwn  I don't really understand the initial "bug". Was the issue caused by "User AstroFile"?

chaozwn:
> Updated the title. @chaozwn I don't really understand the initial "bug". Was the issue caused by "User AstroFile"?



> Updated the title. @chaozwn I don't really understand the initial "bug". Was the issue caused by "User AstroFile"?

This plugin overwrites our y key, but many times it starts up as not loaded, causing the y key to be unusable. Therefore, we need to advance its startup time.

Uzaaft:
Can't you load it after UIEnter, i.e on the event: `LazyVimStarted`?

chaozwn:
> 

After changing to LazyVimStarted, when using nvim to restore the previous page, yy becomes invalid, so we can't use this event.

Uzaaft:
> > 
> 
> After changing to `LazyVimStarted`, when using `nvim` to restore the previous page, `yy` becomes invalid, so we can't use this event.

Got it thanks for checking that out!

Uzaaft (approved):

PR #558, opened 2023-09-08
fix(typescript): fix eslint_d not looking for .eslintrc file

Dartv:
## 📑 Description
Adds support for .eslintrc config file. Otherwise eslint_d is not initialized for projects that use the file.


Uzaaft:
Based upon: https://eslint.org/docs/latest/use/configure/configuration-files, `.eslintrc` is not a file name it supports.

Dartv:
Interesting, I have a project with .eslintrc file and this PR fixes the issue for me.
Changing the file on a large project with lots of people may raise questions. That's why I'd rather have a fix at a plugin level

Uzaaft:
I’d still need a source for the `.eslintrc` file which States that eslint looks after that file before I merge this

Uzaaft:
This pr might fix the issue with eslint not spawning. But does it use the settings from the file?

Dartv:
Yes, it works with .eslintrc config file. I just created a new project with minimal repro.

Dartv:
Eslint tests use .eslintrc file if you need more proof https://github.com/eslint/eslint/blob/cac45d04b890b0700dd8908927300608adad05fe/tests/fixtures/config-extends/.eslintrc#L4

Uzaaft:
> Eslint tests use .eslintrc file if you need more proof [eslint/eslint@`cac45d0`/tests/fixtures/config-extends/.eslintrc#L4](https://github.com/eslint/eslint/blob/cac45d04b890b0700dd8908927300608adad05fe/tests/fixtures/config-extends/.eslintrc#L4)

You've misunderstood that. Take a look at the package.json file.

Uzaaft:
> Yes, it works with .eslintrc config file. I just created a new project with minimal repro. 

I got that it work. But I'm hesitant to add something that's not mentioned in the official docs

owittek:
It's likely that using `.eslintrc` without a file ending is a legacy thing, pretty sure I've seen it before. I honestly don't see a reason to not merge this as I don't think people just have random files with that name lying in their project folder

Uzaaft:
> It's likely that using `.eslintrc` without a file ending is a legacy thing, pretty sure I've seen it before. I honestly don't see a reason to not merge this as I don't think people just have random files with that name lying in their project folder

Even if it's legacy? I would've just left it to the users to add it to their configs instead.

owittek:
what's the downside of just adding it? I think the upside is that there won't be any issues/PRs for this exact issue in the future as legacy projects still account for a big percentage of total projects so it's not unlikely to run into this again

owittek:
to add onto this I'd personally even be fine with a regex match of that sort: `^\.eslintrc(\.\w+)?$`

taskylizard:
> It's likely that using `.eslintrc` without a file ending is a legacy thing, pretty sure I've seen it before.

Just wanted to chime in and [say that it is](https://github.com/eslint/eslintrc), and want to stand with merging of this PR.

We should also cover the new [Flat Config](https://eslint.org/docs/latest/use/configure/configuration-files-new) in v9 if it isn't covered already.

Uzaaft:
> > It's likely that using `.eslintrc` without a file ending is a legacy thing, pretty sure I've seen it before.
> 
> Just wanted to chime in and [say that it is](https://github.com/eslint/eslintrc), and want to stand with merging of this PR.
> 
> We should also cover the new [Flat Config](https://eslint.org/docs/latest/use/configure/configuration-files-new) in v9 if it isn't covered already.

Feel free to open a PR. :)

taskylizard:
I was hoping that can be done in this PR anyway since it's just one file format: `eslint.config.js`

Uzaaft:
> I was hoping that can be done in this PR anyway since it's just one file format: `eslint.config.js`

We prefer that PR's do one thing and one thing only. This one adds support for the legacy .eslintrc file, and that other PR will add support for another file. :)

Dartv:
@Uzaaft sorry, I'm not sure I follow. What thing do I need to fix?

Uzaaft:
I fixed it myself :) Opened a new PR to fix this issue.

PR #561, opened 2023-09-08
fix(vim-matchup): remove matchup_match_paren_deferred option

kaka-ruto:
This will leave the option set to 0 by default

## ℹ Additional Information

See demo video of the issue on https://github.com/AstroNvim/AstroNvim/issues/2276


mehalter (approved):
LGTM

PR #564, opened 2023-09-09
feat(pack): update Zig url

NTBBloodbath:
## 📑 Description

I'm slowly moving away from GitHub and so I will archive all my Neovim-related code here. Updated the `zig-tools.nvim` plugin links to point the new maintained url


owittek (approved):
Thanks for the update!

PR #571, opened 2023-09-10
fix(haskell): make haskell-tools respect the `options` table

mrcjkb:
## 📑 Description

This PR is stacked on top of #570, which should be merged first.

- As discussed in https://github.com/AstroNvim/astrocommunity/pull/553#issuecomment-1705684907, the haskell-tools plugin is not user-configurable
- This PR uses [the same method as the java package](https://github.com/mrcjkb/astrocommunity/blob/82e76559973205eb9c8bd76ee134d3bccdee75b6/lua/astrocommunity/pack/java/init.lua#L36) to allow users to configure the plugin through an `opts` table in `community.lua`.

As mentioned in https://github.com/AstroNvim/astrocommunity/pull/553, I don't feel strongly about this, so please close this PR if you don't agree with this approach.

### DIFF
--- a/lua/astrocommunity/pack/haskell/init.lua
+++ b/lua/astrocommunity/pack/haskell/init.lua
@@ -14,16 +14,28 @@ return {
     "mrcjkb/haskell-tools.nvim",
     dependencies = {
       "nvim-lua/plenary.nvim",
-      { "nvim-telescope/telescope.nvim", optional = true }, -- Optional
+      { "nvim-telescope/telescope.nvim", optional = true },
+      { "mfussenegger/nvim-dap", optional = true },
     },
     branch = "2.x.x", -- Recommended
-    init = function() -- Optional, see Advanced configuration
-      astronvim.lsp.skip_setup = utils.list_insert_unique(astronvim.lsp.skip_setup, "hls")
-      vim.g.haskell_tools = {
+    init = function() astronvim.lsp.skip_setup = utils.list_insert_unique(astronvim.lsp.skip_setup, "hls") end,
+    opts = function(_, opts)
+      -- this allows users to pass opts through an opts table in community.lua
+      opts = opts or {}
+      local defaults = {
         hls = {
           on_attach = function(client, bufnr, _) require("astronvim.utils.lsp").on_attach(client, bufnr) end,
         },
       }
+      opts = vim.tbl_deep_extend("keep", opts, defaults)
+      return opts
+    end,
+    config = function(_, opts)
+      -- haskell-tools reads this config the first time `require('haskell-tools')` is called,
+      -- which is typically when opening a Haskell or Cabal file, or when the
+      -- user executes one of the plugin's commands.
+      -- NOTE: If vim.g.haskell_tools has already been set, its options take precedence.
+      vim.g.haskell_tools = vim.tbl_deep_extend("keep", vim.g.haskell_tools or {}, opts)
     end,
     -- load the plugin when opening one of the following file types
     ft = { "haskell", "lhaskell", "cabal", "cabalproject" },
--- a/lua/astrocommunity/pack/haskell/init.lua
+++ b/lua/astrocommunity/pack/haskell/init.lua
@@ -21,7 +21,6 @@ return {
     init = function() astronvim.lsp.skip_setup = utils.list_insert_unique(astronvim.lsp.skip_setup, "hls") end,
     opts = function(_, opts)
       -- this allows users to pass opts through an opts table in community.lua
-      opts = opts or {}
       local defaults = {
         hls = {
           on_attach = function(client, bufnr, _) require("astronvim.utils.lsp").on_attach(client, bufnr) end,
--- a/lua/astrocommunity/pack/haskell/init.lua
+++ b/lua/astrocommunity/pack/haskell/init.lua
@@ -17,26 +17,17 @@ return {
       { "nvim-telescope/telescope.nvim", optional = true },
       { "mfussenegger/nvim-dap", optional = true },
     },
-    branch = "2.x.x", -- Recommended
-    init = function() astronvim.lsp.skip_setup = utils.list_insert_unique(astronvim.lsp.skip_setup, "hls") end,
-    opts = function(_, opts)
-      -- this allows users to pass opts through an opts table in community.lua
-      local defaults = {
+    version = "^2",
+    -- load the plugin when opening one of the following file types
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/pack/haskell/init.lua
```diff
@@ -14,19 +14,27 @@ return {
...
         },
       }
+      opts = vim.tbl_deep_extend("keep", opts, defaults)
+    end,
+    config = function(_, opts)
+      -- haskell-tools reads this config the first time `require('haskell-tools')` is called,
+      -- which is typically when opening a Haskell or Cabal file, or when the
+      -- user executes one of the plugin's commands.
+      -- NOTE: If vim.g.haskell_tools has already been set, its options take precedence.
+      vim.g.haskell_tools = vim.tbl_deep_extend("keep", vim.g.haskell_tools or {}, opts)
```
mrcjkb, line 37: Not sure if it would be cleaner, but I guess you could also set `vim.g.haskell_tools` in the `opts` function instead?

lua/astrocommunity/pack/haskell/init.lua
```diff
@@ -14,19 +14,27 @@ return {
...
+      local defaults = {
         hls = {
-          on_attach = function(client, bufnr, ht)
-            require("astronvim.utils.lsp").on_attach(client, bufnr)
-            ht.dap.discover_configurations(bufnr)
-          end,
+          on_attach = function(client, bufnr, _) require("astronvim.utils.lsp").on_attach(client, bufnr) end,
         },
       }
+      opts = vim.tbl_deep_extend("keep", opts, defaults)
```
mrcjkb, line 30:
```suggestion
      opts = vim.tbl_deep_extend("keep", opts, defaults)
      return opts
```

lua/astrocommunity/pack/haskell/init.lua
```diff
@@ -14,19 +14,28 @@ return {
...
+      { "mfussenegger/nvim-dap", optional = true },
     },
     branch = "2.x.x", -- Recommended
-    init = function() -- Optional, see Advanced configuration
-      astronvim.lsp.skip_setup = utils.list_insert_unique(astronvim.lsp.skip_setup, "hls")
-      vim.g.haskell_tools = {
+    init = function() astronvim.lsp.skip_setup = utils.list_insert_unique(astronvim.lsp.skip_setup, "hls") end,
+    opts = function(_, opts)
+      -- this allows users to pass opts through an opts table in community.lua
+      opts = opts or {}
```
mehalter, line 24:
```suggestion
```

This is unnecessary. `opts` here is always a table with lazy's architecture.

lua/astrocommunity/pack/haskell/init.lua
```diff
@@ -14,19 +14,28 @@ return {
...
         hls = {
-          on_attach = function(client, bufnr, ht)
-            require("astronvim.utils.lsp").on_attach(client, bufnr)
-            ht.dap.discover_configurations(bufnr)
-          end,
+          on_attach = function(client, bufnr, _) require("astronvim.utils.lsp").on_attach(client, bufnr) end,
         },
       }
+      opts = vim.tbl_deep_extend("keep", opts, defaults)
+      return opts
```
mehalter, line 31:
```suggestion
      return vim.tbl_deep_extend("keep", opts, defaults)
```

lua/astrocommunity/pack/haskell/init.lua
mehalter: setting it here is correct. Moving it to the `opts` function will break the user extensibility

mehalter (changes requested):
small minor change, but seems alright. 

*Note:* This does add a weird level of abstraction that isn't accounted for in the `haskell-tools` documentation which could be something that will confuse users, but I'll leave that decision up to the community since core AstroNvim doesn't maintain these packages. As an alternative, with the original approach the user can use the `config` function to modify the default `vim.g.haskell_tools` variable.

mrcjkb:
> This does add a weird level of abstraction that isn't accounted for in the haskell-tools documentation which could be something that will confuse users

That's a good point...

> As an alternative, with the original approach the user can use the config function to modify the default vim.g.haskell_tools variable.

I'm not really familiar with lazy.nvim. Wouldn't it be possible for users to override the `config` function with this approach?

mehalter:
@mrcjkb Yes they can, but it it will override the original `config` function. Lazy will extend the `opts` table but override the `config` function. So the user will have to specifically remember to set the `vim.g.haskell_tools` variable with the `opts` that are passed in rather than knowing that it is already set.

mrcjkb:
Oh, I see. That does seem a bit unergonomic...

mehalter:
Maybe what we should do @mrcjkb is just keep it how it was with the `init` function setting it but using `tbl_deep_extend` on the current value of `vim.g.haskel_tools` to keep what's already set. Then the user can configure it through the `config` function if they want or they could just configure it through the `options` table like normal. That's a bit more of a normal way to approach it

mrcjkb:
Although... If I understand correctly, this means that they would also have to remember to call `setup`, if overriding the `config` function for other plugins?

mrcjkb:
> Maybe what we should do @mrcjkb is just keep it how it was with the init function setting it but using tbl_deep_extend on the current value of vim.g.haskel_tools to keep what's already set. 

You mean to set the default in `init` and then deep extend in `config`?

Sounds fine to me.

mehalter:
No @mrcjkb , I sent a config that illustrates what I said. I think configuring the options through the `opts` table may seem obvious to some users, but it very much goes against the docs and sorta blends a couple different ideas together and could make things more confusing as to why it does the things it does.

mehalter:
also instead of `branch = "2.x.x"`, it probably makes more sense to do `verion = "^2"`. This aligns more with lazy methodology and is a bit cleaner

mrcjkb:
> No @mrcjkb , I sent a config that illustrates what I said. I think configuring the options through the `opts` table may seem obvious to some users, but it very much goes against the docs and sorta blends a couple different ideas together and could make things more confusing as to why it does the things it does.

Lgtm. I'll update later today when I'm back home.

mehalter:
ok cool no worries, I can actually make the push now and get this merged in. Thanks for brainstorming through this with me!

mrcjkb:
> ok cool no worries, I can actually make the push now and get this merged in. Thanks for brainstorming through this with me!

Thanks 🙏

PR #578, opened 2023-09-15
docs(main): Don't use lazy import shorthand in docs

Uzaaft:

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ return {
   },
   { import = "astrocommunity.completion.copilot-lua" },
   { -- further customize the options set by the community
-    "copilot.lua",
+    "zbirenbaum/copilot.lua",
     opts = {
       suggestion = {
         keymap = {


README.md
```diff
@@ -218,7 +218,10 @@ return {
 }
 ```
 
-Note that you can disable imports by setting the enabled option to false. Also, you have two options to modify an existing plugin: you can use the full repository name, like "m4xshen/smartcolumn.nvim", or the module name, like "catppuccin". Keep in mind that the module name may differ from the folder name.
+**Note:** 
+- You can disable imports by setting the enabled option to false. Also, you have two options to modify an existing plugin: you can use the full repository name, like "m4xshen/smartcolumn.nvim", or the module name, like "catppuccin". Keep in mind that the module name may differ from the folder name.
+- We advice you to avoid using shorthands whenever importing a plugin to further customize it. Use the full import path.
```
rafaelderolez, line 223:
```suggestion
- We advise you to avoid using shorthands whenever importing a plugin to further customize it. Use the full import path.
```

mehalter (approved):
LGTM!

PR #581, opened 2023-09-22
chore(completion): de-duplicate copilot logic

jay-babu:
## 📑 Description

Import copilot from another place to reduce duplicate logic.


Uzaaft:
Correct me if I'm wrong @mehalter  but I think we want to keep the plugin specs as seperate as possible, no?

mehalter:
Idk I'm good with this

Uzaaft (approved):

PR #589, opened 2023-09-26
fix(hardtime-nvim): call enable() explicitly, remove redundant keys

slowsage:
## 📑 Description

`hardtime` hooks to `BufEnter` to defer enabling when setup is called with the plugin enabled. On occasion, `User AstroFile` gets called after BufEnter.

This makes the plugin wait for a `BufEnter` that never happens. This PR calls enable() and setup() explicitly with the opts.

## ℹ Additional Information
This PR also removes redundant keys that are present in the plugin [already](https://github.com/m4xshen/hardtime.nvim/blob/main/lua/hardtime/config.lua).


slowsage:
> ### Review Checklist
> 
> Does this PR follow the [Contribution Guidelines](development guidelines)? Following is a _partial_ checklist:
> 
> Proper conventional commit scoping:
> 
>     * If you are adding a new plugin, the scope would be the name of the category it is being added into. ex. `feat(utility): added noice.nvim plugin`
> 
>     * If you are modifying a pre-existing plugin or pack, the scope would be the name of the plugin folder. ex. `fix(noice-nvim): fix LSP handler error`
> 
>  - [x]  Pull request title has the appropriate conventional commit type and **scope** where the scope is the name of the _pre-existing_ directory in the project as described above
> 
> -  [x]  `README` is properly formatted and uses fenced in links with `<url>` unless they are inside a `[title](url)`
> 
> - [ ]  Proper usage of `opts` table rather than setting things up with the `config` function.

Note on the last checkbox. I reused opts (after removing redundant keys) but I have to call config to call setup. I left the opts variable as is as opposed to making opts a function since we need an opts variable anyway for the remaining options.

Uzaaft:
Sorry for the delay, I’ll be merging this within the next 12h.

Uzaaft (approved):

PR #603, opened 2023-10-11
feat(pack)!: Split python pack into python and python-ruff

### LINKED ISSUE #592: Python pack contains both ruff and isort. Is this necessary
Since ruff includes isort functionality, is it necessary to also include the isort module in the pack?

This seems redundant.

Uzaaft:
Closes #592

### DIFF
--- a/lua/astrocommunity/pack/python-ruff/README.md
+++ b/lua/astrocommunity/pack/python-ruff/README.md
@@ -1,12 +1,11 @@
-# Python Language Pack
+# Python Ruff Language Pack
 
 **Requirements:** `python` must be in your `PATH` and executable
 
 This plugin pack does the following:
 
 - Adds `python` and `toml` Treesitter parsers
-- Adds `pyright` language servers
+- Adds `pyright` and `ruff_lsp` language servers
 - Adds the following `null-ls` sources:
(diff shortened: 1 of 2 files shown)


Uzaaft:
This is tentatively ready to come out of draft. @AstroNvim/astrocommunity-maintainers Wanna review?

mehalter (changes requested):
Shouldn't the `python-ruff` pack just import the python pack and then add `ruff-lsp`? Wouldn't reduce lots of duplicate code

Uzaaft:
> Shouldn't the `python-ruff` pack just import the python pack and then add `ruff-lsp`? Wouldn't reduce lots of duplicate code

I could do that :)

m-rutter:
The primary point in splitting it is not because people don't want Ruff (though that is a faction), but because isort is made redundant by ruff and because the current config somehow messes up the formatting of imports.

Uzaaft:
> The primary point in splitting it is not because people don't want Ruff (though that is a faction), but because isort is made redundant by ruff and because the current config somehow messes up the formatting of imports when they are both present (formatting in the wrong order potentially)

Thanks for re-iterating my point.

mehalter (approved):
LGTM

PR #607, opened 2023-10-14
fix(smartcolumn-nvim): Add sensible default opts for when not to show colorcolumn

menoua:
## 📑 Description

Adds two filetypes (alpha, neo-tree) to the "smartcolumn" plugin's default list of disabled_filetypes (help, text, markdown). As both alpha-nvim and Neo-tree are included in the AstroNvim core and I see no reason why anyone would want a column in either case, I think it's a good idea to add them to the list of disabled filetypes.

## ℹ Additional Information

This is the current behavior:
Screenshot 2023-10-14 at 7 47 52 PM

This is after the change:
Screenshot 2023-10-14 at 7 46 59 PM

### DIFF
--- a/lua/astrocommunity/bars-and-lines/smartcolumn-nvim/init.lua
+++ b/lua/astrocommunity/bars-and-lines/smartcolumn-nvim/init.lua
@@ -2,6 +2,6 @@ return {
   "m4xshen/smartcolumn.nvim",
   event = { "InsertEnter", "User AstroFile" },
   opts = {
-    disabled_filetypes = { "neo-tree", "alpha", "help", "text", "markdown" },
+    disabled_filetypes = { "alpha", "neo-tree", "starter", "help", "text", "markdown" },
   },
 }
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/bars-and-lines/smartcolumn-nvim/init.lua
```diff
@@ -1,5 +1,7 @@
 return {
   "m4xshen/smartcolumn.nvim",
   event = { "InsertEnter", "User AstroFile" },
-  opts = {},
+  opts = {
+    disabled_filetypes = { "neo-tree", "alpha", "help", "text", "markdown" },
```
Uzaaft, line 5: Could you add starter as a filetype?

Uzaaft (changes requested):

lua/astrocommunity/bars-and-lines/smartcolumn-nvim/init.lua
menoua: Done.

Uzaaft (approved):

PR #611, opened 2023-10-17
fix(transparent-nvim): Made cursor line transparent

Infonautica:
## 📑 Description

Adjusted groups to include cursor line, so it can be transparent as well

**Before:**
Note the line where cursor is
before-bg

**After:**
after-bg

## ℹ Additional Information

PR has a couple of unwanted commits (at first i didn't know this package exists already), so i can re-create this PR from scratch if that's a problem


mehalter (approved):
LGTM! Thanks for the contribution 🥳

PR #614, opened 2023-10-18
fix(oil-nvim): ssh support

meijieru:
## 📑 Description

`oil.nvim` is equipped to handle `oil-ssh://`, as demonstrated [here](https://github.com/stevearc/oil.nvim#ssh).
However, attempting to lazy load it may result in an error when using `nvim oil-ssh//xxx`.

### DIFF
--- a/lua/astrocommunity/file-explorer/oil-nvim/init.lua
+++ b/lua/astrocommunity/file-explorer/oil-nvim/init.lua
@@ -1,5 +1,6 @@
 return {
   "stevearc/oil.nvim",
+  lazy = false,
   opts = {},
   enabled = true,
   cmd = "Oil",


lua/astrocommunity/file-explorer/oil-nvim/init.lua
```diff
@@ -1,5 +1,6 @@
 return {
   "stevearc/oil.nvim",
+  lazy = not vim.startswith(vim.fn.argv()[1] or "", "oil-ssh://"),
```
Uzaaft, line 3: I don't understand why this is here. You're checking of neovim was opened with a file, but `User AstroFile` does that itself. If you want to load the plugin if the filename is empty, I'd just swap the event out instead of having this lazy = vim.fn.... thing, this just makes things more convoluted.
oil-ssh:// etc should trigger `User AstroFile`

meijieru: `User AstroFile` is after `VimEnter` or `UIEnter`.
The same for another pr.

Uzaaft: Still doesnt explain the usage. Why not just use those events instead?

meijieru: If `VimEnter` or `UIEnter` is used, there is no advantage gained from lazy loading. Therefore, the approach I've identified is to implement lazy loading correctly by using `User AstroFile` for actual files, and disabling lazy loading when dealing with `oil-ssh://`.

Uzaaft: I’d rather use those events, than the conveluted Logic provided here. Or disable lazy loading if its in the way

Uzaaft: @meijieru  Also applies to your other PR

meijieru: I agree it's a bit complicated. I learn it from [here](https://github.com/kawre/leetcode.nvim#-recipes). Feel free to close those pr.

Uzaaft: I think we should just disable lazy loading and call it a day if lazy loading MIGHT cause an issue :)

meijieru: done

luxus:
i wonder how many people are using this. i guess its the minority and I'm not sure if it makes sense to sacrifify lazyloading on the plugin.
the official installation doesn't have lazy disabled as well.

Uzaaft:
> i wonder how many people are using this. i guess its the minority and I'm not sure if it makes sense to sacrifify lazyloading on the plugin. the official installation doesn't have lazy disabled as well.


The config of the oil.nvim author doesn't lazy load oil.nvim, so I think we shouldn't.
@meijieru  feel free to change this PR so that the only diff is that `lazy = false`

meijieru:
> > i wonder how many people are using this. i guess its the minority and I'm not sure if it makes sense to sacrifify lazyloading on the plugin. the official installation doesn't have lazy disabled as well.
> 
> The config of the oil.nvim author doesn't lazy load oil.nvim, so I think we shouldn't. @meijieru feel free to change this PR so that the only diff is that `lazy = false`

Done.

Uzaaft (approved):

PR #621, opened 2023-10-23
fix(dart): Add missing require

### LINKED ISSUE #468: Dart pack is interfering with rust_analyzer from loading.
I have multiple language packs enabled and suddenly rust_anlyzer isn't loading anymore as long as the dart pack is present. :LspInfo is showing rust_analyzer is loaded and attached to the current rust file buffer.

Uzaaft:
While attempting to fix #468, I came over this issue in the dart pack where the astronvim dependency is just used without any imports. This fixes that.


mehalter (commented):
This is not correct

PR #631, opened 2023-11-08
feat(main)!: Astrocommunity compatible with v4

### LINKED ISSUE #463: lsp.inc-rename-nvim not working
`lsp.inc-rename-nvim` is not working with astrocommunity. `vim.lsp.buf.rename()` is called instead.
My workaround is setting the mapping to `<leader>rn`, since using `<leader>lr` does not work.
There is the following lsp_mapping in `.nvim/lua/astronvim/utils/lsp.lua`
Screenshot 2023-07-19 at 20 20 19
`inc-rename-nvim` does not override that mapping.

I see two possible solutions:
1. map `inc-rename-nvim` to `<leader>rn`
2. disable `vim.lsp.buf.rename()` mapping in the `init.lua` of `inc-rename-nvim`
 
Let me know which solution is better and I will raise PR 🙏

Uzaaft:
# Changelog
### Fixes
- Fix inc-rename. Closes #463

### Features
- rust: Remove rust-tools and use ferris.nvim instead. 
- json: Add schemastore to json pack
- yaml: Add schemastore to yaml pack


lua/astrocommunity/bars-and-lines/heirline-mode-text-statusline/init.lua
```diff
@@ -1,7 +1,7 @@
 return {
   "rebelot/heirline.nvim",
   opts = function(_, opts)
-    local status = require "astronvim.utils.status"
+    local status = require "astrocore.status"
```
kshlm, line 4: Astroui has status now.
```suggestion
    local status = require "astroui.status"
```

lua/astrocommunity/bars-and-lines/heirline-vscode-winbar/init.lua
```diff
@@ -1,7 +1,7 @@
 return {
   "rebelot/heirline.nvim",
   opts = function(_, opts)
-    local status = require "astronvim.utils.status"
+    local status = require "astrocore.status"
```
kshlm, line 4:
```suggestion
    local status = require "astroui.status"
```

lua/astrocommunity/pack/rust/init.lua
```diff
@@ -1,21 +1,63 @@
...
+    end,
+  },
+  {
+    "mrcjkb/rustaceanvim",
+    version = "^3",
+    ft = "rust",
     opts = function()
       local adapter
       local success, package = pcall(function() return require("mason-registry").get_package "codelldb" end)
+      local cfg = require "rustaceanvim.config"
```
mrcjkb, line 60: rustaceanvim auto-detects `codelldb` from the path and from the mason-registry (if installed).
So this should now be redundant :smile:

Uzaaft: Interesting...I'll remove it! :D

Uzaaft: Doesn't this mean that we can remove the whole opts function? @mrcjkb

mrcjkb: I'm not sure what this part does:

```lua
return { server = require("astrolsp").lsp_opts "rust_analyzer", dap = { adapter = adapter } }
```

I suppose ` require("astrolsp").lsp_opts` includes an `on_attach` function that sets keymaps, etc.?

PR #636, opened 2023-11-11
fix(inc-rename-nvim): Fix inc-rename for v4, and add noice preset if noice is enabled

Uzaaft:
# Changelog

### Fixes
* Fix inc-rename bug 

### Features
* Move rust pack from rust-tools to ferris.nvim
* Add noice preset to inc-rename pack


Uzaaft:
Merged directly.

PR #638, opened 2023-11-12
fix(ultimate-autopair-nvim): Fix breaking changes in config, and pin plugin version to 0.6

lougreenwood:
## 📑 Description

Fixes the config for `ultimate-auto-pair` to work with the latest version and also pins the version at 0.6 to avoid future breaking changes (this is also recommended by the plugin).


lougreenwood:
I'm not sure why the `stylua` check is failing, it seems to be asking me to use `"` instead of `'`, but that's what the code is also using. Possibly related to force-pushing maybe?

Uzaaft (approved):

PR #648, opened 2023-11-22
feat(recipes)!: add recipes from the documentation and centralize them in the `recipes` category

mehalter:
## 📑 Description

Now that v4 allows all of the configuration to be accessible through the plugins table we can start moving things from the docs and nice recipes into the AstroCommunity repo. Stuff like the NvChad look-alike recipes and stuff.

We decided that `recipes` is a good category for all of the recipes that come from the docs. That makes this a breaking change since we move the recipes that already exist as well as adding the new ones.

## ℹ Additional Information

We already have some docs recipes in the AstroCommunity like this: https://github.com/AstroNvim/astrocommunity/tree/main/lua/astrocommunity/bars-and-lines/heirline-mode-text-statusline

If we decide to change how these are organized we should remember to find all of them and move them as well.


Uzaaft:
I think personally that a recipe folder would make the most sense, no?
cc @AstroNvim/astrocommunity-maintainers  In case anyone there want to voice their opinion.

owittek:
I don't think we have to differentiate those from other configs. If you look into `bar-and-lines` we already have some "recipes" in there

mehalter:
@owittek if you read my original message I mention those as well and say we should move them if we decide to reorganize.

mehalter:
I think this is ready to go now!

Uzaaft:
Ah man this looks amazing!

Uzaaft (approved):

PR #649, opened 2023-11-23
docs(rust): add guide on how to format with clippy

chaozwn:

### DIFF
--- a/lua/astrocommunity/pack/rust/README.md
+++ b/lua/astrocommunity/pack/rust/README.md
@@ -24,7 +24,7 @@ upside is that you can run `cargo build` without that being blocked while rust-a
 The first three levels of this are AstronVim-specific configurations, the rest is rust-analyzer specific:
 
 ```
-   lsp = {
+  lsp = {
     config = {
       rust_analyzer = {
         settings = {
--- a/lua/astrocommunity/pack/rust/README.md
+++ b/lua/astrocommunity/pack/rust/README.md
@@ -15,7 +15,7 @@ This plugin pack does the following:
 Configuring the LSP is out of scope for community packs, but here are some interesting things you
 may want to do:
 
-### Make Rust-Analyzer use its own profile
+## Make Rust-Analyzer use its own profile
 
 Add this into your LSP configuration to make the rust-analyzer use its own profile. The effect is
 that you will use more disk space as rust-analyzer gets its own special folder in `targets`. The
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/pack/rust/README.md
```diff
@@ -24,14 +24,29 @@ upside is that you can run `cargo build` without that being blocked while rust-a
 The first three levels of this are AstronVim-specific configurations, the rest is rust-analyzer specific:
 
 ```
-  lsp = {
+   lsp = {
```
mehalter, line 27:
```suggestion
  lsp = {
```

mehalter (dismissed):
Just small formatting fix. I don't do any rust so I can't speak to the changes in behavior

Uzaaft (changes requested):
The idea is good, but the execution could be better. 
I'd redo the content, and instead add is as a seperate section in the readme. Kind of like:
{What we already have}
## Use clippy instead of cargo check
...Settings stuff here

Uzaaft:
@chaozwn  change the title of the PR to be more in line of what the PR does

chaozwn:
> @chaozwn change the title of the PR to be more in line of what the PR does

LGTM.

Uzaaft:
Could you fix the things I pointed out and we are good to go

chaozwn:
> Could you fix the things I pointed out and we are good to go

done.

lua/astrocommunity/pack/rust/README.md
```diff
@@ -39,3 +39,62 @@ The first three levels of this are AstronVim-specific configurations, the rest i
...
+              extraArgs = { "--no-deps" },
+            },
+            procMacro = {
+              enable = true,
+              ignored = {
+                ["async-trait"] = { "async_trait" },
+                ["napi-derive"] = { "napi" },
+                ["async-recursion"] = { "async_recursion" },
+              },
+            },
```
Uzaaft, line 94: This shouldn't be here since the section, and PR is about adding a guide on how to format with clippy

lua/astrocommunity/pack/rust/README.md
```diff
@@ -39,3 +39,62 @@ The first three levels of this are AstronVim-specific configurations, the rest i
...
+```
+  lsp = {
+    config = {
+      rust_analyzer = {
+        settings = {
+            cargo = {
+              allFeatures = true,
+              loadOutDirsFromCheck = true,
+              runBuildScripts = true,
+            },
```
Uzaaft, line 80: Thsi should also not be in the PR, since it's about adding a guide on how to format with clippy

lua/astrocommunity/pack/rust/README.md
```diff
@@ -39,3 +39,62 @@ The first three levels of this are AstronVim-specific configurations, the rest i
...
+### Why Use `clippy` Instead of `cargo check`?
+
+You would use `clippy` instead of (or in addition to) `cargo check` when you want not just to check that your code compiles, but also to ensure that it adheres to best practices and is as clean and efficient as possible. `clippy` can catch potential bugs and inefficiencies that `cargo check` would not.
+
+In summary:
+
+- Use `cargo check` when you want a fast way to check that your code compiles.
+- Use `clippy` when you want a more thorough analysis that includes style checks, code complexity, and potential logical errors.
+
+It's common to use both tools as part of your development workflow. You might run `cargo check` frequently while developing, and then run `clippy` periodically to clean up and refine your code.
```
Uzaaft, line 70: I feel like this whole section is really unnessecary.
I think you should instead add a link to the rust docs. 
What do you think? @mehalter

Uzaaft: I'll just make the decision.
I think this section is unnessecary, and should be removed.
Replace it with something like:

If you'd like to run cargo clippy instead of cargo check on safe, add the following...

@chaozwn

chaozwn: > I'll just make the decision. I think this section is unnessecary, and should be removed. Replace it with something like:
> 
> If you'd like to run cargo clippy instead of cargo check on safe, add the following...
> 
> @chaozwn

fine

Uzaaft (approved):

PR #651, opened 2023-11-23
V4 cleanup

mehalter:
## 📑 Description

This does a few things:

- Fixes the linting issues with the grapple plugin spec
- Some plugins like Harpoon/Octo/etc. provide a mapping group with an icon. We can leverage AstroUI for doing this and easily make that icon modifiable by the user as well as automatically handle when icons are disabled and padding
- We want AstroCore mappings to be in the dependencies of the plugins that provide the functionality. If the user sets the plugin to `enable = false` then it will make sure that those mappings do not get set. Previously they would be set.
- Normalie all mappings `<Leader>` should be capitalized, stuff like `<CR>`, `<C-`, `<CMD>` should all be capitalized.

### DIFF
--- a/lua/astrocommunity/color/ccc-nvim/init.lua
+++ b/lua/astrocommunity/color/ccc-nvim/init.lua
@@ -1,19 +1,21 @@
 return {
-  {
-    "AstroNvim/astrocore",
-    opts = {
-      mappings = {
-        ["<leader>uC"] = { "<cmd>CccHighlighterToggle<cr>", desc = "Toggle colorizer" },
-        ["<leader>zc"] = { "<cmd>CccConvert<cr>", desc = "Convert color" },
-        ["<leader>zp"] = { "<cmd>CccPick<cr>", desc = "Pick Color" },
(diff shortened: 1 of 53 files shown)


Uzaaft:
> ## 📑 Description
> This does a few things:
> 
> * Fixes the linting issues with the grapple plugin spec
> * Some plugins like Harpoon/Octo/etc. provide a mapping group with an icon. We can leverage AstroUI for doing this and easily make that icon modifiable by the user as well as automatically handle when icons are disabled and padding
> * We want AstroCore mappings to be in the dependencies of the plugins that provide the functionality. If the user sets the plugin to `enable = false` then it will make sure that those mappings do not get set. Previously they would be set.
> * Normalie all mappings `<Leader>` should be capitalized, stuff like `<CR>`, `<C-`, `<CMD>` should all be capitalized.
> 
> ## ℹ Additional Information

The hero we don't deserve but we need 🫶🏾

Uzaaft (approved):

PR #655, opened 2023-11-26
fix(swift): fix LSP handler error

gacallea:
## 📑 Description

fixing an issue where:

- enabling  `{ import = "astrocommunity.pack.swift" },`
- enabling `lsp = { servers = { "sourcekit"} }` (not installed with mason)

- would trigger this error:

`[lspconfig] Cannot access configuration for sourcekit-lsp. Ensure this server is listed in "server_configurations.md" or added to a custom server.`

### DIFF
--- a/lua/astrocommunity/pack/swift/README.md
+++ b/lua/astrocommunity/pack/swift/README.md
@@ -1,17 +1,23 @@
 # Swift Language Pack
 
-## Note: xbase is a WIP plugin, and the experience provided by this plugin might be lackluster
+## Note
 
-**Requirements:**
+xbase is a WIP plugin, and the experience provided by this plugin might be
+lackluster.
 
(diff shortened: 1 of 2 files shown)


gacallea:
After some considerations and comments from @Uzaaft I gave it another try. here's what works:

1  - pack/swift/init.lua with only xbase (sourcekit line restored with changes, null-ls removed) + treesitter + dap
2 - no lsp config in user/init.lua needed
3 - swiftformat + swiftlint added to user/null-ls.lua (adding them to the pack was opinionated)

notes: if null-ls is present in pack/swift only it won't use them for some reason, two birds with a stone: unopinionated + user defined == this way the change is minimal, unopinionated and 100% working 🙂 

I hope this helps. If I made mistakes I apologize. If this is bad somehow, feel free to refuse it. I won't take it persoally :)

Uzaaft (approved):

PR #658, opened 2023-11-27
feat(haskell): add neotest-haskell, haskell-snippets, telescope-hoogle to pack

gacallea:
## 📑 Description

Adding netotest-haskell, haskell-snippets and **optionally** telescope-hoogle.
Basing my changes on how @mrcjkb [implemented it on lazyvim](https://github.com/LazyVim/LazyVim/pull/2052).

## ℹ Additional Information

Since hoogle is an optionally recommended requirement for haskell-tools, the README reflects that and add instructions on how to enable the pack to install it and configure it properly.

### DIFF
--- a/lua/astrocommunity/pack/haskell/README.md
+++ b/lua/astrocommunity/pack/haskell/README.md
@@ -2,17 +2,34 @@
 
 Requires:
 
-- plenary.nvim
+- plenary.nvim: Since v3 droppped this requirement, it is handled by the pack
+during install.
 
 Optionally:
 
-- A local hoogle installation
-- [fast-tags](https://github.com/elaforge/fast-tags) (reccomended by haskell-tools)
+- A local [hoogle](https://github.com/ndmitchell/hoogle) installation
+(recommended by haskell-tools)
+- [fast-tags](https://github.com/elaforge/fast-tags) (recommended by
+haskell-tools)
 
 This plugin pack does the following:
 
 - Adds `haskell` treesitter parsers
 - Adds `haskell-language-server` language server
+- Adds `haskell-debug-adapter` dap adapter
 - Adds [haskell-tools.nvim](https://github.com/mrcjkb/haskell-tools.nvim)
+(either v2 or v3 depending on nvim version)
 - Adds `yaml` language pack
 - Adds `json` language pack
+- Adds `neotest` from Astrocommunity
+- Adds [neotest-haskell](https://github.com/mrcjkb/neotest-haskell)
+- Adds [haskell-snippets.nvim](https://github.com/mrcjkb/haskell-snippets.nvim)
+
+Optionally available:
+
+> :warning: A local hoogle installation is required for this to work.
+
+- Adds [telescope-hoogle](https://github.com/psiska/telescope-hoogle.nvim)
+
+To install, set `{ "luc-tielen/telescope_hoogle" },` in `user/plugins/core.lua`
+and install this pack as normal.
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/pack/haskell/init.lua
```diff
@@ -12,14 +14,13 @@ return {
...
+    ft = haskell_ft,
     dependencies = {
       "nvim-lua/plenary.nvim",
       { "nvim-telescope/telescope.nvim", optional = true },
       { "mfussenegger/nvim-dap", optional = true },
     },
-    version = "^2",
-    -- load the plugin when opening one of the following file types
-    ft = { "haskell", "lhaskell", "cabal", "cabalproject" },
+    version = "^2", -- version ^3 requires nvim >=0.9
```
Uzaaft, line 23:
```suggestion
    version = vim.fn.has "nvim-0.8" and "^2" or "^3"
```

gacallea: beware that version 3 has a more streamlined configuration:

```
{
  'mrcjkb/haskell-tools.nvim',
  version = '^3', -- Recommended
  ft = { 'haskell', 'lhaskell', 'cabal', 'cabalproject' },
}
```

we could use a block for this as an alternative?

gacallea: I need to check init.lua locally, I may be thinking of something else.

gacallea: not sure about breaking changes, but for one: dependencies on plenary seems to be gone. I may be mistaken tho. I'm about to test your change.

gacallea: I may be doing something wrong, because if I apply the v3 line manually I still get haskell-tools 2.4.0 (working locally on my branch btw)

gacallea: something like this perhaps? at the very least, it seems to be doing something. However, haskell-tools fails to install and it complains.

```lua
    version = function()
      if vim.fn.has "nvim-0.8" then
        return "^2"
      else
        return "^3"
      end
    end,
```

Screen Shot 2023-11-28 at 11 48 23 AM

gacallea: ok I've managed. vim.fn.has requires a comparison! also: since it checks the given version and above, this makes more sense:

```
 version = vim.fn.has "nvim-0.9" == 1 and "^3" or "^2",
```

gacallea: also adding this would take care of the plenary obsolete deps:

```
      -- vim.fn.has >= nvim 0.9 removes plenary dependency
      { "nvim-lua/plenary.nvim", optional = vim.fn.has "nvim-0.9" == 1 and true or false },
```

gacallea: I'm going to force push the changes.

Uzaaft (approved):

lua/astrocommunity/pack/haskell/README.md
```diff
@@ -2,17 +2,34 @@
...
+- Adds [neotest-haskell](https://github.com/mrcjkb/neotest-haskell)
+- Adds [haskell-snippets.nvim](https://github.com/mrcjkb/haskell-snippets.nvim)
+
+Optionally available:
+
+> :warning: A local hoogle installation is required for this to work.
+
+- Adds [telescope-hoogle](https://github.com/psiska/telescope-hoogle.nvim)
+
+To install, set `{ "luc-tielen/telescope_hoogle", optional = false },` in
```
Uzaaft, line 34: I believe the optimal tag here isnt required @gacallea

gacallea: I'll give it a try, but iirc I had to set it to properly install it.

gacallea: nope you are right!

Uzaaft (approved):

PR #659, opened 2023-11-30
fix(python-ruff/rust): python-ruff require error and rust add to mason-lspconfig

whitemoon:
## 📑 Description

- fix: python-ruff `utils` requires error
- fix: rust add to mason-lspconfig and delete disabled clangd code
- fix: cpp clangd is enabled

### DIFF
--- a/lua/astrocommunity/pack/rust/init.lua
+++ b/lua/astrocommunity/pack/rust/init.lua
@@ -1,10 +1,10 @@
+local utils = require "astrocore"
 return {
   { import = "astrocommunity.pack.toml" },
   {
     "AstroNvim/astrolsp",
     ---@type AstroLSPOpts
     opts = {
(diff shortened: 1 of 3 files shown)


Uzaaft:
Three actually. You're also changing the cpp pack.

lua/astrocommunity/pack/rust/init.lua
```diff
@@ -19,6 +19,12 @@ return {
       },
     },
   },
+  {
+    "williamboman/mason-lspconfig.nvim",
+    opts = function(_, opts)
+      opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, { "rust_analyzer" })
+    end,
+  },
```
Uzaaft, line 27: rustup is the prefered way to handle rust_analyzer installations. This will be a no-go

PR #663, opened 2023-12-01
fix(rust): enable cmp to crates and remove `clang` handler disabling

whitemoon:
## 📑 Description

- enable cmp to crates and clean unused code


lua/astrocommunity/pack/rust/init.lua
```diff
@@ -4,7 +4,6 @@ return {
     "AstroNvim/astrolsp",
     ---@type AstroLSPOpts
     opts = {
-      handlers = { clangd = false },
```
Uzaaft, line 7: Nice catch!

Uzaaft (approved):

PR #670, opened 2023-12-04
feat(swift): mv to swift only support (no xcode tooling here)

gacallea:
## 📑 Description

Moving away from any XCode tooling in Swift Lang pack. See: https://github.com/AstroNvim/astrocommunity/discussions/669#discussioncomment-7752199

## ℹ Additional Information

I am aware that opinionated is against the Astrocommunity guidelines. However, this pack **has** to be opinionated to make any sense. Otherwise the only thing in it would be treesitter. Not much to offer, really.

While not standardized (not for lack of trying), SwiftLint and SwiftFormat (not the Apple one) appear to be widely accepted by the community. They offer great customization too, in their own `.dotfiles` and it's an approach widely used by the community.

I'd say that until Apple (or the community) sets on a standard, this is the best we could do.

### DIFF
--- a/lua/astrocommunity/pack/swift/README.md
+++ b/lua/astrocommunity/pack/swift/README.md
@@ -9,6 +9,7 @@ This plugin pack does the following:
 - Adds `swift` Treesitter parsers
 - Adds `sourcekit` in LSP servers
 - Adds `sourcekit` in LSP config
+- Adds `codelldb` for debugging
 
 ## SwiftLint and SwiftFormat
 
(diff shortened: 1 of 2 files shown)


gacallea:
not sure it is the proper way, but this seems to work fine, and I pushed it.

```lua

return {
  {
    opts = function(_, opts)
        opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "swift")
      end
    end,
  },
  {
    ft = { "swift", "objcpp", "objc" },
    init = function() require("astronvim.utils.lsp").setup "sourcekit" end,
  },
}
```

gacallea:
resolves https://github.com/AstroNvim/astrocommunity/discussions/669

gacallea:
> @gacallea Added back codelldb.

my bad. I thought it wasn't part of "basic" support. Good call.

Uzaaft (approved):

PR #671, opened 2023-12-04
feat(programming-language-support): Add xbase

gacallea:
## 📑 Description

Moving XBase to its own plugin as suggested by @Uzaaft . There is also a discussion about Swift/Xcode/Xbase/Xcodebuild here: https://github.com/AstroNvim/astrocommunity/discussions/669

## ℹ Additional Information

This changes (moving XCode baed tools to plugins instead of Swift pack) would require to remove the current Swift pack or if acceptable merge an opinionated version (when it manage to fix it...): https://github.com/AstroNvim/astrocommunity/pull/670  -- I'm fine deleting it altogether if this is not adhering to community standards. I'm not agreeing on having a Swift pack with only treesitter in there...


gacallea:
ps: nvimcraft didn't have a fitting category. as per contributing documentation I came up with one. feel free to move.

Uzaaft:
We've got packs with only treesitter stuff, so that's not an issue tbh.

gacallea:
> We've got packs with only treesitter stuff, so that's not an issue tbh.

I see. Well, I'm not in a position to decide on this. I still don't really agree with that, but I guess that's what it is.

Uzaaft:
> > We've got packs with only treesitter stuff, so that's not an issue tbh.
> 
> I see. Well, I'm not in a position to decide on this. I still don't really agree with that, but I guess that's what it is.

It is what it is indeed. 

I'll be going trough your PR later today.

Uzaaft:
@gacallea  Moved the plugin to what I think is a more sensible folder.

gacallea:
closes #669

Uzaaft (approved):

Uzaaft (approved):

PR #676, opened 2023-12-07
fix(zen-mode-nvim): update to AstroNvim v4

TheSast:


Uzaaft (approved):
Great catch! Appreciate it

PR #678, opened 2023-12-10
Update v4 with upstream changes

Uzaaft:


Uzaaft:
TODO after merging: 
1. Fix haskell pack:
Add back this to the top, alongside the filetypes table
```lua
local utils = require "astrocore"
local lsp_utils = require "astrolsp"
```

2. Add astrocore mappings back to Transparent.nvim:
```lua
{
    "AstroNvim/astrocore",
    opts = {
      mappings = {
        n = {
          [prefix .. "T"] = { desc = "Toggle transparency" },
        },
      },
    },
  }
  ```

3. Add astrocore mappings back to ultimate-autopairs:
```lua
"AstroNvim/astrocore",
    opts = {
      mappings = {
        ["<leader>ua"] = {
          desc = "Toggle Ultimate Autopair",
          function()
            local notify = require("astrocore").notify
            local function bool2str(bool) return bool and "on" or "off" end
            local ok, ultimate_autopair = pcall(require, "ultimate-autopair")
            if ok then
              ultimate_autopair.toggle()
              vim.g.ultimate_autopair_enabled = require("ultimate-autopair.core").disable
              notify(string.format("ultimate-autopair %s", bool2str(not vim.g.ultimate_autopair_enabled)))
            else
              notify "ultimate-autopair not available"
            end
          end,
        },
```

PR #683, opened 2023-12-13
feat(typescript): Move eslint-lsp to the proper plugin, and add autocmd to format on save

Uzaaft:


Uzaaft:
R? @AstroNvim/astrocommunity-maintainers

lua/astrocommunity/pack/typescript/init.lua
```diff
@@ -49,7 +49,19 @@ return {
...
+    init = function()
+      vim.api.nvim_create_autocmd("BufWritePost", {
+        desc = "Fix all eslint errors",
+        pattern = { "*.tsx", "*.ts", "*.jsx", "*.js" },
+        group = "...",
+        callback = function()
+          if vim.fn.exists ":EslintFixAll" > 0 then vim.cmd "EslintFixAll" end
+        end,
+      })
+    end,
```
mehalter, line 64:
```suggestion
    opts = function(_, opts)
      opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "tsserver", "eslint")

      vim.api.nvim_create_autocmd("LspAttach", {
        group = vim.api.nvim_create_augroup("eslint_fix_creator", { clear = true }),
        desc = "Create autocommand in buffers where eslint attaches",
        callback = function(args)
          if assert(vim.lsp.get_client_by_id(args.data.client_id)).name == "eslint" then
            vim.api.nvim_create_autocmd("BufWritePost", {
              desc = "Fix all eslint errors",
              buffer = args.buf,
              group = vim.api.nvim_create_augroup(("eslint_fix_%d"):format(args.buf), { clear = true }),
              callback = function() pcall(vim.cmd.EslintFixAll) end,
            })
          end
        end,
      })
    end,
```

Suggestion does a few things:

- We only need to create auto commands for `:EslintFixAll` for buffers that have `eslint` attached.
- Makes it easier to disable and also `init` functions do not stack so that's a bad insertion point
- fixes bug `eslint-lsp` should be `eslint` since that's the name of the language server

mehalter (changes requested):

lua/astrocommunity/pack/typescript/init.lua
mehalter: This also decreases the maintenance burden of maintaining a separate list of filetypes that eslint may or may not use. and rather relies on the filetype detection of the language server attaching.

mehalter: actually this is wrong, let me revise the suggestion. Apparently the `pcall` here is not ideal

lua/astrocommunity/pack/typescript/init.lua
```diff
@@ -49,7 +49,26 @@ return {
   },
   {
     "williamboman/mason-lspconfig.nvim",
-    opts = function(_, opts) opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "tsserver") end,
+    opts = function(_, opts)
+      opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "tsserver", "eslint")
+
```
mehalter, line 54:
```suggestion

```

hehe

mehalter (changes requested):
whoopsie

mehalter (approved):
If you test it and it works, this looks good to me!

lua/astrocommunity/pack/typescript/init.lua
```diff
@@ -49,7 +49,25 @@ return {
   },
   {
     "williamboman/mason-lspconfig.nvim",
-    opts = function(_, opts) opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "tsserver") end,
+    opts = function(_, opts)
+      opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, "tsserver", "eslint")
+      vim.api.nvim_create_autocmd("LspAttach", {
```
mehalter, line 54:
```suggestion
      opts.ensure_installed = utils.list_insert_unique(opts.ensure_installed, { "tsserver", "eslint" })

      vim.api.nvim_create_autocmd("LspAttach", {
```

mehalter (changes requested):

mehalter (approved):
Fixed bug in insertion. But looks good now

PR #692, opened 2023-12-24
feat(pack): add ron tree-sitter for rust develop.

chaozwn:


Uzaaft:
Could you provide some context as to why this should be in the rust pack?

Uzaaft:
Ping @chaozwn

chaozwn:
> Ping @chaozwn

This is actually a commonly used Rust library, and I believe our package should offer an experience as close to that of an IDE as possible. We should provide syntax highlighting for some of the common patterns.

Uzaaft:
Ron is a file format, it should be it's own pack imo.

Uzaaft:
Going ahead and closing this. Please add it as a separate pack.

PR #698, opened 2023-12-30
fix(dial-nvim): change to new mapping api and adding g<C-x> g<C-a> mappings

luxus:
fixing mapping to the new syntax of dial.nvim
https://github.com/monaqa/dial.nvim#usage


Uzaaft (approved):

Uzaaft:
@luxus  Next time, please upstream the patch to the main branch, and we'll downstream it to v4 :)

luxus:
I thought the same after seeing you pushing nim to main :), at least it is tested on v4

PR #701, opened 2024-01-01
fix(editing-support): wrap all require calls to refactoring in function

jay-babu:
Before this, I got a module not found error. Wrapping it in a function, the require call is deferred until the keypress is run where the module is now available

### DIFF
--- a/lua/astrocommunity/editing-support/refactoring-nvim/init.lua
+++ b/lua/astrocommunity/editing-support/refactoring-nvim/init.lua
@@ -1,5 +1,6 @@
 return {
   "ThePrimeagen/refactoring.nvim",
+  event = "User AstroFile",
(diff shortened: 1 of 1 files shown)


Uzaaft:
@jay-babu  I was debugging this myself. Thanks for catching it! And happy new year

Uzaaft:
Fixed one small issue @jay-babu Where the issue still appears when running `nvim`. This should work on all edge cases now.

Uzaaft (approved):

mehalter:
Just for future reference, this should be `fix(refactoring-nvim): ...`

Uzaaft:
@mehalter That one slipped trough by me. I thought I had changed the PR title. My bad!

PR #702, opened 2024-01-01
fix(true-zen-nvim): removing mode from key mapping in true-zen-nvim as it is not valid

jay-babu:
Happy New Year!!!

## 📑 Description

fix(editing-support): removing mode from key mapping in true-zen-nvim as it is not valid

image


Uzaaft (approved):

Uzaaft:
Happy new year 🎉 @jay-babu

PR #718, opened 2024-01-12
fix: refactor utils so that `astrocore` is lazily loaded

mehalter:
## 📑 Description

Now that utils come from `astrocore` which is a plugin, we shouldn't load it at the top of files. It should be loaded within functions so that it happens after all of the AstroCore configuration has happened rather than loading before `lazy.nvim` is finished.


Uzaaft (approved):
Amazing

PR #719, opened 2024-01-13
feat(rust)!: move to rustaceanvim instead of ferris to get DAP support

mehalter:


lua/astrocommunity/pack/rust/init.lua
```diff
@@ -25,32 +49,40 @@ return {
...
-            end
-          end
-        end,
-      })
+    "mrcjkb/rustaceanvim",
+    version = "^3",
+    ft = "rust",
+    opts = function()
+      local adapter
+      local success, package = pcall(function() return require("mason-registry").get_package "codelldb" end)
```
Uzaaft, line 63: We should propably make the dap configuration optional based upon whetever `nvim-dap` is enabled or not in the user's config, no?

mehalter: Yup! Updated in latest commit and also made other stuff `optional` as well, totally slipped my mind.

mehalter: Actually, I looked into the codebase and the dap stuff is already optional. we are setting up the adapter but rustaceanvim handles the optional nature of dap. So this is already good to go actually.

Uzaaft (approved):

PR #723, opened 2024-01-16
fix(nvim-coverage): updated old keyword

jduhamel:
changed requires to dependencies as per lazy.nvim

## 📑 Description

Changes init.lua in test/nvim-coverage


mehalter (approved):
LGTM thanks!

PR #727, opened 2024-01-19
refactor(typescript): set autocmd with AstroLSP

mehalter:
## 📑 Description

This uses the new notation in AstroLSP for setting up autocommands when attaching a language server. This makes it easy for users to opt out of the "ESLintFixAll" autocommand with something like this in their user plugins:

```lua
return {
  "AstroNvim/astrolsp",
  opts = {
    autocmds = {
      eslint_fix_on_save = false
    }
  }
}
```

### DIFF
--- a/lua/astrocommunity/pack/typescript/init.lua
+++ b/lua/astrocommunity/pack/typescript/init.lua
@@ -36,23 +36,6 @@ return {
     "williamboman/mason-lspconfig.nvim",
     opts = function(_, opts)
       opts.ensure_installed = require("astrocore").list_insert_unique(opts.ensure_installed, { "tsserver", "eslint" })
-
-      vim.api.nvim_create_autocmd("LspAttach", {
-        group = vim.api.nvim_create_augroup("eslint_fix_creator", { clear = true }),
-        desc = "Create autocommand in buffers where eslint attaches",
-        callback = function(args)
-          if assert(vim.lsp.get_client_by_id(args.data.client_id)).name == "eslint" then
-            vim.api.nvim_create_autocmd("BufWritePost", {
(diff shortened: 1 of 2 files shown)


mehalter:
I also checked the rest of the AstroCommunity specs and this is the only autocommand that is applicable for this new notation

Uzaaft:
> I also checked the rest of the AstroCommunity specs and this is the only autocommand that is applicable for this new notation

Damn this is neato!

lua/astrocommunity/pack/typescript/init.lua
```diff
@@ -105,6 +88,16 @@ return {
       "AstroNvim/astrolsp",
       ---@diagnostic disable: missing-fields
       opts = {
+        autocmds = {
+          eslint_fix_on_save = {
+            cond = function(client) return client.name == "eslint" and vim.fn.exists ":EslintFixAll" > 0 end,
```
mehalter, line 93:
```suggestion
            cond = function() return vim.fn.exists ":EslintFixAll" > 0 end,
```

Maybe we should just simplify it to this @Uzaaft what do you think?

Uzaaft:
@mehalter  I added some basic docs into this PR

lua/astrocommunity/pack/typescript/init.lua
Uzaaft: I have zero opinions about this 🙊

mehalter: Ok, let's just leave it as is

Uzaaft (approved):

Uzaaft:
LGTM 🚀

PR #728, opened 2024-01-19
feat(neotest): move neotest adapters to language packs

mehalter:
## 📑 Description

The current setup for neotest chooses a random selection of neotest adapters and installs them. This moves that setup to language packs instead to make it a bit more intelligent to the user's needs. It also allows the user to use the `opts` table in the adapters to configure the adapters easily from a user configuration.

### DIFF
--- a/lua/astrocommunity/test/neotest/init.lua
+++ b/lua/astrocommunity/test/neotest/init.lua
@@ -3,9 +3,6 @@ return {
     "nvim-neotest/neotest",
     ft = { "go", "rust", "python" },
     dependencies = {
-      "nvim-neotest/neotest-go",
-      "nvim-neotest/neotest-python",
-      "rouge8/neotest-rust",
       {
         "folke/neodev.nvim",
         opts = function(_, opts)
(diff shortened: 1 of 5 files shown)


lua/astrocommunity/test/neotest/init.lua
```diff
@@ -17,27 +14,16 @@ return {
...
-        -- your neotest config here
-        adapters = {
-          require "neotest-go",
-          require "neotest-rust",
-          require "neotest-python",
-        },
-      }
-    end,
+    opts = {},
     config = function(_, opts)
```
Uzaaft, line 18: Since we already define the config, do we need the opts table?

mehalter: we do not here actually. I will remove it

mehalter: Removed in commit

Uzaaft (approved):

mehalter:
Cool thanks! I'll handle this merge manually :)

PR #736, opened 2024-01-28
feat(nix): replace rnix with nil_ls

manuuurino:
## 📑 Description

[rnix](https://github.com/nix-community/rnix-lsp) is archived and unmantained for a while now, also hasnt been working for me.


mehalter:
This will be handled with rebasing

PR #742, opened 2024-01-29
feat(media): add image.nvim

belafonte:
I recently came across this nvim plugin:
https://github.com/3rd/image.nvim

Which can display images in astronvim via kittys image protocol, which is neat!
Works very well and also prevents astronvim from crashing when opening images :).

It has some requirements, so it does not work out of the box though.
I have only tested it with kitty.

Happy to discuss it or change something if it seems wrong.

### DIFF
--- a/lua/astrocommunity/media/image-nvim/README.md
+++ b/lua/astrocommunity/media/image-nvim/README.md
@@ -15,6 +15,7 @@ These are things you have to setup on your own
 
 * ImageMagick - **mandatory**
 * magick LuaRock - **mandatory** (luarocks --local --lua-version=5.1 install magick)
+    - Magick LuaRock requires Lua 5.1 
 * Kitty >= 28.0 - for the kitty backend
 * ueberzugpp - for the ueberzug backend
 * curl - for remote images
--- a/lua/astrocommunity/media/image-nvim/README.md
+++ b/lua/astrocommunity/media/image-nvim/README.md
@@ -1,0 +1,22 @@
+# image.nvim
+
+This plugin attempts to add image support to Neovim.
+
+**Repository:** <https://github.com/3rd/image.nvim>
+
+Works with Kitty + Tmux, and it handles painful things like rendering an image at a given position in a buffer, scrolling, windows, etc.
+
+It has built-in Markdown and Neorg integrations that you can use right now.
+It can also render image files as images when opened.
+
+**Requirements**
+
+These are things you have to setup on your own
+
+* ImageMagick - **mandatory**
+* magick LuaRock - **mandatory** (luarocks --local --lua-version=5.1 install magick)
+* Kitty >= 28.0 - for the kitty backend
+* ueberzugpp - for the ueberzug backend
+* curl - for remote images
+
+**This plugin is configured to be used with kitty terminal and lua 5.1**
--- a/lua/astrocommunity/media/image-nvim/init.lua
+++ b/lua/astrocommunity/media/image-nvim/init.lua
@@ -1,0 +1,44 @@
+package.path = package.path .. ";" .. vim.fn.expand "$HOME" .. "/.luarocks/share/lua/5.1/?/init.lua;"
+package.path = package.path .. ";" .. vim.fn.expand "$HOME" .. "/.luarocks/share/lua/5.1/?.lua;"
+
+return {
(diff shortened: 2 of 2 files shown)


lua/astrocommunity/media/image-nvim/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+Requirements
+These are things you have to setup on your own:
+
+ImageMagick - mandatory
+magick LuaRock - mandatory (luarocks --local install magick or through your package manager)
+Kitty >= 28.0 - for the kitty backend
+ueberzugpp - for the ueberzug backend
+curl - for remote images
+
```
Uzaaft, line 18: The format of the requirements should follow other README's with similar content

belafonte: Yes sorry, my bad.

Formatted the readme.
Hope it is ok now.

lua/astrocommunity/media/image-nvim/README.md
```diff
@@ -0,0 +1,22 @@
...
+These are things you have to setup on your own
+
+* ImageMagick - **mandatory**
+* magick LuaRock - **mandatory** (luarocks --local --lua-version=5.1 install magick)
+* Kitty >= 28.0 - for the kitty backend
+* ueberzugpp - for the ueberzug backend
+* curl - for remote images
+
+**This plugin is configured to be used with kitty terminal and lua 5.1**
```
Uzaaft, line 22: I don't think lua 5.1 is mentioned explicitly, so that should  be dropped.

belafonte: At the moment lua 5.1 is in a way directly referenced, since the first two lines in the init.lua are referencing like so:

```
package.path = package.path .. ";" .. vim.fn.expand "$HOME" .. "/.luarocks/share/lua/5.1/?/init.lua;"
package.path = package.path .. ";" .. vim.fn.expand "$HOME" .. "/.luarocks/share/lua/5.1/?.lua;"
```

This is exactly the same as the plugin does it.
Not sure if it works with a higher version of lua ( might be ) but untested.

These lines can also be somewhere else, ( polish function for instance? ) but then it becomes even more manual configuration.

Maybe there is a more "astro" way to do this?

Uzaaft: > At the moment lua 5.1 is in a way directly referenced, since the first two lines in the init.lua are referencing like so:
> 
> ```
> package.path = package.path .. ";" .. vim.fn.expand "$HOME" .. "/.luarocks/share/lua/5.1/?/init.lua;"
> package.path = package.path .. ";" .. vim.fn.expand "$HOME" .. "/.luarocks/share/lua/5.1/?.lua;"
> ```
> 
> This is exactly the same as the plugin does it. Not sure if it works with a higher version of lua ( might be ) but untested.
> 
> These lines can also be somewhere else, ( polish function for instance? ) but then it becomes even more manual configuration.
> 
> Maybe there is a more "astro" way to do this?

We should do this in a better way. I’ll have to think on it a bit.

belafonte: Ok good.
yeah, probably not the best way to hardcode this into the plugin...

Uzaaft: Ah seems to be that the limitation of lua 5.1 is imposed by the luarock bit. magick's lua bindings actually requires luajit

lua/astrocommunity/media/image-nvim/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+It has built-in Markdown and Neorg integrations that you can use right now.
+It can also render image files as images when opened.
+
+**Requirements**
+
+These are things you have to setup on your own
+
+* ImageMagick - **mandatory**
+* magick LuaRock - **mandatory** (luarocks --local --lua-version=5.1 install magick)
```
Uzaaft, line 17:
```suggestion
* magick LuaRock - **mandatory** (luarocks --local --lua-version=5.1 install magick)
    * Magick LuaRock requires Lua 5.1 
```

Uzaaft (approved):

Uzaaft (approved):

Uzaaft:
This might require some changes in the future but seems good to go for now.

PR #744, opened 2024-01-29
feat(rust):change the archived rust-tools.nvim to its fork:rustaceanvim

limuy2022:
## 📑 Description

Because the rust-tools is archived and rustaceanvim is recommended by the author of the rust-tools.I replace the rust-tools with rustaceanvim,which is an active project and forked from the rust-tools.nvim


Uzaaft:
This is already on the pipetone with v4, so wont merge this in.

PR #746, opened 2024-02-01
fix(pack): disable `go.nvim` defaults to prevent conflicts

tobihans:
This replace `x-ray/go.nvim` with a much more simpler alternative `olexsmir/gopher.nvim`.

### DIFF
--- a/lua/astrocommunity/pack/go/init.lua
+++ b/lua/astrocommunity/pack/go/init.lua
@@ -64,12 +64,18 @@ return {
     opts = {},
   },
   {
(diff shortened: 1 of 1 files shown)


Uzaaft:
Converting this as a draft for now.
Also, @tobihans If we were to merge this, we should merge it to main and downstream it

tobihans:
Turns out there is a settings to disable default features (lsp handling, diagnostics, ...) which conflicts with AstroLsp. The end user can set mappings for the provided commands. Thanks

Uzaaft (approved):

Issue #560, opened 2023-09-08
Loading on `CmdlineEnter` for go pack seems necessary?


jay-babu:
https://github.com/AstroNvim/astrocommunity/blame/1c67074be443bfc101807e0239317ef175de70a2/lua/astrocommunity/pack/go/init.lua#L44

Uzaaft:
Any reason why it seems necessary?
It's part of the config that the plugin maintainer provides in the readme.

PR #747, opened 2024-02-02
fix(vim-wakatime): follow instructions in plugin repo

jay-babu:
https://github.com/wakatime/vim-wakatime

### DIFF
--- a/lua/astrocommunity/media/vim-wakatime/init.lua
+++ b/lua/astrocommunity/media/vim-wakatime/init.lua
@@ -1,1 +1,7 @@
-return { "wakatime/vim-wakatime", event = "User AstroFile" }
+return { 
+  "wakatime/vim-wakatime",
+  event = "User AstroFile",
+  setup = function()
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/media/vim-wakatime/init.lua
```diff
@@ -1 +1,7 @@
-return { "wakatime/vim-wakatime", event = "User AstroFile" }
+return { 
+  "wakatime/vim-wakatime",
+  event = "User AstroFile",
```
jay-babu, line 3: it is weird how in go.nvim we adhere to the readme of go.nvim plugin's setup instructions but here we don't. how is this determined? this stems from a previous issue I created

jay-babu: https://github.com/AstroNvim/astrocommunity/issues/560

jay-babu:
Don't merge this. their instructions are wrong

PR #749, opened 2024-02-04
feat(pack): add Pkl language support

mehalter:
## 📑 Description

This adds support for the new Pkl language from Apple.


lua/astrocommunity/pack/pkl/init.lua
```diff
@@ -0,0 +1,35 @@
+return {
+  "https://github.com/apple/pkl-neovim",
```
Uzaaft, line 2:
```suggestion
"apple/pkl-neovim"
```

mehalter: Whoopsie 😅

Uzaaft (approved):

PR #753, opened 2024-02-07
Add `nvim-lint` and continue adding support for `conform` and `nvim-lint` to packs

mehalter:
## 📑 Description

This adds `nvim-lint` and adds optional support to language packs. I believe this finishes adding support for `nvim-lint` and `conform` to all packs (at least the ones that were previously being configured through `null-ls`. The only packs that haven't been moved are the typescript packs (@Uzaaft and @owittek should probably figure out how to do that).

## ℹ Additional Information

This PR does quite a bit. I will handle manually merging it after review to rebase it rather than our squash model.


owittek (approved):
LGTM 🚀

PR #754, opened 2024-02-07
chore(mini-surround): remove disabling of vim-sandwich

TheSast:
Non-core plugins shouldn't be disabled. Someone would have to explicitly add both plugins to encounter any issues.


mehalter (approved):
Looks great! Thanks!

PR #758, opened 2024-02-10
Update init.lua

divan:
Currently, there is an issue with treesitter for Dart, which makes it extremely slow on basic editing tasks (like inserting new line takes 1 second on my M3 Pro).

The issue is tracked here: https://github.com/nvim-treesitter/nvim-treesitter/issues/4945

One of the proposed solutions in the comments that worked for me is to use a latest working version of treesitter: https://github.com/nvim-treesitter/nvim-treesitter/issues/4945#issuecomment-1691168369

Until this is fixed, I propose to switch this pack to use working version as per the comment above.

### DIFF
--- a/lua/astrocommunity/pack/dart/init.lua
+++ b/lua/astrocommunity/pack/dart/init.lua
@@ -3,6 +3,7 @@ return {
   { import = "astrocommunity.pack.yaml" },
   {
     "nvim-treesitter/nvim-treesitter",
+    commit = "f2778bd1a28b74adf5b1aa51aa57da85adfa3d16",
     opts = function(_, opts)
       if opts.ensure_installed ~= "all" then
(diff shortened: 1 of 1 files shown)


Uzaaft:
Please fix the PR title and the CI @divan

mehalter (changes requested):
I think a better approach is to just disable treesitter based indentation for dart files as described here: https://github.com/UserNobody14/tree-sitter-dart/issues/60#issuecomment-1867049690

Reverting to an old treesitter commit improves it but at the end of the day it's due to the parser itself and not nvim-treesitter. The issues are the same with other editors like helix as well. We should just disable the dart indentation stuff instead since that's the borked part of that parser

Uzaaft:
Closing due to inactivity. @divan  Please re-open the PR with the suggestions from mehalter in the future if you still feel like doing so.

PR #762, opened 2024-02-17
Update init.lua

divan:
Fix warning(error?) – "this function expects 2 arguments, but receives 5"


Uzaaft:
@divan  Please follow the contribution rules and fix your PR.

Uzaaft:
@divan  Please update your local astronvim. 
The new list_insert_unique function is like this:
``` 
--- Insert one or more values into a list like table and maintain that you do not insert non-unique values (THIS MODIFIES `lst`)
---@param lst any[]|nil The list like table that you want to insert into
---@param vals any|any[] Either a list like table of values to be inserted or a single value to be inserted
---@return any[] # The modified list like table
function M.list_insert_unique(lst, vals)
```

Issue #183, opened 2023-05-03
Update README files for packs


sjcobb2022:
Recently there have been a few issues regarding caveats with certain packs. Notably #179 and #101.

I think that it would be a good idea to require / request more documentation and examples in the README when adding packs.

This is just a suggestion and isn't necessary, but it may help newer users.

luxus:
yeah we had a lot of ideas in the same directions, but this means rewriting the guidelines and extending all existing readmes.

luxus:
with some of the new functions we have now we can cover many cases

PR #763, opened 2024-02-21
docs(java): Add note about root pattern

Uzaaft:
Will fix one of the issues raised in #183.
I think we can go ahead and close that issue after me amend the contrib guide, any thoughts? @mehalter


mehalter (approved):
LGTM

PR #769, opened 2024-02-24
fix(pack): replace deprecated none-ls builtins

tku137:
## 📑 Description

I started getting deprecation warnings from `null-ls`. It seems `null-ls` wants to clean unmaintained builtins, see: https://github.com/nvimtools/none-ls.nvim/issues/58

This PR replaces some of the deprecated builtins with the suggested alternatives, but is by no means exhaustive. For example, I am not sure about `eslint`, as I don't use it.

### DIFF
--- a/lua/astrocommunity/pack/bash/README.md
+++ b/lua/astrocommunity/pack/bash/README.md
@@ -5,5 +5,5 @@ This plugin pack does the following:
 - Adds `bash` Treesitter parser
 - Adds `bashls` language server
 - Adds `shfmt` formatter
-- Adds `shellcheck` linter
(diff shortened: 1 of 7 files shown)


mochaaP:
As for ESLint, use the LSP here: <https://github.com/hrsh7th/vscode-langservers-extracted> (extracted from <https://github.com/microsoft/vscode-eslint>).

This is available in mason registry as `eslint-lsp`. <https://github.com/mason-org/mason-registry/blob/main/packages/eslint-lsp/package.yaml>

mehalter (approved):
LGTM

PR #777, opened 2024-02-28
Added Keymapping for go pack

Rishikesh01:
Imo these keymapping should atleast be provided in community repo, as debugging and running tests is quite common

### DIFF
--- a/lua/astrocommunity/pack/go/init.lua
+++ b/lua/astrocommunity/pack/go/init.lua
@@ -140,4 +140,18 @@ return {
       },
     },
   },
+  n = {
+    ["<leader>dt"] = {
+      function() require("dap-go").debug_test() end,
+      desc = "Debug single test",
+    },
+    ["<leader>jt"] = {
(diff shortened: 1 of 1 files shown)


Uzaaft:
Your PR doesnt make sense.

Uzaaft:
The keys you added should be in the nvim-dap-go spec, not where you placed them.

Rishikesh01:
@Uzaaft The aim of this PR is improve user experience as

Running `Pack` commands like `:GoTestFunc` ruin's user experience IMO.

For example:

In VS code if I want to run a single test I would Simply click on the inlay button `run test`.
VScode is a GUI application sense to provides a `run test` feature via point and click method as a User would expect.


The point I am trying to make here is, Astronvim is a config that greatly simplifies using nvim TUI, it should have keymapping for common operations like running tests, debugging test, adding tags, filling structs etc... in it's community plugin.

Uzaaft:
I’m not disputing the value of those commands. But your code wouldnt work.

PR #780, opened 2024-03-01
feat(pack): add Templ language pack

mehalter:
## 📑 Description

This adds a pack for the [Templ](https://templ.guide/) HTML templating language

### DIFF
--- a/lua/astrocommunity/pack/templ/README.md
+++ b/lua/astrocommunity/pack/templ/README.md
@@ -1,0 +1,6 @@
+# Templ Language Pack
+
+This plugin pack does the following:
+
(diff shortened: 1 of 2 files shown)


Uzaaft:
@mehalter  Doesn't templ require go? Shouldnt that be included in the pack, or mentioned in the docs.

mehalter:
No it doesn't seem to require go

mehalter:
I mean to run the language server on a `.templ` file and to parse the treesitter it doesn't require Go

mehalter:
we could add it tho

mehalter:
oh actually it very much doesn't require `gopls` or anything. So it is good as is

PR #781, opened 2024-03-02
feat(grapple-nvim): better keymaps + add nvim-web-devicons dependency

cbochs:
## 📑 Description

Hello, again 👋 

I received a bit of feedback after releasing the rewrite the other day. Thought I would send one more PR to address that here as well 🙂 

### Changes

- Add `nvim-tree/nvim-web-devicons` as a lazy (optional) dependency
- Use `toggle_*` instead of `open_*` so Grapple windows can be open/closed with the same mapping


Uzaaft (approved):
@cbochs Thank you for the PR!

PR #788, opened 2024-03-06
feat(coc): add marketplace

wroyca:
## 📑 Description

Add coc-marketplace to be pre-installed; this is similar to, e.g., Mason, where users can easily navigate, find, and install Coc adapters/linter, and more.

## Additional Information

Screenshot from 2024-03-06 14-31-18

### DIFF
--- a/lua/astrocommunity/lsp/coc-nvim/init.lua
+++ b/lua/astrocommunity/lsp/coc-nvim/init.lua
@@ -31,6 +31,7 @@ return {
         if not opts.options.g then opts.options.g = {} end
         opts.options.g.coc_global_extensions = {
           "coc-json",
+          "coc-marketplace",
         }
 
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/lsp/coc-nvim/init.lua
```diff
@@ -2,6 +2,11 @@ return {
   {
     "neoclide/coc.nvim",
     branch = "release",
+    init = function()
+      vim.g.coc_global_extensions = {
+        "coc-marketplace",
```
mehalter, line 7: This should be configured in AstroCore `options.g` table which exposes it to the user rather than a closed box `init` function that cannot be extended.

mehalter (changes requested):

mehalter (approved):
LGTM

PR #789, opened 2024-03-06
feat(coc): add json for coc configuration

wroyca:
## 📑 Description

Pre-install `coc-json` for Coc configuration, which comes with its own completion source.

## Additional Information

image

### DIFF
--- a/lua/astrocommunity/lsp/coc-nvim/init.lua
+++ b/lua/astrocommunity/lsp/coc-nvim/init.lua
@@ -2,11 +2,6 @@ return {
   {
     "neoclide/coc.nvim",
     branch = "release",
-    init = function()
-      vim.g.coc_global_extensions = {
-        "coc-json",
-      }
-    end,
     cmd = {
       "CocCommand",
       "CocConfig",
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/lsp/coc-nvim/init.lua
```diff
@@ -2,6 +2,11 @@ return {
   {
     "neoclide/coc.nvim",
     branch = "release",
+    init = function()
+      vim.g.coc_global_extensions = {
+        "coc-json",
+      }
+    end,
```
mehalter, line 9: This should be configured in AstroCore `options.g` table which exposes it to the user rather than a closed box `init` function that cannot be extended.

mehalter (changes requested):

lua/astrocommunity/lsp/coc-nvim/init.lua
wroyca: Being clueless about astronvim codebase, do you mean something like:

```lua
---@param opts AstroCoreOpts
opts = function(_, opts)
opts = require ("astrocore").extend_tbl({ coc_global_extensions = { "coc-json" }}, opts)
vim.g.coc_global_extensions = opts.coc_global_extensions or {}
```

mehalter (approved):
LGTM

PR #791, opened 2024-03-06
feat(coc): add service to statusbar

wroyca:
## 📑 Description

This add "service" to the statusbar. vim.g.coc_service_initialized always returns true even when no services are running. So, here we use vim.g.coc_workspace_initialized instead, which holds true only when LSP is actually running. The downside is that it will not work outside of workspace(s) - but this won't result in an error, so it's fine to overlook until we can think through a way to fetch services (if possible).

## Additional Information

image

image


mehalter:
This other commit mainly (1) saves on calculating condition two times and saves a bit of white space

wroyca:
> Took a slightly different approach and manually added it here:
> 
> [62943d1](https://github.com/AstroNvim/astrocommunity/commit/62943d19a81df01e8c4539a4b04a8c04cf35f585)
> 
> Let me know what you think! Also this might not be super fool proof just because I noticed that sources don't always show something in the status (`coc-json` when editing the coc config doesn't show up so this also wouldn't show up 🤷🏼 not sure if that matters)

Yeah, it's unfortunate that CoC doesn't expose an API for that, but whatever, good enough haha

PR #794, opened 2024-03-07
fix(coc-nvim): fix heirline with coc list

wroyca:
## 📑 Description

Fix a very annoying issue where CoC would overwrite heirline. Note that we do lose the count information, so we'll have to integrate it into heirline, probably by looking at how it is done internally

### Before:
image

### After:
image

### DIFF
--- a/lua/astrocommunity/lsp/coc-nvim/init.lua
+++ b/lua/astrocommunity/lsp/coc-nvim/init.lua
@@ -40,10 +40,17 @@ return {
         if not opts.autocmds then opts.autocmds = {} end
         opts.autocmds.coc_settings = {
           {
-            event = "FileType",
-            pattern = "list",
-            desc = "disable foldcolumn for coc lists",
-            callback = function() vim.wo.foldcolumn = "0" end,
+            event = { "BufWinEnter", "BufWinLeave" },
+            pattern = "*",
+            desc = "disable foldcolumn and statusline for coc lists",
+            callback = function()
+              if vim.bo.filetype == [[list]] then
+                vim.wo.foldcolumn = "0"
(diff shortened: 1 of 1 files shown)


Uzaaft:
@wroyca  I'll take a look at coc.nvim tomorrow to see if we can reproduce whatever they do, if not I'll merge this in.

Can you resolve the conflicts while you wait?

Uzaaft:
Didn't manage to find out how they do things. 
I think we can add a note about the behavior in the readme, and call it a day.
cc @mehalter

mehalter (changes requested):
there are a lot of issues here:

1. this is super non-performant because setting `vim.opts` for some reason is super expensive imo
2. This is pulling the `laststatus` from the `opts` which is not necessarily true if the user has changed it at runtime. This should be saved to a temporary variable the old value.

These can both be fixed by (1) saving a temp variable with the current `laststatus`, check if they are equal and only do the assignment if they are different. It shouldn't reference `options.options.opt.laststatus` at all.

Ideally this is just not our problem and should be reported upstream haha

wroyca:
> there are a lot of issues here:
> 
>     1. this is super non-performant because setting `vim.opts` for some reason is super expensive imo
> 
>     2. This is pulling the `laststatus` from the `opts` which is not necessarily true if the user has changed it at runtime. This should be saved to a temporary variable the old value.
> 
> 
> These can both be fixed by (1) saving a temp variable with the current `laststatus`, check if they are equal and only do the assignment if they are different. It shouldn't reference `options.options.opt.laststatus` at all.
> 
> Ideally this is just not our problem and should be reported upstream haha

It's a trade-off for sure. I don't mind closing this for now, and then we may see with upstream how to approach it later

PR #795, opened 2024-03-08
fix(harpoon): commend out undeclared command

kaankoken:
## 📑 Description

This PR removes `prefix .. t` option since `term` does not exist in `harpoon2` yet. Instead of completely removing the option it is commended-out. 

Also, this PR removes `catppuccin` config from `harpoon` since it should be in [catppuccin](https://github.com/AstroNvim/astrocommunity/tree/main/lua/astrocommunity/colorscheme/catppuccin).


Uzaaft:
No the harpoon configuration of catppuccin should be in the harpoon spec.

Uzaaft:
@kaankoken  I've been thinking, and I think we should just remove the code instead of commenting it out. 
But I'll be doing it in v4, since rebasing main to v4 would be a pain in the ass.

Uzaaft:
Wait. This is being used in the code @kaankoken . Closing.

PR #797, opened 2024-03-09
feat(editing-support): add vim-visual-multi

wroyca:
## 📑 Description

Add vim-visual-multi which is a multiple cursors plugin for vim/neovim.

https://github.com/mg979/vim-visual-multi

1 p0ssV5I5cUY11pLBJoPZIQ

### DIFF
--- a/lua/astrocommunity/editing-support/vim-visual-multi/init.lua
+++ b/lua/astrocommunity/editing-support/vim-visual-multi/init.lua
@@ -1,0 +1,33 @@
+return {
+  "mg979/vim-visual-multi",
+  event = { "User AstroFile", "InsertEnter" },
+  dependencies = {
+    "astronvim/astrocore",
+    ---@param opts astrocoreopts
+    opts = function(_, opts)
+      if not opts.options then opts.options = {} end
+      if not opts.options.g then opts.options.g = {} end
+      if not opts.options.opt then opts.options.opt = {} end
+      opts.options.g.VM_silent_exit = 1
+      opts.options.g.VM_show_warnings = 0
+
+      if not opts.autocmds then opts.autocmds = {} end
+      opts.autocmds.visual_multi_exit = {
+        {
+          event = { "User" },
+          pattern = "visual_multi_exit",
+          desc = "Avoid spurious 'hit-enter-prompt' when exiting vim-visual-multi",
+          callback = function()
+            vim.o.cmdheight = 1
+            vim.schedule(function() vim.o.cmdheight = opts.options.opt.cmdheight end)
+          end,
+        },
+      }
+
+      if not opts.mappings then opts.mappings = require("astrocore").empty_map_table() end
+      local maps = assert(opts.mappings)
+      maps.n["<C-up>"] = { "<C-u>call vm#commands#add_cursor_up(0, v:count1)<cr>", desc = "Add cursor above" }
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/editing-support/vim-visual-multi/init.lua
```diff
@@ -0,0 +1,33 @@
+return {
+  "mg979/vim-visual-multi",
+  event = { "User AstroFile", "InsertEnter" },
+  dependencies = {
+    "astronvim/astrocore",
+    ---@param opts astrocoreopts
+    opts = function(_, opts)
+      if not opts.options then opts.options = {} end
+      if not opts.options.g then opts.options.g = {} end
+      if not opts.options.opt then opts.options.opt = {} end
```
mehalter, line 10:
```suggestion
```

lua/astrocommunity/editing-support/vim-visual-multi/init.lua
```diff
@@ -0,0 +1,33 @@
...
+      opts.options.g.VM_silent_exit = 1
+      opts.options.g.VM_show_warnings = 0
+
+      if not opts.autocmds then opts.autocmds = {} end
+      opts.autocmds.visual_multi_exit = {
+        {
+          event = { "User" },
```
mehalter, line 17:
```suggestion
          event = "User",
```

lua/astrocommunity/editing-support/vim-visual-multi/init.lua
```diff
@@ -0,0 +1,33 @@
...
+          pattern = "visual_multi_exit",
+          desc = "Avoid spurious 'hit-enter-prompt' when exiting vim-visual-multi",
+          callback = function()
+            vim.o.cmdheight = 1
+            vim.schedule(function() vim.o.cmdheight = opts.options.opt.cmdheight end)
+          end,
```
mehalter, line 23: I'll just put it here that I'm extremely not a fan of this approach because this is a pretty brutal hack on how scheduling functions work and doesn't necessarily guarantee that the scheduled function will happen at the time you want it to. Instead this issue should be run upstream rather than a hacky fix here. If the user wants to avoid this issue then they can just set `cmdheight=1` in their config

mehalter (changes requested):
nits

lua/astrocommunity/editing-support/vim-visual-multi/init.lua
mehalter: that being said, it's up to you if you want to keep it here :joy: I don't care what ends up in AstroCommunity

wroyca: I agree – then again, we don't care when it happens. We just need to quickly set cmdheight to 1 on the exit event and flip it back to 0 whenever. If my understanding stand, flipping back to 0 is scheduled to run in the next event loop, which will be executed after the current one (where exit is at, assuming sequential order) has finished, so timing issues shouldn't happen.

Let's keep it as it is for now. I'll raise an issue upstream, and we can remove this once we have a proper fix in our hands.

mehalter (approved):
LGTM

lua/astrocommunity/editing-support/vim-visual-multi/init.lua
wroyca: Thank!

PR #798, opened 2024-03-10
fix(coq_nvim): fix coq bindings

tku137:
## 📑 Description

After coq is lazy loaded, meaning after the first autocompletion is triggered, it's default binding `<C-h>` for `jump_to_mark` (https://github.com/ms-jpq/coq_nvim/blob/coq/docs/SNIPS.md#snippets) clashes with AstroNVims window movement. Basically, as soon as coq is loaded, you can't switch to the left window anymore.

This is an attempt to fix that. I chose `<C-t>` to jump to the next snippet edit, which is absolutely a debatable choice. I think this clashes the least with both AstroNVims and Neovims default bindings.

### DIFF
--- a/lua/astrocommunity/completion/coq_nvim/init.lua
+++ b/lua/astrocommunity/completion/coq_nvim/init.lua
@@ -17,7 +17,7 @@ return {
             g = {
               coq_settings = {
                 auto_start = "shut-up",
-                keymap = { jump_to_mark = "<C-T>" },
+                keymap = { jump_to_mark = "<Tab>" },
               },
             },
           },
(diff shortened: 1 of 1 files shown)


Uzaaft:
> @Uzaaft sorry I can't follow. What do we use in astronvim? `<C-H>` is astronvims "move to left split. But this gets overwritten as soon as coq is loaded, because `<C-H>` is coq's default "jump to mark" binding. Instead of changing astronvim's split movement, I suggest to change coq's jump to mark here.

Lol my message was written in haste. 
In astronvim, we use tab to jump to the next node in a snippet. We should use the same here.

tku137:
Good point, thanks. I just changed it.

mehalter (approved):
LGTM

PR #801, opened 2024-03-13
feat(keybinding): Add mini.clue

Uzaaft:


luxus:
the implementation of the `which_key_queue` is nice.

mehalter (changes requested):
1. Formatting needs to be fixed
2. Keys for lazy loading don't seem correct. There are a lot more triggers than just `<Leader>`

mehalter (approved):
LGTM

Issue #634, opened 2023-11-10
🥶 ⛄️V4 feature freeze


Uzaaft:
Unless the [V4](https://github.com/AstroNvim/AstroNvim/pull/2196) PR get's pushed back, we are instituting a PR freeze of new plugins. Bug fixes are still welcome! 🐞

(PS: Keep the PR coming, they just won't be merged until after v4 is merged. We'll mark all PR's with new features as draft, and I advice you do open them against the v4 branch)

Uzaaft:
Due to the delay in the v4 to January/February, we'll be lifting the feature freeze. 
cc @AstroNvim/astrocommunity-maintainers

Uzaaft:
Reopening now that we are getting close to the v4 release.

Uzaaft:
V4 is released now. Closing this.

PR #813, opened 2024-03-25
fix(neorg)!: fix breaking change in v8.0.0

ALVAROPING1:
## 📑 Description

Neorg recently released version 8.0.0 with a breaking change that requires slightly changing the config. This PR applies the necessary changes to get the plugin working correctly again. This is a breaking change because it requires a couple new system dependencies: Lua 5.1/LuaJIT and the Lua header files.

## ℹ Additional Information

Neorg now includes a `build.lua` file to handle the build process. The old build step specified with `build` is no longer sufficient, so specifying it leads to a broken installation as it overrides the included `build.lua` script.

Neorg now uses luarocks to manage its own dependencies, which is automatically compiled and installed by the `luarocks.nvim` plugin dependency using a system installation of Lua 5.1/LuaJIT with the Lua header files.


Uzaaft:
See #634.

lua/astrocommunity/note-taking/neorg/init.lua
```diff
@@ -1,23 +1,29 @@
...
-      ["core.journal"] = {}, -- Enables support for the journal module
-      ["core.dirman"] = { -- Manages Neorg workspaces
-        config = {
-          workspaces = {
-            notes = "~/projects/notes",
+  {
+    "vhyrro/luarocks.nvim",
+    priority = 1000, -- We'd like this plugin to load first out of the rest
+    config = true,
+  },
```
Uzaaft, line 6: This should. be a dependencies of neorg.

ALVAROPING1: fixed

Uzaaft:
Changed target branch to v4. 
@ALVAROPING1  Are you running v4 or v3 of astronvim?

ALVAROPING1:
> Changed target branch to v4. @ALVAROPING1 Are you running v4 or v3 of astronvim?

I'm running v3 of astronvim currently

Uzaaft:
> > Changed target branch to v4. @ALVAROPING1 Are you running v4 or v3 of astronvim?
> 
> I'm running v3 of astronvim currently. If this is going to be postponed until after v4 is released, it might be worth it to pin neorg's version to its last commit before the recent changes (`583a82e`) as a temp fix, since otherwise this spec won't work with neorg's latest commit

Sure, sounds like a plan. I'll do that in v4 then. The v3 branch is frozen until then

Uzaaft (approved):

lua/astrocommunity/note-taking/neorg/init.lua
```diff
@@ -1,8 +1,12 @@
 return {
   "nvim-neorg/neorg",
```
Uzaaft, line 2:
```suggestion
  "nvim-neorg/neorg",
  version = "^8",
```

PR #815, opened 2024-03-25
fix(neorg): pin version to last commit before v8.0.0's breaking changes

ALVAROPING1:
## 📑 Description

Neorg recently released v8.0.0 with some breaking changes. This PR is a temp fix until AstroNvim v4 is released and #813 is merged, since fixing the breaking changes requires a breaking change of our own due to new required system dependencies.


Uzaaft:
Done in f219659b67b246584c421074d73db0a941af5cbd

PR #817, opened 2024-03-29
feat(pack): add fish pack

manuuurino:


Uzaaft:
@manuuurino  I'd appreciate it if you take a look at other similar packs before.
Also, please add `WhoIsSethDaniel/mason-tool-installer.nvim` into the pack as well with optional support

Uzaaft (changes requested):
please add WhoIsSethDaniel/mason-tool-installer.nvim into the pack

manuuurino:
forgot to mention there is no mason package for fish and fish_indent, also tried your commit and it tells me this aswell

Uzaaft:
> forgot to mention there is no mason package for fish and fish_indent, also tried your commit and it tells me this aswell

You should mention the dependency on fish in the readme, even if its a no-brainer that it's required

Uzaaft (approved):

PR #828, opened 2024-04-04
feat(media): add codesnap.nvim

mehalter:
## 📑 Description

Adds [CodeSnap.nvim](https://github.com/mistricky/codesnap.nvim) plugin with some default bindings

### DIFF
--- a/lua/astrocommunity/media/codesnap-nvim/init.lua
+++ b/lua/astrocommunity/media/codesnap-nvim/init.lua
@@ -1,0 +1,24 @@
+local prefix = "<Leader>c"
+return {
+  "mistricky/codesnap.nvim",
+  build = "make",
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/media/codesnap-nvim/init.lua
```diff
@@ -0,0 +1,24 @@
+local prefix = "<Leader>c"
+return {
+  "mistricky/codesnap.nvim",
+  build = "make",
```
Uzaaft, line 4:
```suggestion
  build = "build_generator",
```
We probably shouldnt use a library compiled with someone else…

mehalter: tested and it doesn't work well. I think this is a fine default

PR #836, opened 2024-04-06
feat(startup): add mini-starter plugin

CMOISDEAD:
## 📑 Description

Added ['mini-starter'](https://github.com/echasnovski/mini.starter) plugin and configured it for Astronvim.

## ℹ Additional Information

Disabled the alpha plugin and configured the buffer close command to open nvim-starter when all buffers are closed.

### DIFF
--- a/lua/astrocommunity/startup/mini-starter/init.lua
+++ b/lua/astrocommunity/startup/mini-starter/init.lua
@@ -7,6 +7,7 @@ return {
   },
   {
     "AstroNvim/astrocore",
+    optional = true,
     ---@type AstroCoreOpts
     opts = {
       mappings = {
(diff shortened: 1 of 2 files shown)


Uzaaft (approved):
Looks like a great starting point! :)

lua/astrocommunity/startup/mini-starter/init.lua
```diff
@@ -0,0 +1,29 @@
+return {
+  { "goolord/alpha-nvim", enabled = false },
+  {
+    "echasnovski/mini.starter",
+    version = "*", -- stable version
+    config = function() require("mini.starter").setup() end,
+  },
+  {
+    "AstroNvim/astrocore",
```
Uzaaft, line 9:
```suggestion
    "AstroNvim/astrocore",
    optional = true,
```

PR #837, opened 2024-04-06
fix(java): rename clang_format to clang-format

krzysztofgb:
## 📑 Description

Update `clang_format` formatter name with `clang-format`.

## ℹ Additional Information

Example of the error I'm currently getting with the latest version of `AstroNvim/astrocommunity`. This fixes that.

```
Error executing vim.schedule lua callback: ...l/share/nvim/lazy/mason.nvim/lua/mason-registry/init.lua:80: Cannot find p
ackage "clang_format".                                                                                                  
stack traceback:                                                                                                        
        [C]: in function 'error'                                                                                        
        ...l/share/nvim/lazy/mason.nvim/lua/mason-registry/init.lua:80: in function 'get_package'                       
        ...on-tool-installer.nvim/lua/mason-tool-installer/init.lua:127: in function 'callback'                         
        ...share/nvim/lazy/mason.nvim/lua/mason-core/async/init.lua:87: in function 'step'                              
        ...share/nvim/lazy/mason.nvim/lua/mason-core/async/init.lua:96: in function 'run'                               
        ...l/share/nvim/lazy/mason.nvim/lua/mason-registry/init.lua:202: in function 'refresh'                          
        ...on-tool-installer.nvim/lua/mason-tool-installer/init.lua:156: in function ''                                 
        vim/_editor.lua: in function ''                                                                                 
        vim/_editor.lua: in function <vim/_editor.lua:0>
```

PR #841, opened 2024-04-07
fix(mason-tool-installer-nvim): Use mappings function for null-ls names

krzysztofgb:
## 📑 Description

Replaces `null_ls_to_package` mappings table with `mason-null-ls`'s provided `getPackageFromNullLs` function for mapping package names to Mason names.

## ℹ Additional Information

* Link to mason-null-ls function used: [here](https://github.com/jay-babu/mason-null-ls.nvim/blob/main/lua/mason-null-ls/mappings/source.lua#L23-L27)
   - This function uses the `null_ls_to_package` table under the hood while also fixing the names of package names following "normal" conventions
   - This function was introduced [here](https://github.com/jay-babu/mason-null-ls.nvim/pull/40/files) and a lot of "normal" mappings (including `clang_format` -> `clang-format`) got removed from the mapping sources as a result
* This provides an alternative fix to the issue I was having in https://github.com/AstroNvim/astrocommunity/pull/837

### DIFF
--- a/lua/astrocommunity/utility/mason-tool-installer-nvim/init.lua
+++ b/lua/astrocommunity/utility/mason-tool-installer-nvim/init.lua
@@ -24,11 +24,17 @@ return {
           if plugin == "mason-lspconfig.nvim" then
             mappings = require("mason-lspconfig").get_mappings().lspconfig_to_mason
           elseif plugin == "mason-null-ls.nvim" then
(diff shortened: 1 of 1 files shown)


Uzaaft:
@mehalter  You're the expert on this stuff, more than me 😅 Can you take a look whenever you get time and see if this makes sense?

mehalter (approved):
Thanks! LGTM!

PR #845, opened 2024-04-08
added dependency FixCursorHold

PatMulligan:
## 📑 Description
Recommendation per [neotest installation]](https://github.com/nvim-neotest/neotest?tab=readme-ov-file#installation)
> Neotest uses the CursorHold event. This uses the updatetime setting which is by default very high, and lowering this can lead to excessive writes to disk. It's recommended to use https://github.com/antoinemadec/FixCursorHold.nvim which allows detaching updatetime from the frequency of the CursorHold event. The repo claims it is no longer needed but it is still recommended (See this issue)


Uzaaft:
This is not needed for neovim v0.9

PR #851, opened 2024-04-09
feat(vscode-icons): add more and update lspkind icons

manuuurino:
## 📑 Description

Ive noticed that the lspkind icons look kinda outdated? So thats why i would like to update them, since i saw the config from kabin.

## ℹ Additional Information

Snippet from kabin's config: https://github.com/kabinspace/AstroNvim_user/blob/c114034e98a5291c1d4369f74143f21ba2e28079/lua/plugins/lspkind.lua#L7-L31


Uzaaft (approved):

mehalter:
This was incorrect. I will revert.

mehalter:
All of the icons that this got changed to _are_ the default `codicons` when setting the preset: https://github.com/onsails/lspkind.nvim/blob/master/lua/lspkind/init.lua#L34-L60

We add the kinds that are _missing_ from that list. Which are up to date and correct from `codicons`

mehalter:
Reverted with https://github.com/AstroNvim/astrocommunity/commit/df5475b047290e880a8cc02fb8d36d65eec8a5a1

PR #853, opened 2024-04-09
docs(rust): Update docs to reflect the pack

Reun85:
# Update Rust pack README
## 📑 Description
 [] Not Completed
--> [x] Completed

The Rust pack already uses the rustacenvim plugin instead of the rust-tools.nvim plugin, since that is deprecated.
The configuration uses cargo clippy, the most sensible default in my opinion, the paragraph detailing how it can be setup is useless.


Uzaaft:
@Reun85  Please read the contrib guide next time

Uzaaft (approved):

PR #855, opened 2024-04-09
feat(lsp): add `actions-preview.nvim` plugin

manuuurino:


lua/astrocommunity/lsp/actions-preview-nvim/init.lua
```diff
@@ -0,0 +1,28 @@
...
+    function() require("actions-preview").code_actions() end,
+    desc = "LSP code action",
+    cond = "testDocument/codeAction",
+  },
+}
+
+local mappings = {
+  n = keymaps,
+  v = keymaps,
+}
```
Uzaaft, line 12: We should do this similar to other plugins with similarly small mappings

Uzaaft (changes requested):
See comment

lua/astrocommunity/lsp/actions-preview-nvim/init.lua
manuuurino: you mean something similar like this? https://github.com/AstroNvim/astrocommunity/blob/f2d802f9514e1b37f24289029451aa604075f931/lua/astrocommunity/motion/nvim-spider/init.lua#L1-L19

lua/astrocommunity/lsp/actions-preview-nvim/init.lua
```diff
@@ -0,0 +1,28 @@
...
+
+---@type LazySpec
+return {
+  "aznhe21/actions-preview.nvim",
+  event = "User AstroLspSetup",
```
mehalter, line 17:
```suggestion
  lazy = true,
```

This should be much more aggressively lazy loaded. And only get loaded by module since it's fully driven by a single keybind.

mehalter (changes requested):

lua/astrocommunity/lsp/actions-preview-nvim/init.lua
Uzaaft: More like(an example: 
```lua
{
      "AstroNvim/astrocore",
      opts = {
        mappings = {
          n = {
            ["<Leader>fe"] = { "<Cmd>Telescope file_browser<CR>", desc = "Open File browser" },
          },
          v = {
            ["<Leader>fe"] = { "<Cmd>Telescope file_browser<CR>", desc = "Open File browser" },
          },
        },
      },
},
```

manuuurino:
I also want to add an information, that this binding will override the lsp code action binding with this one, should i perhaps mention it in the readme? @Uzaaft

mehalter (approved):
LGTM!

PR #856, opened 2024-04-09
perf: set appropiate/missing lazy loading events

manuuurino:
## 📑 Description

Noticed some lazy loading improvements from my personal config. These were the one i am using, sorry but i cannot check all plugins, i hope i can improve the startup/running performance a tiny bit further.

### DIFF
--- a/lua/astrocommunity/editing-support/true-zen-nvim/init.lua
+++ b/lua/astrocommunity/editing-support/true-zen-nvim/init.lua
@@ -1,5 +1,6 @@
 return {
   "Pocco81/true-zen.nvim",
+  lazy = true,
   dependencies = {
     {
       "AstroNvim/astrocore",
--- a/lua/astrocommunity/editing-support/vim-move/init.lua
+++ b/lua/astrocommunity/editing-support/vim-move/init.lua
@@ -1,4 +1,4 @@
 return {
   "matze/vim-move",
-  event = "BufEnter",
+  event = { "User AstroFile", "InsertEnter" },
 }
--- a/lua/astrocommunity/lsp/lsp-inlayhints-nvim/init.lua
+++ b/lua/astrocommunity/lsp/lsp-inlayhints-nvim/init.lua
@@ -1,6 +1,7 @@
 return {
   {
     "lvimuser/lsp-inlayhints.nvim",
+    lazy = true,
     dependencies = {
       "AstroNvim/astrocore",
       opts = {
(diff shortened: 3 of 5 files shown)


lua/astrocommunity/pack/yaml/init.lua
```diff
@@ -1,7 +1,8 @@
 return {
   {
     "b0o/SchemaStore.nvim",
-    lazy = true,
+    event = "User AstroFile",
+    ft = { "yaml", "json" },
```
Uzaaft, line 5: Shouldn't toml also be added?

Uzaaft: My bad, taplo has it built in by default.

lua/astrocommunity/editing-support/true-zen-nvim/init.lua
```diff
@@ -1,5 +1,6 @@
 return {
   "Pocco81/true-zen.nvim",
+  event = "VeryLazy",
```
mehalter, line 3:
```suggestion
```



This should not be lazy loaded on event. It is lazy loaded on module and keybinding

lua/astrocommunity/editing-support/vim-move/init.lua
```diff
@@ -1,4 +1,4 @@
 return {
   "matze/vim-move",
-  event = "BufEnter",
+  event = "User AstroFile",
```
mehalter, line 3:
```suggestion
  event = { "User AstroFile", "InsertEnter" },
```

This lets it work on new files as well

lua/astrocommunity/editing-support/true-zen-nvim/init.lua
luxus, line 1: will it not get loaded when using the mappings?

lua/astrocommunity/lsp/lsp-inlayhints-nvim/init.lua
```diff
@@ -1,6 +1,7 @@
 return {
   {
     "lvimuser/lsp-inlayhints.nvim",
+    event = "User AstroLspSetup",
```
mehalter, line 4:
```suggestion
```

This lazy loading is handled by the autocmd and should not be loaded on event. This is less performant

lua/astrocommunity/pack/yaml/init.lua
mehalter, line 5:
```suggestion
    lazy = true,
```

this should be reverted. This is more performant as is. It is lazy loaded based on module only when needed.

mehalter (changes requested):
A few of these need to be reverted because they are significantly less performant. Other than that, thanks for the recommendations!

mehalter (approved):
I went in and resolved the performance issues. Thanks!

PR #857, opened 2024-04-09
feat(editing-support): add `mini-operators` plugin

manuuurino:
## 📑 Description

Making this a draft PR, because i am unsure which binding would match it neatly. I left there also an note on the source code.

### DIFF
--- a/lua/astrocommunity/editing-support/mini-operators/init.lua
+++ b/lua/astrocommunity/editing-support/mini-operators/init.lua
@@ -1,33 +1,25 @@
--- NOTE: the default prefix is `g`, but this will override some bindings.
-local prefix = "g"
-
 ---@type LazySpec
 return {
   "echasnovski/mini.operators",
-  event = "User AstroFile",
-  dependencies = {
-    "AstroNvim/astrocore",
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/editing-support/mini-operators/init.lua
```diff
@@ -0,0 +1,31 @@
...
+--       the default prefix would be g, but this will also override some bindings.
+local prefix = "<Leader>o"
+
+local mappings = {
+  n = {
+    [prefix] = {
+      name = "Text edit operators",
+    },
+  },
+}
```
Uzaaft, line 11: 1. I think overriding the neo-tree default is a no-go. @AstroNvim/astrocommunity-maintainers  Any ideas? 
2. Similar to other reviews: Lets use the same method for mappings as we do most of the other plugins.

mehalter: I agree not to override default neo-tree operator

mehalter (commented):
I will take over this PR and get it ready for merge

mehalter (approved):

PR #858, opened 2024-04-09
feat(editing-support): add `nvim-context-vt`

manuuurino:


lua/astrocommunity/lsp/nvim-context-vt/init.lua
```diff
@@ -0,0 +1,19 @@
+---@type LazySpec
+return {
+  "andersevenrud/nvim_context_vt",
+  event = "User AstroFile",
```
Uzaaft, line 4:
```suggestion
  event = "User AstroFile",
  cmd = {"NvimContextVtToggle"},
```

Uzaaft (changes requested):
The cmd should be added

Uzaaft (approved):

Uzaaft:
@manuuurino  This plugin should be in editing-support imo.

mehalter (approved):

PR #862, opened 2024-04-10
fix(python): incorrect path for windows debugpy

lucklyisgood:
## 📑 Description

fix incorrect path for windows debugpy.
The path for installing debugpy on Mason under Windows is`~/<mason pkg path>/debugpy/venv/Scripts/python.exe`


mehalter (approved):
LGTM

PR #864, opened 2024-04-10
fix(blame-nvim): Align the git blame window properly when using heirline

rameshsanth:
## 📑 Description

Before fix

Screenshot 2024-04-10 at 10 34 17 AM

After fix
Screenshot 2024-04-10 at 10 35 21 AM


mehalter (approved):
LGTM!

PR #870, opened 2024-04-11
fix(harpoon): deprecated function in configuration

amqndin:
## 📑 Description
replaced deprecated function `append` with updated function `add`.


mehalter (approved):
Thanks!

PR #880, opened 2024-04-14
fix(java): fix mason package `clang_format` not found

manuuurino:
## 📑 Description

there is a typo for the mason package `clang_format`, the correct package name is `clang-format`.

## ℹ Additional Information

image

### DIFF
--- a/lua/astrocommunity/pack/java/init.lua
+++ b/lua/astrocommunity/pack/java/init.lua
@@ -19,7 +19,7 @@ return {
     "jay-babu/mason-null-ls.nvim",
     optional = true,
     opts = function(_, opts)
-      opts.ensure_installed = require("astrocore").list_insert_unique(opts.ensure_installed, { "clang_format" })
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/pack/java/init.lua
```diff
@@ -19,7 +19,7 @@ return {
     "jay-babu/mason-null-ls.nvim",
     optional = true,
     opts = function(_, opts)
-      opts.ensure_installed = require("astrocore").list_insert_unique(opts.ensure_installed, { "clang_format" })
+      opts.ensure_installed = require("astrocore").list_insert_unique(opts.ensure_installed, { "clang-format" })
```
Uzaaft, line 22: The null-ls formatter is called clang_format. When using `:NullLsInstall` it appears as `clang_format` in the list.

manuuurino: ah ic, i somehow had it in my memory that it didnt worked on my work pc and had to fix it. ig i forgot to update then, sry to bother

PR #881, opened 2024-04-14
fix(dial-nvim): lazify dial-nvim, add month / weekday crementing

Axlefublr:
- **lazify plugin**
- **add increment / decrement month / weekday**

## 📑 Description

Since we already add mappings, we might as well lazy-load the plugin

The reasoning behind putting the `g` mappings in visual mode only, is that `v` maps to both visual mode and select mode.
Select mode is used in luasnip, for insert nodes that have text inside of them.
The user might expect to be able to just type in the letter g, but instead these two dial mappings will hold on the user pressing g, rather than immediatley continuing.
Can't think of another reason to be in select mode aside from that, so I don't see a disadvantage in making the experience less annoying with luasnip.

I found it super useful to be able to decrement / increment month names and weekdays, so, I added it as well.


lua/astrocommunity/editing-support/dial-nvim/init.lua
```diff
@@ -1,5 +1,27 @@
 return {
   "monaqa/dial.nvim",
```
Uzaaft, line 2: @Axlefublr Does adding lazy = true break anything ?

Axlefublr: At least doesn't for me. Even with `lazy = true`, the mappings work the first time I press them, as expected.

Uzaaft:
```suggestion
  "monaqa/dial.nvim",
  lazy = true,
```

Uzaaft: Feel free to add it then :)

Axlefublr: sure! but doesn't `keys` already make it lazy? are we adding `lazy = true` as a sanity check?

Uzaaft: I am stupid.

Uzaaft (approved):

PR #888, opened 2024-04-16
fix(pack-java): remove clang-formatter

Juniar-Rakhman:
## 📑 Description

In the java pack, clang_formatter is unnecessary. It also overrides the formatting capability of "mfussenegger/nvim-jdtls". 

Having clang_formatter for java pack does not make sense, as the default formatter works just fine.

list of changes:
- [x] removed unnecessary clang_formatter and everything related to it. 
- [x] updated readme to include custom formatter example.

### DIFF
--- a/lua/astrocommunity/pack/java/README.md
+++ b/lua/astrocommunity/pack/java/README.md
@@ -11,20 +11,18 @@ This plugin pack does the following:
 - Adds `jdtls` language server
 - Adds `lemminx` language server (XML)
 - Adds [nvim-jdtls](https://github.com/mfussenegger/nvim-jdtls) for language specific tooling
-  - Enables [hot reloading](https://github.com/mfussenegger/nvim-jdtls/issues/80)
-- Adds `clang_format` through null-ls
+- Enables [hot reloading](https://github.com/mfussenegger/nvim-jdtls/issues/80)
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/pack/java/README.md
```diff
@@ -43,7 +41,14 @@ We require that the root folder of your projects inlude one of these files/folde
             },
           },
         },
+        format = {
+          enabled = true,
+          settings = { -- you can use your preferred format style
+            url = os.getenv "HOME" .. ".config/nvim/formatter/eclipse-java-google-style.xml",
```
mehalter, line 47: this needs to be changed since this file is not guaranteed to exist

mehalter (changes requested):

lua/astrocommunity/pack/java/README.md
mehalter:
```suggestion
            url = "https://raw.githubusercontent.com/google/styleguide/gh-pages/eclipse-java-google-style.xml",
```

This seems to be a good default from the `vscode-java` docs: https://github.com/redhat-developer/vscode-java/wiki/Formatter-settings

Juniar-Rakhman: Agree

mehalter (approved):
LGTM

PR #889, opened 2024-04-16
fix(neogen): keymap for type-aware automatic comment generation

CMOISDEAD:
Keyboard shortcut **(leader + a + CR)** for automatic comment generation did not correctly detect element type (function, class, etc.). This commit fixes the bug so that the shortcut correctly detects element type and generates the appropriate comment.

## 📑 Description

The "keymap" **(leader + a + CR)** that triggers the automatic neogen generation does not work and throws an error message saying "current type is not supported".

### current behavior

https://github.com/AstroNvim/astrocommunity/assets/51010598/e9d9cabf-54be-4b6e-a7c0-ca116e988fcb

### expected behavior

https://github.com/AstroNvim/astrocommunity/assets/51010598/5d63fa71-c803-48f0-aa32-50dfcf0e9da8


mehalter (approved):
Thanks!

PR #890, opened 2024-04-17
fix(sql): drop `sqlfluff` as `sqls` supports formatting

mehalter:
## 📑 Description

`sqls` supports formatting so we don't need to add `sqlfluff`.


Uzaaft:
@mehalter  Technically, it's a linter. (I know, it's my mistake 😂)

mehalter:
It also provides formatting and it can conflict with the formatting of `sqls`. It also seems to mainly just be giving people problems rather than being particularly useful

mehalter:
we'll just leave it 🤷🏼

Uzaaft (approved):

PR #892, opened 2024-04-17
fix(blame-nvim): adjust to the new rewrite

manuuurino:
## 📑 Description

The blame.nvim plugin has been rewritten and "broke" this spec. Updated to this state https://github.com/FabijanZulj/blame.nvim/tree/39f6123b78bde45917de071a07eba17b136d9f2c.

### DIFF
--- a/lua/astrocommunity/git/blame-nvim/init.lua
+++ b/lua/astrocommunity/git/blame-nvim/init.lua
@@ -1,7 +1,7 @@
 return {
   "FabijanZulj/blame.nvim",
   cmd = "BlameToggle",
-  config = function(_, opts) require("blame").setup(opts) end,
+  opts = {},
   dependencies = {
     {
       "AstroNvim/astrocore",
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/git/blame-nvim/init.lua
```diff
@@ -1,6 +1,7 @@
 return {
   "FabijanZulj/blame.nvim",
-  event = "User AstroGitFile",
+  cmd = "BlameToggle",
+  config = function(_, opts) require("blame").setup(opts) end,
```
mehalter, line 4:
```suggestion
  opts = {}
```

No need to redefine the default `config` function

mehalter (changes requested):
Small nit

lua/astrocommunity/git/blame-nvim/init.lua
manuuurino: Thanks, teached me something new :)!

mehalter (approved):
LGTM Thanks!

PR #898, opened 2024-04-19
docs(rust): Remove unnecessary disclaimer

RigoOnRails:
## 📑 Description

Hi, I don't really think this is necessary to add to the README.


Uzaaft:
I think it can stay, one never know.
Cc @mehalter

mehalter (approved):

PR #900, opened 2024-04-20
feat(cpp): Add `<Leader>lw` keymap to switch between source and header

rameshsanth:
<leader>lw to sWitch between source and header.

"w" isn't assigned to any other key currently.

## ℹ Additional Information

Caveat: The keybinding will stay for all LSP although only Clangd supports it. Ideally the keybinding should appear only for buffers Clangd is attached.

### DIFF
--- a/lua/astrocommunity/pack/cpp/init.lua
+++ b/lua/astrocommunity/pack/cpp/init.lua
@@ -46,14 +46,6 @@ return {
                 if assert(vim.lsp.get_client_by_id(args.data.client_id)).name == "clangd" then
                   require "clangd_extensions"
                   vim.api.nvim_del_augroup_by_name "clangd_extensions"
-                  require("astrocore").set_mappings({
-                    n = {
-                      ["<Leader>lw"] = {
-                        "<cmd>ClangdSwitchSourceHeader<CR>",
-                        desc = "Clangd: Switch source/header file",
-                      },
-                    },
-                  }, {})
                 end
               end,
             },
--- a/lua/astrocommunity/pack/cpp/init.lua
+++ b/lua/astrocommunity/pack/cpp/init.lua
@@ -46,6 +46,12 @@ return {
                 if assert(vim.lsp.get_client_by_id(args.data.client_id)).name == "clangd" then
                   require "clangd_extensions"
                   vim.api.nvim_del_augroup_by_name "clangd_extensions"
+                  vim.keymap.set(
+                    "n",
+                    "<Leader>lw",
+                    "<cmd>ClangdSwitchSourceHeader<CR>",
+                    { silent = true, desc = "Clangd: Switch source/header file" }
+                  )
                 end
               end,
             },
--- a/lua/astrocommunity/pack/cpp/init.lua
+++ b/lua/astrocommunity/pack/cpp/init.lua
@@ -58,6 +58,21 @@ return {
               end,
             },
           },
+          clangd_extension_mappings = {
+            {
+              event = "LspAttach",
+              desc = "Load clangd_extensions with clangd",
+              callback = function(args)
+                if assert(vim.lsp.get_client_by_id(args.data.client_id)).name == "clangd" then
+                  require("astrocore").set_mappings({
+                    n = {
+                      ["<Leader>lw"] = { "<Cmd>ClangdSwitchSourceHeader<CR>", desc = "Switch source/header file" },
+                    },
+                  }, { buffer = args.buf })
+                end
+              end,
+            },
+          },
         },
       },
     },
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/pack/cpp/init.lua
```diff
@@ -46,6 +46,12 @@ return {
                 if assert(vim.lsp.get_client_by_id(args.data.client_id)).name == "clangd" then
                   require "clangd_extensions"
                   vim.api.nvim_del_augroup_by_name "clangd_extensions"
+                  vim.keymap.set(
```
Uzaaft, line 49: Please use astrocore to set keymaps

Uzaaft (dismissed):
See feedback

lua/astrocommunity/pack/cpp/init.lua
rameshsanth: https://github.com/AstroNvim/astrocore/blob/main/doc/api.md#set_mappings
Did you mean this api ?

rameshsanth:
Caveat: The keybinding will stay for all LSP although only Clangd supports it. Ideally the keybinding should appear only for buffers Clangd is attached.
If you know the trick to fix make the key map only to C/C++ files, let me know. Otherwise its a minor hassle

Uzaaft:
> Caveat: The keybinding will stay for all LSP although only Clangd supports it. Ideally the keybinding should appear only for buffers Clangd is attached. If you know the trick to fix make the key map only to C/C++ files, let me know. Otherwise its a minor hassle

We'll have to fix this.

Uzaaft:
@rameshsanth  I'm pretty sure we have the logic for it somewhere in astrocommunity. :)

Uzaaft:
Marking this as a draft until we fix it.

rameshsanth:
> Marking this as a draft until we fix it.

IMHO, this is not a major issue. There are several key maps that don't apply for every buffer. However it's still loaded and available in vim.
When you get use this short cut on a non CPP file you see this notification. 
Screenshot 2024-04-20 at 12 31 37 AM

Uzaaft:
We should add an autocmd for cleaning up. 
For semantic token, we've used this before:
```lua
vim.api.nvim_create_autocmd("lspattach", {
  callback = function(args)
    local client = vim.lsp.get_client_by_id(args.data.client_id)
    client.server_capabilities.semantictokensprovider = nil
  end
```
Should use same logic, just with the astrocore functionality instead.

lua/astrocommunity/pack/cpp/init.lua
```diff
@@ -46,6 +46,14 @@ return {
...
                   require "clangd_extensions"
                   vim.api.nvim_del_augroup_by_name "clangd_extensions"
+                  require("astrocore").set_mappings({
+                    n = {
+                      ["<Leader>lw"] = {
+                        "<cmd>ClangdSwitchSourceHeader<CR>",
+                        desc = "Clangd: Switch source/header file",
+                      },
+                    },
+                  }, {})
```
mehalter, line 56:
```suggestion
                  }, { buffer = args.buf })
```

Only set mappings for buffer that clangd is attached to

mehalter (changes requested):

lua/astrocommunity/pack/cpp/init.lua
mehalter: actually this isn't a super great method because this will only work for the first language server. one sec

mehalter, line 56:
```suggestion
```

This is only executed once so it isn't a good place to put this mapping.

lua/astrocommunity/pack/cpp/init.lua
```diff
@@ -46,6 +46,14 @@ return {
...
                 end
               end,
             },
```
mehalter, line 59:
```suggestion
            },
          },
          clangd_extension_mappings = {
            {
              event = "LspAttach",
              desc = "Load clangd_extensions with clangd",
              callback = function(args)
                if assert(vim.lsp.get_client_by_id(args.data.client_id)).name == "clangd" then
                  require("astrocore").set_mappings({
                    n = {
                      ["<Leader>lw"] = { "<Cmd>ClangdSwitchSourceHeader<CR>", desc = "Switch source/header file" },
                    },
                  }, { buffer = args.buf })
                end
              end,
            },
```

Create a new mapping that attaches with each `clangd` that doesn't delete itself and handles creating the buffer local mapping

mehalter (changes requested):

Uzaaft:
@rameshsanth Can you test to see if this works as expected now? :)

mehalter (approved):

rameshsanth:
Sorry I got busy at work. Thanks for completing this. it is working correctly.

PR #901, opened 2024-04-20
feat(note-taking): Add venn-nvim

rameshsanth:
Draw ASCII diagrams easily in Neovim

## 📑 Description

Added new plugin with necessary configuration per the author

### DIFF
--- a/lua/astrocommunity/note-taking/venn-nvim/README.md
+++ b/lua/astrocommunity/note-taking/venn-nvim/README.md
@@ -2,22 +2,24 @@
 
 Draw ASCII diagrams in Neovim with ease.
 
-    1. Enable venn mode, :lua Toggle_Venn<cr>
-       - note this enable vim virtual edit mode which allows editing anywhere in the window
-    2. now click any place and write down the texts
-    3. use visual block [Ctrl-v] mode to wrap around the text and press f to draw the box
-    4. connect the boxes using HJKL towards the next box to connect
-                            ┌───┐
-                   ┌────────│ A │──────────┐
-                   │        └───┘          │
-                   │                       │
-                   ▼        ┌───┐          ▼
-                   B───────►│ C │          D
-                            └─┬─┘          │
-                              │            │
-                            ┌─┴─┐          │
-                            │ E │◄─────────┘
-                            └───┘
+1. Enable venn mode with `:ToggleVenn`
+   - note this enable vim virtual edit mode which allows editing anywhere in the window
+2. now click any place and write down the texts
+3. use visual block [Ctrl-v] mode to wrap around the text and press f to draw the box
+4. connect the boxes using HJKL towards the next box to connect
+
+```
+                        ┌───┐
+               ┌────────│ A │──────────┐
+               │        └───┘          │
+               │                       │
+               ▼        ┌───┐          ▼
+               B───────►│ C │          D
+                        └─┬─┘          │
+                          │            │
+                        ┌─┴─┐          │
+                        │ E │◄─────────┘
+                        └───┘
+```
 
 **Repository:** <https://github.com/jbyuki/venn.nvim>
-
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/note-taking/venn-nvim/init.lua
```diff
@@ -0,0 +1,39 @@
...
+        vim.cmd [[setlocal ve=]]
+        vim.api.nvim_buf_del_keymap(0, "n", "J")
+        vim.api.nvim_buf_del_keymap(0, "n", "K")
+        vim.api.nvim_buf_del_keymap(0, "n", "L")
+        vim.api.nvim_buf_del_keymap(0, "n", "H")
+        vim.api.nvim_buf_del_keymap(0, "v", "f")
+        vim.b.venn_enabled = nil
+      end
+    end
+  end,
```
Uzaaft, line 38: Please set the mappings with astrocore.

Uzaaft (changes requested):
See comments.

rameshsanth:
> See comments.

how do you remove mapping. set_mappings seems to only set ?

lua/astrocommunity/note-taking/venn-nvim/init.lua
rameshsanth: I've pushed the changes to set the mappings. However there is no astrocore API to remove the mappings. So leaving the remove part as is.

mehalter:
I'll handle this :)

mehalter:
@rameshsanth I modified this implementation a bit to integrate a bit more seamlessly with AstroCore and make the mapping more easily configurable by the user. This also makes it a bit easier to start venn mode by setting up a user command `:VennToggle`.

It's also important to note that this plugin's method of setting up the buffer mappings is pretty bad and comes in and clobbers any previous buffer local mappings such as `K` which is used for LSP documentation. So any buffer that uses a language server will have at least that mapping borked after toggling venn mode. Not sure the best way to fix this other than caching already existing mappings on enable and restoring them on disable. Not sure if you want to implement that or not. I think as is this is fine enough to merge since it implements exactly what the README for that plugin recommends even though it's a bad recommendation.

mehalter:
@Uzaaft this is ready for review

Uzaaft:
> @Uzaaft this is ready for review

Reviewing this soon

Uzaaft (approved):

lua/astrocommunity/note-taking/venn-nvim/README.md
```diff
@@ -0,0 +1,25 @@
...
+               ▼        ┌───┐          ▼
+               B───────►│ C │          D
+                        └─┬─┘          │
+                          │            │
+                        ┌─┴─┐          │
+                        │ E │◄─────────┘
+                        └───┘
+```
+
+**Repository:** <https://github.com/jbyuki/venn.nvim>
```
Uzaaft, line 25:
```suggestion
# venn.nvim

Draw ASCII diagrams in Neovim with ease.

**Repository:** <https://github.com/jbyuki/venn.nvim>

1. Enable venn mode with `:ToggleVenn`
   - note this enable vim virtual edit mode which allows editing anywhere in the window
2. now click any place and write down the texts
3. use visual block [Ctrl-v] mode to wrap around the text and press f to draw the box
4. connect the boxes using HJKL towards the next box to connect

```
                        ┌───┐
               ┌────────│ A │──────────┐
               │        └───┘          │
               │                       │
               ▼        ┌───┐          ▼
               B───────►│ C │          D
                        └─┬─┘          │
                          │            │
                        ┌─┴─┐          │
                        │ E │◄─────────┘
                        └───┘
```

```

Uzaaft (approved):

mehalter:
@rameshsanth I actually just added [Hydra.nvim](https://github.com/anuvyklack/hydra.nvim) to AstroCommunity in a way that it can be integrated with other plugins. I modified this config so that if hydra is available it defines the diagramming mode as a submode in Neovim (As described here: https://github.com/anuvyklack/hydra.nvim/wiki/Draw-diagrams)

If you want to try it out simply add hydra from AstroCommunity :)

PR #907, opened 2024-04-21
feat(typst): added typst-preview.nvim plugin support

### LINKED ISSUE #849: Typst: typst-preview integration

With long texts it's essential to be capable of finding the results of editing.

### Describe the new feature

Integration of [typst-preview](https://github.com/Enter-tainer/typst-preview). It seems to be possible since it has a command line interface, besides being VSCode plugin.

### Additional context

_No response_

Std-Enigma:
## 📑 Description
Adds preview support for typst through [typst-preview.nvim](https://github.com/chomosuke/typst-preview.nvim)

### DIFF
--- a/lua/astrocommunity/pack/typst/init.lua
+++ b/lua/astrocommunity/pack/typst/init.lua
@@ -15,7 +15,7 @@ return {
   },
   {
     "chomosuke/typst-preview.nvim",
-    ft = "typst",
+    cmd = { "TypstPreview", "TypstPreviewToggle" },
     version = "0.1.*",
     build = function() require("typst-preview").update() end,
     opts = {},
--- a/lua/astrocommunity/pack/typst/init.lua
+++ b/lua/astrocommunity/pack/typst/init.lua
@@ -15,8 +15,9 @@ return {
   },
   {
     "chomosuke/typst-preview.nvim",
-    cmd = { "TypstPreview", "TypstPreviewToggle" },
+    ft = "typst",
     version = "0.1.*",
     build = function() require("typst-preview").update() end,
+    opts = {},
   },
 }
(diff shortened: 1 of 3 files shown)


Uzaaft:
@Std-Enigma  The typo fix should come in a seperate PR.

lua/astrocommunity/pack/typst/init.lua
```diff
@@ -13,4 +13,10 @@ return {
       opts.ensure_installed = require("astrocore").list_insert_unique(opts.ensure_installed, { "typst-lsp" })
     end,
   },
+  {
+    "chomosuke/typst-preview.nvim",
+    cmd = { "TypstPreview", "TypstPreviewToggle" },
+    version = "0.1.*",
```
Uzaaft, line 19: This should have an empty opts table

Uzaaft: Should we perhaps lazy load this on typst files?

Std-Enigma: that does make sense but at the end of they if they want to enable it they have to run the command so i don't think that would make that much difference maybe the start up time would be effected since it has to load it right a way in this case?

Uzaaft: What is the command for typst-preview.nvim?

Std-Enigma: I did change it to lazy load on filetype but I do believe it doesn't make that much difference since for enabling the preview you have to use the command anyway

Std-Enigma: > What is the command for typst-preview.nvim?

It has quite a bunch but the ones that are capable of starting the functionality are: TypstPreview and TypstPreviewToggle

Std-Enigma: it's now back to lazy load on command

Uzaaft (approved):

PR #912, opened 2024-04-25
fix(tailwindcss): add root_dir detection to tailwindcss pack

rafaelderolez:
## 📑 Description

The tailwindcss-language-server is currently always up and running. I believe it should only run when a Tailwind CSS config is found. 

This PR fixes that.

### DIFF
--- a/lua/astrocommunity/pack/tailwindcss/init.lua
+++ b/lua/astrocommunity/pack/tailwindcss/init.lua
@@ -12,20 +12,18 @@ return {
         optional = true,
         ---@type AstroLSPOpts
         opts = {
-          handlers = {
-            tailwindcss = function()
-              require("lspconfig").tailwindcss.setup {
-                root_dir = function(fname)
-                  local root_pattern = require("lspconfig").util.root_pattern(
-                    "tailwind.config.cjs",
-                    "tailwind.config.js",
-                    "tailwind.config.ts",
-                    "postcss.config.js"
-                  )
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/pack/tailwindcss/init.lua
```diff
@@ -6,6 +6,30 @@ return {
...
+                    "tailwind.config.js",
+                    "tailwind.config.ts",
+                    "postcss.config.js"
+                  )
+                  return root_pattern(fname)
+                end,
+              }
+            end,
+          },
+        },
```
mehalter, line 30:
```suggestion
        opts = {
          config = {
            tailwindcss = {
              root_dir = function(fname)
                local root_pattern = require("lspconfig").util.root_pattern(
                  "tailwind.config.cjs",
                  "tailwind.config.js",
                  "tailwind.config.ts",
                  "postcss.config.js"
                )
                return root_pattern(fname)
              end,
            },
          },
        },
```

This should be done through the config table. This currently breaks the integration with AstroLSP like setting up our capabilities/on_attach as well as any configuration done by the user

mehalter (changes requested):

lua/astrocommunity/pack/tailwindcss/init.lua
rafaelderolez: @mehalter thanks!

PR #913, opened 2024-04-25
fix(typst): added missing command to typst-preview.nvim

Std-Enigma:
## 📑 Description

This adds the ability to manually build the binaries for typst-preview.nvim in case it have failed doing so with lazy


mehalter (approved):
Thanks!

PR #915, opened 2024-04-25
fix(tailwindcss): fix TailwindCSS pack not loading when using tailwindcss-rails Ruby Gem

natematykiewicz:
## 📑 Description

#912 caused the Tailwind LSP to not work, when using the tailwindcss-rails Ruby gem.

The tailwindcss-rails Ruby gem keeps the config at config/tailwind.config.js, which is not in the new `root_dir` config.

https://github.com/rails/tailwindcss-rails?tab=readme-ov-file#configuration-and-commands

### DIFF
--- a/lua/astrocommunity/pack/tailwindcss/init.lua
+++ b/lua/astrocommunity/pack/tailwindcss/init.lua
@@ -19,7 +19,8 @@ return {
                   "tailwind.config.cjs",
                   "tailwind.config.js",
                   "tailwind.config.ts",
-                  "postcss.config.js"
+                  "postcss.config.js",
+                  "config/tailwind.config.js"
                 )
                 return root_pattern(fname)
               end,


mehalter (approved):
LGTM!

rafaelderolez:
Sorry about that!

This actually made me go look at the docs and it seems custom config location and names are supported: https://tailwindcss.com/docs/configuration#using-a-different-file-name.

We could consider adding a statement to check package.json for a `tailwindcss` key, and the equivalent for Ruby projects (and others?).

Thoughts? @mehalter @natematykiewicz

natematykiewicz:
FWIW, the official Tailwind guides mention this config file. 

https://tailwindcss.com/docs/guides/ruby-on-rails

treshenry:
And... Elixir Phoenix projects put the tailwind config at assets/tailwind.config.js. https://github.com/phoenixframework/tailwind

Could not figure out what broke, and had to go digging. 🤷‍♂️  I'll put up a PR shortly.

PR #919, opened 2024-04-26
feat(hehe): add `Hydra.nvim`

mehalter:
## 📑 Description

This adds `Hydra.nvim` and provides an API for configuring it through the `opts` table. This could be useful for AstroCommunity entries to create submenus if hydra is available

I removed the automatic lazy loading based on keys because it is hyper unstable based on how the `body` + `heads` work. It's also not very slow to load so it's ok to just load it on `VeryLazy` like we do which-key

### DIFF
--- a/lua/astrocommunity/keybinding/hydra-nvim/README.md
+++ b/lua/astrocommunity/keybinding/hydra-nvim/README.md
@@ -1,0 +1,35 @@
+# Hydra.nvim
+
+Create custom submodes and menus
+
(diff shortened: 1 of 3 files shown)


owittek (approved):
LGTM 🚀

Uzaaft:
@mehalter  Should the title mention the integration with venv-nvim?

mehalter:
here @Uzaaft I changed the title to break the CI/CD, I will merge this manually

PR #927, opened 2024-04-29
Mason LSP buf was rename to buf-language-server

JADC362:
## 📑 Description

While trying to import the `pack.proto` file, Mason failed to install the `bufls` LSP package. It seems like the package was rename as `buf-language-server`:
```
Available (223)
    ...
    ✗ buf-language-server bufls (keywords: protobuf)
```

### DIFF
--- a/lua/astrocommunity/pack/proto/init.lua
+++ b/lua/astrocommunity/pack/proto/init.lua
@@ -13,7 +13,7 @@ return {
     "williamboman/mason-lspconfig.nvim",
     optional = true,
     opts = function(_, opts)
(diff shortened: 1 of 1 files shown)


mehalter:
This is the name of the language server in `lspconfig` and not the name of the mason package. There is no bug here and the pack works as intended. Thanks for looking out though!

mehalter:
If you are getting errors adding `bufls` then you may be missing a dependency of some sort or the package is failing to install. Check Mason logs for details

JADC362:
Forgive me @mehalter, you were right. I didn't look at the log error msg: `golang` was a required dependency that I didn't have on my computer. Thanks!

PR #937, opened 2024-05-02
fix(templ): Use astrocore instead of `astronvim.utils`

Uzaaft:


luxus (approved):

mehalter:
This is still incorrect. I'll fix in a follow up PR

mehalter:
Fixed for reals here: https://github.com/AstroNvim/astrocommunity/commit/f44fdf056734f01c18f6b334a9bd45bb4c75bb74

PR #942, opened 2024-05-06
coc-nvim: better integration

wroyca:
## 📑 Description

This PR is a draft to add better defaults to coc.nvim and fix some compatibilities with plugins (e.g. Neodev). There's a lot of area that I'm not sure how to approach due to AstroNvim abstraction, so if anyone wants to review, it's probably best to go ahead by pushing commits directly on the PR without worries, I will sync and rebase locally

Remaining things to do:
- [ ] figure out how to move coc#config to opts table somehow 
- [x] neodev
- [ ] catppuccin
- [x] pack/cpp
- [ ] ?

### DIFF
--- a/lua/astrocommunity/lsp/coc-nvim/init.lua
+++ b/lua/astrocommunity/lsp/coc-nvim/init.lua
@@ -178,6 +178,12 @@ return {
       -- }
     end,
   },
+  {
+    "catppuccin/nvim",
+    optional = true,
+    ---@type CatppuccinOptions
+    opts = { integrations = { coc_nvim = true } },
+  },
   -- disable core lsp plugins
   { "AstroNvim/astrolsp", enabled = false },
   { "folke/neoconf.nvim", enabled = false },
--- a/lua/astrocommunity/lsp/coc-nvim/init.lua
+++ b/lua/astrocommunity/lsp/coc-nvim/init.lua
@@ -24,6 +24,37 @@ return {
       "CocWatch",
     },
     event = { "User AstroFile", "InsertEnter" },
+    config = function()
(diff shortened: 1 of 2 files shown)


mehalter:
This is not a good approach because of this notice: https://github.com/neoclide/coc.nvim/blob/master/doc/coc.txt#L1843-L1845

> Note: those configuration would overwrite the configuration from the user's settings file, unless you have to use some dynamic variables, using the settings file is recommended.

This is not a good approach for configuration for a shared configuration that users can override.

Also the interaction with `neodev` is also incorrect. Part of the advantage of `neodev` is that it doesn't mess up the lua language server for non-neovim lua projects which this setup would do as well. It would be nice to add good integration in the language packs for coc but as it stands currently and the way coc.nvim is configured it is highly not recommended to do it this way especially in an environment where people will want to change values.

wroyca:
> This is not a good approach because of this notice: https://github.com/neoclide/coc.nvim/blob/master/doc/coc.txt#L1843-L1845
> 
> > Note: those configuration would overwrite the configuration from the user's settings file, unless you have to use some dynamic variables, using the settings file is recommended.
> 
> This is not a good approach for configuration for a shared configuration that users can override.
> 
> Also the interaction with `neodev` is also incorrect. Part of the advantage of `neodev` is that it doesn't mess up the lua language server for non-neovim lua projects which this setup would do as well. It would be nice to add good integration in the language packs for coc but as it stands currently and the way coc.nvim is configured it is highly not recommended to do it this way especially in an environment where people will want to change values.

Make sense, unfortunate that user configuration doesn't override, thank for the review!

PR #948, opened 2024-05-07
fix(coc-nvim): disable heirline coc_status update

wroyca:


mehalter (approved):
Hopefully someone can fix this, luckily very few people use this so seems fine

PR #956, opened 2024-05-10
feat(jj): Add vim-jjdescription to jj pack

Cretezy:
## 📑 Description

Add [vim-jjdescription](https://github.com/avm99963/vim-jjdescription) to jj pack.

image

### DIFF
--- a/lua/astrocommunity/pack/jj/init.lua
+++ b/lua/astrocommunity/pack/jj/init.lua
@@ -57,5 +57,5 @@ return {
       },
     },
   },
-  { "avm99963/vim-jjdescription" },
+  { "avm99963/vim-jjdescription", lazy = false },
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/pack/jj/init.lua
```diff
@@ -57,4 +57,5 @@ return {
       },
     },
   },
+  { "avm99963/vim-jjdescription" },
```
Uzaaft, line 60: we should enforce lazy = true, or lazy loading on a filetype, just as a sanity check against wonky compatibility with lazy loading

```suggestion
  { 
  "avm99963/vim-jjdescription",
  lazy = false
   },
```

mehalter (approved):

PR #967, opened 2024-05-16
feat(completion): add `copilot-cmp`

corintho:
## 📑 Description

Adds a new way to integrate copilot with cmp, besides the current options available. This one puts copilot within cmp, instead of setting it up in parallel with it.

### DIFF
--- a/lua/astrocommunity/completion/copilot-cmp/README.md
+++ b/lua/astrocommunity/completion/copilot-cmp/README.md
@@ -1,0 +1,10 @@
+# copilot.lua within CMP
+
+Fully featured & enhanced replacement for copilot.vim complete with API for interacting with Github Copilot
+
+**Repositories:** 
+<https://github.com/zbirenbaum/copilot.lua>
+<https://github.com/zbirenbaum/copilot-cmp>
+
+_Note_: This plugin will also reconfigure `cmp` and add `copilot` as a completion source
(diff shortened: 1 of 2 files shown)


Uzaaft (changes requested):
Fix the naming of the folder to follow the convention we use. Marking this one as a draft until you fix this.

corintho:
Hi Uzaaft, what do you mean by the naming convention for the folder? I am looking at it and I do not see how it could be any different than the one used, which also sits along with other copilot configurations

corintho:
Thinking about it further, I think you want me to use `copilot-lua-cmp` as a plugin folder name. But that name is already taken by another configuration with `copilot` and `cmp`. Although that one focus on having `copilot` **outside** of `cmp`, but both of them working together.
Mine focus on having `copilot` show **within** `cmp` and a regular completion option

Uzaaft:
This should be copilot-cmp.

Uzaaft (changes requested):
please follow naming convention of the folders.

luxus (approved):
i did pretty much the same Integration just some weeks ago and disliked it, it feels very slow :( but it works

corintho:
I like it because it is less intrusive, but now I got curious to see how other options perform compared to this one

Uzaaft (approved):

PR #972, opened 2024-05-17
fix(toggleterm-manager-nvim):  fix terminal selection in insert mode

refLog2112:


lua/astrocommunity/terminal-integration/toggleterm-manager-nvim/init.lua
```diff
@@ -32,6 +32,9 @@ return {
           ["d"] = { action = actions.delete_term, exit_on_action = false }, -- deletes a terminal buffer
           ["n"] = { action = actions.create_term, exit_on_action = false }, -- creates a new terminal buffer
         },
+        i = {
+          ["<CR>"] = { action = actions.toggle_term, exit_on_action = true }, -- toggles terminal open/closed
```
Uzaaft, line 36: Why isn't exit_on_action false here?

PR #975, opened 2024-05-19
fix(java): use builtin nvim 0.10 "go to definition/declaration" function

Juniar-Rakhman:
## 📑 Description

Using Neovim 0.10 and java-pack, the default keybinding for go to definition "gd" and go to declaration "gD" is no longer working.

The following error message is shown:
```
[telescope.builtin.lsp_*]: server does not support definitionProvider
```

To fix this the default "gd" and "gD" need to be mapped to the following command:
```
vim.lsp.buf.definition()
vim.lsp.buf.declaration()
```

List of changes:
- [x] added config to override default telescope "gd" and "gD" key binding.

### DIFF
--- a/lua/astrocommunity/pack/java/init.lua
+++ b/lua/astrocommunity/pack/java/init.lua
@@ -1,5 +1,25 @@
 return {
   {
+    "AstroNvim/astrolsp",
+    ---@type AstroLSPOpts
+    opts = {
+      mappings = {
+        n = {
+          gD = {
+            function() vim.lsp.buf.declaration() end,
+            desc = "Declaration of current symbol",
+            cond = "textDocument/declaration",
+          },
+          gd = {
+            function() vim.lsp.buf.definition() end,
+            desc = "Show definition of current symbol",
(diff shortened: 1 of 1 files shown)


Uzaaft:
are these functions only available on neovim 0.10?

Juniar-Rakhman:
@Uzaaft 

I thought it was new to 0.10 but it was also available in 0.9.5. I never had to look for it in 0.9.5 as it was working just fine.

After some more googling, I think this is more of a Telescope issue:
https://github.com/nvim-telescope/telescope.nvim/issues/2898#issuecomment-2117962082

I'm fine with this changes not included in the pack. I can modify my own keymaps until Telescope release a fix. 

I'm not sure if this is affecting other LSP's though. If it is then maybe we should change default "gd" and "gD" behavior in the core?

mehalter:
Telescope has released a new version with this fix

Juniar-Rakhman:
Sorry, but this issue is still happening in telescope v0.1.7.

PR #976, opened 2024-05-19
fix(catppuccin): set name to "catppuccin"

amqndin:
## 📑 Description
i tried installing hop as astrocommunity package, and in the repo i saw that it had code that supposedly should automatically enable integration for hop in catppuccin theme. yet that didnt work for some reason, and so i just enabled it manually. and today after thinking about it and 10 minutes of experimenting i figured it out. to fix this, all you have to do is set `name = "catppuccin"`, then its going to work. i just took hop as an example, but its with every single astrocommunity package i tried. like mini-clue or trouble and so on.

- [x] fixed for every plugin that had a catppuccin integration.

### DIFF
--- a/lua/astrocommunity/bars-and-lines/dropbar-nvim/init.lua
+++ b/lua/astrocommunity/bars-and-lines/dropbar-nvim/init.lua
@@ -7,6 +7,7 @@ return {
   },
   {
     "catppuccin/nvim",
+    name = "catppuccin",
     optional = true,
     ---@type CatppuccinOptions
     opts = { integrations = { dropbar = { enabled = true } } },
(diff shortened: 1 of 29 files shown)


Uzaaft:
@mehalter Does this make sense to add...?

lua/astrocommunity/bars-and-lines/dropbar-nvim/init.lua
```diff
@@ -7,6 +7,7 @@ return {
   },
   {
     "catppuccin/nvim",
+    name = "catppuccin",
```
mehalter, line 10:
```suggestion
     "catppuccin",
```

I think maybe in this case we should just replace `catppuccin/nvim` with the shorthanded `catppuccin` rather than giving the repo and name each time for it to be a bit cleaner. We do try to avoid the shorthand notation but in the case of a custom name it could make more sense

amqndin: hey, thanks for looking into this. i have done the change you suggested!

luxus (approved):
lgtm

PR #977, opened 2024-05-19
fix(rust): preserve loading of rust-analyzer.json

Ryex:
## 📑 Description

merge astrolsp settings in such a way as to preserve the default behavior of loading local project settings from rust-analyzer.json

## ℹ Additional Information

The default behavior of rustaceanvim is to load a `rust-analyzer.json` at the project root to load rust-anlyzer settings
see:
[rustaceanvim/config/server.lua#L15](https://github.com/mrcjkb/rustaceanvim/blob/b5d6ac50b3f87a6e8fae27cb8d8e49a6ca653aa6/lua/rustaceanvim/config/server.lua#L15)
[rustaceanvim/config/internal.lua#L287](https://github.com/mrcjkb/rustaceanvim/blob/b5d6ac50b3f87a6e8fae27cb8d8e49a6ca653aa6/lua/rustaceanvim/config/internal.lua#L287)

this file can have any of the keys defined in https://rust-analyzer.github.io/manual.html#configuration
either at the top level of the json or under a `'rust-analyzer'` top level key

This is done by having `server.settings` be a function that takes the project root path and the default settings configured at `server.default_settings` and calling rustaceanvim's `rustaceanvim.config.server.load_rust_analyzer_settings(root, opts)` function.

This change configures rustaceanvim by merging the astrolsp configured opts *around* a similar function that itself merges the astrolsp configured settings table with the defaults before loading the `rust-analyzer.json`

image

as you can see, it's working as expected

### DIFF
--- a/lua/astrocommunity/pack/rust/init.lua
+++ b/lua/astrocommunity/pack/rust/init.lua
@@ -86,7 +86,23 @@ return {
       end
 
       local astrolsp_avail, astrolsp = pcall(require, "astrolsp")
-      return { server = astrolsp_avail and astrolsp.lsp_opts "rust_analyzer", dap = { adapter = adapter } }
+      local astrolsp_opts = (astrolsp_avail and astrolsp.lsp_opts "rust_analyzer") or {}
+      local server = {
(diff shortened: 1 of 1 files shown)


mehalter (approved):
Looks great!

PR #985, opened 2024-05-24
fix(noice-nvim)!: disabled the signature to fix the overwritten lsp signature

Richie-Z:
## 📑 Description
when first opening neovim v0.10.0 with noice plugin installed it will throws error like this 
`vim.lsp.handlers["textDocument/signatureHelp"] has been overwritten by another plugin?`

add ```lsp = { signature = { enabled = false } }``` to disabled the overwritten lsp signature


Uzaaft:
That's not the case with just the noice plugin. It must be a another plugin that causes that issue. Tested in base template +  astrocommunity with noice.

PR #986, opened 2024-05-24
feat(jj): Add lazyjj

Cretezy:
## 📑 Description

- Adds [lazyjj](https://github.com/Cretezy/lazyjj) (a TUI for jj, similar to lazygit) mapping

Disclamer: I am the author of lazyjj.

image


Uzaaft (changes requested):
You should add a requirement about lazyjj being in path.
Or actually, we should make it an optional dependency. Similar to how lazygit is today.

Cretezy:
@Uzaaft: How should that work? Should it not include the key mapping if `lazyjj` is not present, or should it show an error when it's executed?

For lazygit, AstroNvim doesn't add the mapping if it's not in the path. I'm inclined to do the same.

Uzaaft:
That was my point, I should have written it better. Lets go for the same approach as how astronvim does it today with lazygit.

Uzaaft (approved):

PR #989, opened 2024-05-24
docs(project-nvim): add note about built-in functionality

rocode:
## 📑 Description

Adds a brief note to the project-nvim README.md file that AstroNvim contains a built-in Project Rooter, and if the community plugin is enabled, the built-in functionality will be disabled.

## ℹ Additional Information

This PR originated from a discord conversation in the #astrocommunity channel.

> @rocode: Should community plugins have a note on them if their functionality was already upstreamed or incorporated into AstroCore? For example, I just realized I didn't need project-nvim after seeing the Project Rooter recipe in the docs, but trying to migrate over caused confusion because it isn't documented that it turns off the built-in functionality.
>[...]
>@mehalter: yeah that would be great! If you open a PR it would be good to note on that plugin (and maybe the other project plugins) that there is a new rooter in AstroNvim that is on demand by default and opt in for automatic root changing

### DIFF
--- a/lua/astrocommunity/project/project-nvim/README.md
+++ b/lua/astrocommunity/project/project-nvim/README.md
@@ -3,3 +3,5 @@
 The superior project management solution for neovim.
 
 **Repository:** <https://github.com/jay-babu/project.nvim>
+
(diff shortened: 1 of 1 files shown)


mehalter (approved):
Awesome thanks!

PR #997, opened 2024-05-27
feat(cpp): add support for linux/arm64

### LINKED ISSUE #994: C/C++ Language Pack on linux/arm64

Using the C/C++ Language Pack on a linux arm64 system fails.  This is because Mason fails to download an arm64 clangd binary from github.  

### Describe the new feature

Clangd is supported on linux arm64. It can be installed from a distro package manager or installed from source.  In this case clangd should jsut be added to the servers section and omitted from the ensure_installed in the C/C++ Language Pack init script

lsp/servers.lua
```
return {
  "clangd"
 }
```

### Additional context

_No response_

joshheyse:
## 📑 Description

Conditionally ensures clangd installed or uses opts.servers for arm64 system

## ℹ Additional Information

I'm not familiar with lua / astronvim coding best practices, I'm sure the control flow could be better and interested in learning how to make it more idiomatic.

### DIFF
--- a/lua/astrocommunity/pack/cpp/init.lua
+++ b/lua/astrocommunity/pack/cpp/init.lua
@@ -1,18 +1,24 @@
+local machine = vim.loop.os_uname().machine
+local is_arm = machine == "aarch64" or machine.starts "arm"
+
 return {
   {
     "AstroNvim/astrolsp",
     optional = true,
-    ---@type AstroLSPOpts
-    opts = {
+    opts = function(_, opts)
       ---@diagnostic disable: missing-fields
-      config = {
+      opts = vim.tbl_deep_extend("keep", opts, {
         clangd = {
           capabilities = {
             offsetEncoding = "utf-8",
           },
         },
-      },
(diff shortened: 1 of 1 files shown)


mehalter:
There were some bugs, so I cleaned this up and committed it manually, thanks!

Uzaaft:
Heads up. This possibly breaks stuff for Mac users.

joshheyse:
Right. Good call.

We need a check for OS too, it should be is_linux_arm.

Sent from my iPhone

On May 27, 2024, at 23:51, Uzair Aftab ***@***.***> wrote:

﻿

Heads up. This possibly breaks stuff for Mac users.

mehalter:
should be fixed in latest commit to AstroCommunity

joshheyse:
@Uzaaft thanks for the catch! 

@mehalter thanks for the immediate action!

Sorry for submitting such a glaring issue. I even use macOS for a lot of my development in astronvim. 🤦

PR #998, opened 2024-05-28
feat(colorscheme): add hybrid.nvim

0xricksanchez:


lua/astrocommunity/colorscheme/hybrid-nvim/init.lua
```diff
@@ -0,0 +1,3 @@
+return {
+  "HoNamDuong/hybrid.nvim",
```
mehalter, line 2:
```suggestion
  "HoNamDuong/hybrid.nvim",
  opts = {},
```

this should be called

mehalter (changes requested):

mehalter (approved):
LGTM

PR #1006, opened 2024-05-29
feat(typescript): biome pack

### LINKED ISSUE #962: [Typescript pack] support biome for linter and formatter

_No response_

### Describe the new feature

Enable biome lsp for lspconfig and null-ls when project has @biomejs/biome dependency or has biome.json

### Additional context

_No response_

willparsons:
Closes #962 

## TODO
- [ ] Disable prettierd only if biome is running and should be the formatter
- [ ] Support null-ls and conform

## 📑 Description

- Added `biome` pack: `pack.typescript-biome`.
- Pulled out shared config into `pack.typescript.core`.
- Updated README

## ℹ Additional Information

I pulled out the `core` of the typescript pack so that the eslint+prettier and biome configuration can be separated easily. The default for `pack.typescript` hasn't changed.


lua/astrocommunity/pack/typescript/biome.lua
```diff
@@ -0,0 +1,32 @@
...
+    opts = function(_, opts)
+      opts.ensure_installed = require("astrocore").list_insert_unique(opts.ensure_installed, { "biome" })
+    end,
+  },
+  {
+    "jay-babu/mason-null-ls.nvim",
+    opts = function(_, opts)
+      if not opts.handlers then opts.handlers = {} end
+
+      if not has_biome() then return end
```
willparsons, line 25: Not really happy about this. I tried to use `get_clients` to just check if biome is attached, but in this case it didn't seem feasible because the lsp is attaching after this config is running.

lua/astrocommunity/pack/typescript/biome.lua
```diff
@@ -0,0 +1,32 @@
...
+
+      -- TODO: prettier should be off for all filetypes that biome supports,
+      -- currently this is { "javascript", "javascriptreact", "typescript", "typescriptreact", "json" }
+      opts.handlers.prettierd = function() end
```
willparsons, line 29: I don't know much about configuring null-ls, how would you do this?

The reason for wanting to do this is that prettierd will attach to `json` files even though biome is running and should be the primary formatter, but for markdown it should still be prettier.

willparsons:
just found this for picking the formatter dynamically: https://github.com/LazyVim/LazyVim/issues/2116#issuecomment-1971805431

will look into it

Uzaaft:
I think at this point it should just be typescript-biome and not like this.

Uzaaft:
I.e pack/typescript-biome

willparsons:
updated the structure 👍🏻

lua/astrocommunity/pack/typescript/core.lua
Uzaaft, line 1: This can be reverted

lua/astrocommunity/pack/typescript/init.lua
Uzaaft, line 1: This can be reverted

lua/astrocommunity/pack/typescript/core.lua
willparsons: aren't we going to end up having to copy the config around though?

willparsons:
Closing this as I don't have time to it anymore.

PR #1007, opened 2024-05-29
feat(vue): configure volar as tsserver plugin

gabrielscaranello:
## 📑 Description
Add Volar as a tsserver plugin, configured with vtsls.

## ℹ Additional Information

In  [`volar >= 2.0`](https://github.com/vuejs/language-tools/tree/master/packages/typescript-plugin), some synchronization problems between tsserver and volar appear, causing errors when working with .ts and .vue files at the same time.

An issue for this is being tracked (yioneko/vtsls#148)

### DIFF
--- a/lua/astrocommunity/pack/vue/init.lua
+++ b/lua/astrocommunity/pack/vue/init.lua
@@ -1,4 +1,5 @@
 return {
+  { import = "astrocommunity.pack.typescript" },
   {
     "AstroNvim/astrolsp",
     optional = true,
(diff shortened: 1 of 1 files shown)


Uzaaft (commented):
This seems alright on a first glance. What do you think our in-house LSP expert? @mehalter

mehalter (commented):
Have you tested this to make sure it works? Do we also need to reenable hybrid mode for volar?

Idk anything about this process but the code looks fine. I saw we just merge it if the user says they have actually tested it.

gabrielscaranello:
> Have you tested this to make sure it works? Do we also need to reenable hybrid mode for volar?
> 
> Idk anything about this process but the code looks fine. I saw we just merge it if the user says they have actually tested it.

I tested it on three projects and it worked as expected.  Enable hybrid mode works at first glance, but causes duplications in cmp entries and strange behavior in TS projects that don't use Vue.

mehalter:
Ok cool! Then looks good to me! :shipit:

mehalter (approved):

PR #1009, opened 2024-05-29
fix(bufferline-nvim): add bufferline cycle bindings

acarl005:
Map `[b` and `]b` to the bufferline-specific calls for cycling buffers.

## 📑 Description

I included `astrocommunity.bars-and-lines.bufferline-nvim` and observed the `]b` and `[b` bindings not behaving correctly. They were switching buffers, but out-of-order w.r.t. how they were displayed in bufferline. This PR remaps those to call bufferline's specific API for cycling through tabs. With this custom binding added, I now observe the correct behavior.

## ℹ Additional Information

It's not super straightforward to repro. Sometimes, I open a project, open various buffers, and press `]b` and the ordering is correct. Other times, after opening some buffers, say with `gd` or from telescope searches, the buffer navigation starts happening out of order.

### DIFF
--- a/lua/astrocommunity/bars-and-lines/bufferline-nvim/init.lua
+++ b/lua/astrocommunity/bars-and-lines/bufferline-nvim/init.lua
@@ -1,6 +1,19 @@
 return {
   {
     "akinsho/bufferline.nvim",
+    dependencies = {
+      {
+        "AstroNvim/astrocore",
+        opts = {
+          mappings = {
+            n = {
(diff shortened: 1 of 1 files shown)


Uzaaft (changes requested):
You should add a readme note about the keybinding since you're overriding the default

mehalter (approved):
:shipit:

mehalter:
I built on this a bit more here: https://github.com/AstroNvim/astrocommunity/commit/b31ee20529e912a071eae15b2cddd4f31f999c34

acarl005:
Thanks y'all :)

PR #1012, opened 2024-05-30
feat(git): Add `mini.git`

Uzaaft:


Uzaaft:
I don't think opts is needed for this plugin.

mehalter:
it's not necessary but there are `opts` and a setup call, so might as well put it 🤷🏼

lua/astrocommunity/git/mini-git/init.lua
```diff
@@ -0,0 +1,8 @@
+---@type LazySpec
+return {
+  "echasnovski/mini-git",
+  event = "User AstroFile",
```
mehalter, line 4:
```suggestion
  event = "User AstroGitFile",
```

only load automatically for git files

lua/astrocommunity/git/mini-git/init.lua
```diff
@@ -0,0 +1,8 @@
...
+  cmd = { "Git" },
+  version = false,
+  main = "mini.git",
```
mehalter, line 7:
```suggestion
```

this is automatic by lazy i'm pretty sure

lua/astrocommunity/git/mini-git/init.lua
```diff
@@ -0,0 +1,8 @@
+---@type LazySpec
+return {
+  "echasnovski/mini-git",
+  event = "User AstroFile",
+  cmd = { "Git" },
+  version = false,
```
mehalter, line 6:
```suggestion
```

we don't need to force this on the user

lua/astrocommunity/git/mini-git/init.lua
```diff
@@ -0,0 +1,8 @@
...
+  main = "mini.git",
+}
```
mehalter, line 8:
```suggestion
  opts = {},
}
```

might as well put this since there is a `setup` function

mehalter (changes requested):

lua/astrocommunity/git/mini-git/init.lua
```diff
@@ -0,0 +1,8 @@
+---@type LazySpec
+return {
+  "echasnovski/mini-git",
+  event = "User AstroFile",
+  cmd = { "Git" },
```
mehalter, line 5:
```suggestion
  cmd = "Git",
```

unnecessary table

mehalter (approved):
:shipit:

PR #1013, opened 2024-05-30
fix(trouble-nvim): change commands in keymaps to v3 replacements

FraI3mega:
## 📑 Description
Changes the commands in keymaps to the v3 replacements.


mehalter:
This actually needs a bit more, I have this included in #1014  Thanks for the help!

PR #1018, opened 2024-05-31
feat(cs): add `csharpls-extended-lsp.nvim` for better go to definition

### LINKED ISSUE #951: Support csharpls-extended-lsp.nvim

I would like go to definition to work properly with C#, needs https://github.com/Decodetalkers/csharpls-extended-lsp.nvim.

Doing `:lua require('csharpls_extended').lsp_definitions()` on a symbol makes it go to the decompiled version.

### Describe the new feature

Go to definition will work for non-project code in C#.

### Additional context

I have tried setting it up to override the `textDocument/definition` lsp_handler like so:

```lua
{
  "Decodetalkers/csharpls-extended-lsp.nvim",
  lazy = false,
  dependencies = {
    "AstroNvim/astrolsp",
    opts = {
      config = {
        csharp_ls = {
          handlers = {
            ["textDocument/definition"] = function(...)
              require("csharpls_extended").handler(...)
            end,
            ["textDocument/typeDefinition"] = function(...)
              require("csharpls_extended").handler(...)
            end,
          },
        },
      },
    },
  },
},
```

However this isn't working. I've also tried this with no success:

```lua
opts = function(_, opts)
  if not opts.lsp_handlers then
    opts.lsp_handlers = {}
  end
  local definition_event = "textDocument/definition"
  local definition_event_original = opts.lsp_handlers[definition_event] or vim.lsp.handlers[definition_event]
  opts.lsp_handlers[definition_event] = function(err, result, ctx, config)
    print("Definition handler")
    local client = vim.lsp.get_client_by_id(ctx.client_id)
    if client and vim.tbl_contains({
          "csharp_ls",
        }, client.name) then
      require("csharpls_extended").handler(err, result, ctx, config)
    end
    definition_event_original(err, result, ctx, config)
  end

  local type_definition_event = "textDocument/typeDefinition"
  local type_definition_event_original = opts.lsp_handlers[type_definition_event]
      or vim.lsp.handlers[type_definition_event]
  opts.lsp_handlers[type_definition_event] = function(err, result, ctx, config)
    print("TypeDefinition handler")
    local client = vim.lsp.get_client_by_id(ctx.client_id)
    if client and vim.tbl_contains({
          "csharp_ls",
        }, client.name) then
      require("csharpls_extended").handler(err, result, ctx, config)
    end
    type_definition_event_original(err, result, ctx, config)
  end
end,
```

mehalter:
Closes #951 

## 📑 Description

This is untested because I don't have a C# environment. @Cretezy this is the correct way to set it up, it's the same way you mentioned you tried it. If it doesn't work it could just be the plugin is broken and we shouldn't add it. Could you give this a try?

### DIFF
--- a/lua/astrocommunity/pack/cs/README.md
+++ b/lua/astrocommunity/pack/cs/README.md
@@ -6,6 +6,7 @@ This plugin pack does the following:
 
 - Adds `c_sharp` Treesitter parsers
 - Adds `csharp_ls` language server
+- Adds [`csharpls-extended-lsp.nvim`](https://github.com/Decodetalkers/csharpls-extended-lsp.nvim) for better go to definition/type definition
 - Adds the following `null-ls` sources:
   - [csharpier](https://github.com/belav/csharpier)
 - Adds `coreclr` debug adapters for nvim-dap
--- a/lua/astrocommunity/pack/cs/init.lua
+++ b/lua/astrocommunity/pack/cs/init.lua
@@ -24,6 +24,24 @@ return {
     end,
   },
   {
+    "Decodetalkers/csharpls-extended-lsp.nvim",
+    dependencies = {
+      {
+        "AstroNvim/astrolsp",
+        opts = {
+          config = {
+            csharp_ls = {
+              handlers = {
+                ["textDocument/definition"] = function(...) require("csharpls_extended").handler(...) end,
+                ["textDocument/typeDefinition"] = function(...) require("csharpls_extended").handler(...) end,
+              },
+            },
+          },
+        },
+      },
+    },
+  },
+  {
     "jay-babu/mason-nvim-dap.nvim",
     optional = true,
     opts = function(_, opts)


Uzaaft:
testing this since I have a working dotnet env on my pc

Cretezy:
I'm not going to be doing C# until next Tuesday, I can test and review then

Uzaaft:
Didnt work for me, the LSP in the pack is broken for Mac🥹

ahmtsen:
I think the problem is caused by telescope as mentioned here https://github.com/Decodetalkers/csharpls-extended-lsp.nvim/issues/7. Maybe overriding gd to use vim.lsp.buf.definition() instead of telescope methods will solve the problem.

ahmtsen:
> I think the problem is caused by telescope as mentioned here [Decodetalkers/csharpls-extended-lsp.nvim#7](https://github.com/Decodetalkers/csharpls-extended-lsp.nvim/issues/7). Maybe overriding gd to use vim.lsp.buf.definition() instead of telescope methods will solve the problem.

The following configuration works for me. The fix is to override `gd` to use `vim.lsp.buf.definition`.

```lua
  {
    "Decodetalkers/csharpls-extended-lsp.nvim",
    dependencies = {
      {
        "AstroNvim/astrolsp",
        opts = {
          config = {
            csharp_ls = {
              handlers = {
                ["textDocument/definition"] = function(...) require("csharpls_extended").handler(...) end,
                ["textDocument/typeDefinition"] = function(...) require("csharpls_extended").handler(...) end,
              },
            },
          },
          on_attach = function(client, bufnr)
            if client.name ~= "csharp_ls" then return end

            vim.keymap.set(
              "n",
              "gd",
              vim.lsp.buf.definition,
              { buffer = bufnr, noremap = true, silent = true, desc = "Go to definition" }
            )
          end,
        },
      },
    },
  },

```

mehalter:
Thanks for pointing this out @ahmtsen ! I am going to be removing the telescope LSP mappings in the next patch release of AstroNvim because apparently there are many problems with them. Also thanks for the workaround! Once the next release is made I'll go ahead and merge this in.


Also sidenote, you know you can also add `on_attach` functions to the config of language servers themselves to add settings for specific language servers. Rather than checking the `client.name` you can do something like this:

```lua
{
  "Decodetalkers/csharpls-extended-lsp.nvim",
  dependencies = {
    {
      "AstroNvim/astrolsp",
      opts = {
        config = {
          csharp_ls = {
            handlers = {
              ["textDocument/definition"] = function(...) require("csharpls_extended").handler(...) end,
              ["textDocument/typeDefinition"] = function(...) require("csharpls_extended").handler(...) end,
            },
            on_attach = function(_, bufnr)
              vim.keymap.set("n", "gd", vim.lsp.buf.definition, { buffer = bufnr, desc = "Go to definition" })
            end,
          },
        },
      },
    },
  },
},
```

mehalter:
This should work as intended now that AstroNvim v4.10.2 is released that doesn't use Telescope for LSP mappings out of the box.

PR #1035, opened 2024-06-13
Auto session restore

Kamilcuk:
## 📑 Description

Add Auto session restore with checking also for stdin.


Kamilcuk:
Och I can't recreate (?)

PR #1036, opened 2024-06-13
feat(recipes): add auto-session-restore

Kamilcuk:
## 📑 Description

Adds auto-session-restore like in documentation, but also detect if nvim has been started with stdin redirected, like `stream | nvim` or like `export MAPAGER='nvim +Man!'; man man`.

The stdin detection is taken from https://github.com/thaerkh/vim-workspace/blob/master/plugin/workspace.vim#L268 .

## ℹ Additional Information

Would be nice to add the stdin input handling also to documentation.

### DIFF
--- a/lua/astrocommunity/recipes/auto-session-restore/init.lua
+++ b/lua/astrocommunity/recipes/auto-session-restore/init.lua
@@ -1,0 +1,33 @@
+local read_from_stdin = false
+return {
+  {
+    "AstroNvim/astrocore",
+    ---@type AstroCoreOpts
+    opts = {
+      autocmds = {
+        -- disable alpha autostart
+        alpha_autostart = false,
+        restore_session = {
+          {
+            event = "StdinReadPost",
+            desc = "Check if we are reading from stdin",
+            nested = true,
+            callback = function() read_from_stdin = true end,
+          },
+          {
+            event = "VimEnter",
+            desc = "Restore previous directory session if neovim opened with no arguments",
+            nested = true, -- trigger other autocommands as buffers open
+            callback = function()
+              -- Only load the session if nvim was started with no args
+              if not read_from_stdin and vim.fn.argc(-1) == 0 then
+                -- try to load a directory session using the current working directory
+                require("resession").load(vim.fn.getcwd(), { dir = "dirsession", silence_errors = true })
+              end
+            end,
+          },
+        },
+      },
+    },
+  },
+}
--- a/lua/astrocommunity/recipes/auto-session-restore/init.lua
+++ b/lua/astrocommunity/recipes/auto-session-restore/init.lua
@@ -1,4 +1,3 @@
-local read_from_stdin = false
 return {
   {
     "AstroNvim/astrocore",
@@ -9,20 +8,37 @@ return {
         alpha_autostart = false,
         restore_session = {
           {
-            event = "StdinReadPost",
-            desc = "Check if we are reading from stdin",
-            nested = true,
-            callback = function() read_from_stdin = true end,
-          },
-          {
             event = "VimEnter",
             desc = "Restore previous directory session if neovim opened with no arguments",
             nested = true, -- trigger other autocommands as buffers open
             callback = function()
-              -- Only load the session if nvim was started with no args
-              if not read_from_stdin and vim.fn.argc(-1) == 0 then
-                -- try to load a directory session using the current working directory
-                require("resession").load(vim.fn.getcwd(), { dir = "dirsession", silence_errors = true })
+              -- Logic copied from https://github.com/AstroNvim/AstroNvim/blob/365aa6e083dcd25fa3d1c8a2515d7e71a03d51d3/lua/astronvim/plugins/alpha.lua#L49
+              local should_skip
+              local lines = vim.api.nvim_buf_get_lines(0, 0, 2, false)
+              if
+                vim.fn.argc() > 0 -- don't start when opening a file
+                or #lines > 1 -- don't open if current buffer has more than 1 line
+                or (#lines == 1 and lines[1]:len() > 0) -- don't open the current buffer if it has anything on the first line
+                or #vim.tbl_filter(function(bufnr) return vim.bo[bufnr].buflisted end, vim.api.nvim_list_bufs()) > 1 -- don't open if any listed buffers
+                or not vim.o.modifiable -- don't open if not modifiable
+              then
+                should_skip = true
+              else
+                for _, arg in pairs(vim.v.argv) do
+                  if arg == "-b" or arg == "-c" or vim.startswith(arg, "+") or arg == "-S" then
+                    should_skip = true
+                    break
+                  end
+                end
+              end
+              if should_skip then return end
+              -- if possible, load session
+              if not pcall(function() require("resession").load(vim.fn.getcwd(), { dir = "dirsession" }) end) then
+                -- if session was not loaded, if possible, load alpha
+                require("lazy").load { plugins = { "alpha-nvim" } }
+                if pcall(function() require("alpha").start(true) end) then
+                  vim.schedule(function() vim.cmd.doautocmd "FileType" end)
+                end
               end
             end,
           },
--- a/lua/astrocommunity/recipes/auto-session-restore/README.md
+++ b/lua/astrocommunity/recipes/auto-session-restore/README.md
@@ -1,0 +1,6 @@
+# AstroLSP - Automatically Restore Previous Session
+
+**Website:** <https://docs.astronvim.com/recipes/sessions/#automatically-restore-previous-session>
+
+This plugin specification configures AstroLSP to automatically
+restore their previous session for a given directory when opening Neovim with no arguments.


Uzaaft:
Do you have the link to our documentation?

Uzaaft:
Link to docs: https://docs.astronvim.com/recipes/sessions/

lua/astrocommunity/recipes/auto-session-restore/init.lua
```diff
@@ -0,0 +1,35 @@
+local read_from_stdin = false
```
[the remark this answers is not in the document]
Kamilcuk: Hi, I mentioned it in the description.

"detect if nvim has been started with stdin redirected, like stream | nvim or like export MAPAGER='nvim +Man!'; man man."

I also mention I have taken the method to detect from vim-workspace plugin.

The issue is that if the snippet from documentation is used and `MANPAGER='nvim +Man!'` is set, it is _not possible_ to open a man page. Instead of opening a man page, the session is loaded.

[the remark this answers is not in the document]
Uzaaft: Then it's not sure if its something that should be in the recipes, those are for the examples from the docs. @mehalter has the final word on this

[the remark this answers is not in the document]
mehalter: this variable makes sense, I do think there is potentially a more performant method. But also this should totally be in the recipes!

lua/astrocommunity/recipes/auto-session-restore/init.lua
```diff
@@ -0,0 +1,33 @@
...
+                -- try to load a directory session using the current working directory
+                require("resession").load(vim.fn.getcwd(), { dir = "dirsession", silence_errors = true })
+              end
+            end,
+          },
+        },
+      },
+    },
+  },
+}
```
mehalter, line 33:
```suggestion
return {
  "AstroNvim/astrocore",
  ---@type AstroCoreOpts
  opts = {
    autocmds = {
      -- disable alpha autostart
      alpha_autostart = false,
      restore_session = {
        {
          event = { "VimEnter", "StdinReadPost" },
          desc = "Restore previous directory session if neovim opened with no arguments",
          once = true, -- delete itself after executing once
          nested = true, -- trigger other autocommands as buffers open
          callback = function(args)
            -- Only load the session if nvim was started with no args
            if args.event == "VimEnter" and vim.fn.argc(-1) == 0 then
              -- try to load a directory session using the current working directory
              require("resession").load(vim.fn.getcwd(), { dir = "dirsession", silence_errors = true })
            end
          end,
        },
      },
    },
  },
}
```

I think this approach may be a bit faster and also avoids the necessity for a separate variable and multiple autocommands. Rather than having a variable we have the autocommand trigger on both events. If the `StdinReadPost` happens then it simply deletes itself, if not then the `VimEnter` happens and it restores the session and then deletes itself.

mehalter (changes requested):
A small suggestion to improve the performance. Thanks so much for opening this up, it's a great addition! Also really like the improvement for the `StdinReadPost`

lua/astrocommunity/recipes/auto-session-restore/init.lua
mehalter: Let me know what you think

Kamilcuk: Sorry, it does not work. It is not removed on `once=true`. I see it triggers twice - once for StdinReadPost and then for VimEnter. So the `once=true` does not protect it.

mehalter: Ok, I have a better way then. Standby

Kamilcuk: Yea. So I think every event is registered separately, so `autocmd ++once` is evaluated for every event separately. I think it is the loop here https://github.com/neovim/neovim/blob/ba70404c558169b813f51d5a0bb71cd540555c93/src/nvim/api/auto cmd.c#L488 .

lua/astrocommunity/recipes/auto-session-restore/init.lua
```diff
@@ -0,0 +1,25 @@
...
+          desc = "Restore previous directory session if neovim opened with no arguments",
+          once = true, -- delete itself after executing once
+          nested = true, -- trigger other autocommands as buffers open
+          callback = function(args)
+            -- Only load the session if nvim was started with no args
+            if args.event == "VimEnter" and vim.fn.argc(-1) == 0 then
+              -- try to load a directory session using the current working directory
+              require("resession").load(vim.fn.getcwd(), { dir = "dirsession", silence_errors = true })
+            end
+          end,
```
mehalter, line 20:
```suggestion
          nested = true, -- trigger other autocommands as buffers open
          callback = function(args)
            -- Only load the session if nvim was started with no args
            if args.event == "VimEnter" and vim.fn.argc(-1) == 0 then
              -- try to load a directory session using the current working directory
              require("resession").load(vim.fn.getcwd(), { dir = "dirsession", silence_errors = true })
            end
            return true -- delete itself after executing a single time
          end,
```

Autocommand callbacks can return `true` to indicate they should delete themselves.

mehalter: woah github diff algorithm so bad :joy:

mehalter (changes requested):

lua/astrocommunity/recipes/auto-session-restore/init.lua
mehalter: let's see if this works maybe, if not then we can always add in a `vim.api` call to just delete the `augroup`

mehalter:
@Kamilcuk here are 2 options:

1. return `true`, this might work? hopefully but it might be the same as `once = true`:

```lua
return {
  "AstroNvim/astrocore",
  ---@type AstroCoreOpts
  opts = {
    autocmds = {
      -- disable alpha autostart
      alpha_autostart = false,
      restore_session = {
        {
          event = { "VimEnter", "StdinReadPost" },
          desc = "Restore previous directory session if neovim opened with no arguments",
          nested = true, -- trigger other autocommands as buffers open
          callback = function(args)
            -- Only load the session if nvim was started with no args
            if args.event == "VimEnter" and vim.fn.argc(-1) == 0 then
              -- try to load a directory session using the current working directory
              require("resession").load(vim.fn.getcwd(), { dir = "dirsession", silence_errors = true })
            end
            return true
          end,
        },
      },
    },
  },
}
```

2. This will definitely work

```lua
return {
  "AstroNvim/astrocore",
  ---@type AstroCoreOpts
  opts = {
    autocmds = {
      -- disable alpha autostart
      alpha_autostart = false,
      restore_session = {
        {
          event = { "VimEnter", "StdinReadPost" },
          desc = "Restore previous directory session if neovim opened with no arguments",
          nested = true, -- trigger other autocommands as buffers open
          callback = function(args)
            -- Only load the session if nvim was started with no args
            if args.event == "VimEnter" and vim.fn.argc(-1) == 0 then
              -- try to load a directory session using the current working directory
              require("resession").load(vim.fn.getcwd(), { dir = "dirsession", silence_errors = true })
            end
            vim.api.nvim_del_augroup_by_name "restore_session"
          end,
        },
      },
    },
  },
}
```

Kamilcuk:
hi:

1. I am testing by doing `callback = function(args) print(args.event) .....` and then executing `man man`. I am getting both events:

```
:messages                                                                                                                                                                                                         
StdinReadPost                                                                                                                                                                                                     
VimEnter     
```

Both events are printed, so it does not work.

2. Yup, it works.

I would be happy with that, but now I have a better version:

```
  {
    "AstroNvim/astrocore",
    ---@type AstroCoreOpts
    opts = {
      autocmds = {
        -- disable alpha autostart
        alpha_autostart = false,
        restore_session = {
          {
            event = { "VimEnter" },
            desc = "Restore previous directory session if neovim opened with no arguments",
            nested = true, -- trigger other autocommands as buffers open
            callback = function(args)
              local should_skip
              local lines = vim.api.nvim_buf_get_lines(0, 0, 2, false)
              if
                vim.fn.argc() > 0 -- don't start when opening a file
                or #lines > 1 -- don't open if current buffer has more than 1 line
                or (#lines == 1 and lines[1]:len() > 0) -- don't open the current buffer if it has anything on the first line
                or #vim.tbl_filter(function(bufnr) return vim.bo[bufnr].buflisted end, vim.api.nvim_list_bufs()) > 1 -- don't open if any listed buffers
                or not vim.o.modifiable -- don't open if not modifiable
              then
                should_skip = true
              else
                for _, arg in pairs(vim.v.argv) do
                  if arg == "-b" or arg == "-c" or vim.startswith(arg, "+") or arg == "-S" then
                    should_skip = true
                    break
                  end
                end
              end
              if should_skip then return end
              if
                vim.tbl_contains(require("resession").list { dir = "dirsession" }, (vim.fn.getcwd():gsub("/", "_")))
              then
                require("resession").load(vim.fn.getcwd(), { dir = "dirsession" })
              else
                require("lazy").load { plugins = { "alpha-nvim" } }
                require("alpha").start(true)
                vim.schedule(function() vim.cmd.doautocmd "FileType" end)
              end
            end,
          },
        },
      },
    },
  },
```

I don't like the idea that alpha will not start never. So I copied code from https://github.com/AstroNvim/AstroNvim/blob/365aa6e083dcd25fa3d1c8a2515d7e71a03d51d3/lua/astronvim/plugins/alpha.lua#L44 . If there is a session, use it, otherwise show alpha. This actually works and shows alpha on directories without session, but loads session with directories with session. And also `echo hi | nvim` and `man man` both work, because I think they catch `not vim.o.modifiable`.

mehalter:
Nice! I like this one! I think the only thing is adding protection against if alpha isn't installed

mehalter (approved):
Looks great! Thanks so much!

PR #1039, opened 2024-06-14
fix(go): Remove iferr package from jay-babu/mason-null-ls.nvim and WhoIsSethDaniel/mason-tool-installer.nvim lists

lucazz:
## 📑 Description

The iferr package install currently fails with the following error msg: 
```sh
$ go install github.com/koron/iferr@latest
go: github.com/koron/iferr@latest (in github.com/koron/iferr@v0.0.0-20240122035601-9c3e2fbe4bd1): go.mod:3: invalid go version '1.21.5': must match format 1.23
$ 
```
Until the package maintainer fixes it, we should remove this package from the list to avoid a bad user experience.

### DIFF
--- a/lua/astrocommunity/pack/go/init.lua
+++ b/lua/astrocommunity/pack/go/init.lua
@@ -72,7 +72,7 @@ return {
     opts = function(_, opts)
       opts.ensure_installed = require("astrocore").list_insert_unique(
         opts.ensure_installed,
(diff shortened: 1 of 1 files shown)


Uzaaft:
We should pin it instead.

Uzaaft:
I can compile it just fine. Is your go version up to date?

Uzaaft:
I can reproduce with a go version older than 1.21. User error.

PR #1056, opened 2024-06-20
feat(completion): add `nvim-cmp-buffer-lines`

Axlefublr:
## 📑 Description

Adds the `nvim-cmp-buffer-lines` nvim-cmp source, but instead of just adding it to normal completions, keeps it local to a hotkey.

The reasoning for this is that, even with a low priority, whole line completions end up being super annoying, when mixed in with all the other (usually more useful) suggestions. So it's for this reason that this source is effectively "opt-in".

The mapping used is `<C-x><C-l>`, because that's the default way to do whole line completions. What you basically get with this community pack, is you replace built in whole line completions with ones that use nvim-cmp.

The hotkey is wack, but it is default! So I think it makes the most sense, even if there could be other, more ergonomic mappings.

At the very least, if a person wants the functionality, but on a different mapping, they now have the option to take a peek at this community pack, copy paste it into their config, and change the mapping to what they want.

### DIFF
--- a/lua/astrocommunity/completion/nvim-cmp-buffer-lines/README.md
+++ b/lua/astrocommunity/completion/nvim-cmp-buffer-lines/README.md
@@ -1,0 +1,5 @@
+# nvim-cmp-buffer-lines
+
+_Replaces_ the `<C-x><C-l>` whole line complete, to use nvim-cmp instead, using the following source:
+
+**Repository:** <https://github.com/amarakon/nvim-cmp-buffer-lines>
--- a/lua/astrocommunity/completion/nvim-cmp-buffer-lines/README.md
+++ b/lua/astrocommunity/completion/nvim-cmp-buffer-lines/README.md
@@ -3,3 +3,5 @@
 _Replaces_ the `<C-x><C-l>` whole line complete, to use nvim-cmp instead, using the following source:
 
 **Repository:** <https://github.com/amarakon/nvim-cmp-buffer-lines>
+
+- _Replaces_ the `<C-x><C-l>` whole line complete, to use nvim-cmp instead.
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/completion/nvim-cmp-buffer-lines/README.md
```diff
@@ -0,0 +1,5 @@
+# nvim-cmp-buffer-lines
+
+_Replaces_ the `<C-x><C-l>` whole line complete, to use nvim-cmp instead, using the following source:
```
Uzaaft, line 3: The standard is to use the "one-liner" from the github repo

```suggestion
nvim-cmp source for buffer words
```

lua/astrocommunity/completion/nvim-cmp-buffer-lines/README.md
```diff
@@ -0,0 +1,5 @@
...
+
+**Repository:** <https://github.com/amarakon/nvim-cmp-buffer-lines>
```
Uzaaft, line 5:
```suggestion
**Repository:** <https://github.com/amarakon/nvim-cmp-buffer-lines>

- _Replaces_ the `<C-x><C-l>` whole line complete, to use nvim-cmp instead.
```

Uzaaft:
Curious as to if you want to keep the note about <C-x><C-l> or not. Besides that, this looks good!

Axlefublr:
You wanted me to remove it, no?

Uzaaft:
> You wanted me to remove it, no?

I moved it to under the repository link 😅

Uzaaft:
Nice. I was wondering why you weren't in the discord anymore. Seeing this PR from you made my day @Axlefublr

Uzaaft (approved):

PR #1059, opened 2024-06-20
feat(go)!: move from `go.nvim` to `gopher.nvim`

mehalter:
## 📑 Description

This moves away from `go.nvim` as it is very error prone and also very heavy. This also uses Mason for go dependencies by default rather than the synchronous and slow dependency installation command.


rami3l (approved):
LGTM! This has covered some missing points in #1060 so I'll probably close it in favor of this PR. The README should definitely be reverted as well though.

GregDuhamel (approved):

Uzaaft (dismissed):
Amazing shit

lua/astrocommunity/pack/go/README.md
Uzaaft, line 1: gofumpt has been removed, and gotests has been added as a source

Uzaaft:
The readme is outdated since we've removed gofumpt and added gotests

lua/astrocommunity/pack/go/README.md
mehalter: Fixed

PR #1071, opened 2024-06-24
fix(vue):  properly configure hybrid mode

chaozwn:
Now this latest version should be, tsserver supports vue, then enable hybridMode, volar provides the template prompt.


Uzaaft:
Please fix the description so that it makes more sense, and the PR title to follow the rules.

mehalter (approved):

PR #1077, opened 2024-07-02
fix(codeium-vim): port mappings to `astrocore` and fix lazy loading event

manuuurino:

### DIFF
--- a/lua/astrocommunity/completion/codeium-vim/init.lua
+++ b/lua/astrocommunity/completion/codeium-vim/init.lua
@@ -16,13 +16,7 @@ return {
       mappings = {
         n = {
           ["<Leader>;"] = {
-            function()
-              if vim.g.codeium_enabled == true then
-                vim.cmd "CodeiumDisable"
-              else
-                vim.cmd "CodeiumEnable"
-              end
-            end,
+            "<Cmd>CodeiumToggle<CR>",
             noremap = true,
             desc = "Toggle Codeium active",
           },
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/completion/codeium-vim/init.lua
```diff
@@ -1,17 +1,45 @@
 return {
   "Exafunction/codeium.vim",
-  event = "User AstroFile",
```
manuuurino, line 3: In my testing, ive noticed that this event no longer works? Cant explain why, havent done the research on it, kinda tired rn. @mehalter Is it a potential bug from AstroNvim? Could look sometime later if you arent intrested.

Uzaaft: Lazy loading with vim pugins is wonky. We should just explicitly disable it with
`lazy = false` instead.

manuuurino: Would rather go with the `BufEnter` event and it works, thats how they used it in their [examples](https://github.com/Exafunction/codeium.vim/blob/7a01ffb1e6538fe634f294908a50a7b680039284/README.md#-lazy), also ive implemented it already in this pr.

Uzaaft: Sure if you've tested it lets go for that.

Uzaaft (approved):

PR #1080, opened 2024-07-02
BREAKING CHANGE: remove scope.nvim

yofriadi:
## 📑 Description

Since AstroNvim manage it's own implementation of scoped tab buffer I think we can remove this plugin as it redundant and source of confuse.


Uzaaft:
If the user want to use the plugin they should be able to do so.

PR #1082, opened 2024-07-03
fix(codeium-nvim): resolve loading issues for nvim-cmp and lspkind.nvim

khas-amir:

### DIFF
--- a/lua/astrocommunity/completion/codeium-nvim/init.lua
+++ b/lua/astrocommunity/completion/codeium-nvim/init.lua
@@ -1,63 +1,61 @@
 return {
-  {
-    "Exafunction/codeium.nvim",
-    event = "User AstroFile",
-    cmd = "Codeium",
-    opts = {
-      enable_chat = true,
-    },
-    dependencies = {
-      {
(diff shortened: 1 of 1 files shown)


Uzaaft:
Please provide more info regarding the problem you're fixing in the PR.

lua/astrocommunity/completion/codeium-nvim/init.lua
```diff
@@ -36,28 +36,28 @@ return {
...
+    end,
+  },
+  {
+    "onsails/lspkind.nvim",
+    optional = true,
+    -- Adds icon for codeium using lspkind
+    opts = function(_, opts)
+      opts.symbol_map = {
+        Codeium = require("astroui").get_icon("Codeium", 1, true),
+      }
```
mehalter, line 60:
```suggestion
      if not opts.symbol_map then opts.symbol_map = {} end
      opts.symbol_map.Codeium = require("astroui").get_icon("Codeium", 1, true)
```

This is a bug and will overwrite `symbol_map` if it already exists

mehalter (approved):

Issue #1081, opened 2024-07-02
Blade for Laravel project is incomplete


MrMic:
### Checklist

- [X] I have searched through the AstroNvim documentation
- [X] I have searched through the existing issues of this project
- [X] I have searched the existing issues of plugins related to this issue
- [X] I can replicate the bug with the minimal `repro.lua` provided below

### Neovim version (nvim -v)

0.10.0

### Operating system/version

Linux Ubuntu 22.04

### Terminal/GUI

kitty

### Describe the bug

Where can I find:
- laravel-dev-generators

Cannot find anywhere on the web!!!

### Steps to Reproduce

Installation through standard package: Blade

### Expected behavior

Blade lsp and others stuff

### Minimal configuration

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "runtime", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  -- stylua: ignore
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", "--branch=stable", lazypath })
end
vim.opt.rtp:prepend(vim.env.LAZY or lazypath)

-- install plugins
local plugins = {
  { "AstroNvim/AstroNvim", import = "astronvim.plugins" },
  { "AstroNvim/astrocommunity" },

  -- add any other plugins/customizations here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

-- add anything else here (autocommands, vim.filetype, etc.)
```

Uzaaft:
https://github.com/haringsrob/laravel-dev-tools Its in the releases here. 

@MrMic  You could have solved this yourself by putting in some effort. This is not a bug, more like a skill issue.

MrMic:
There is only a: 
**laravel-dev-tools**  
and NOT a:
**laravel-dev-generators**

Uzaaft:
https://github.com/haringsrob/laravel-dev-tools/issues/9

Uzaaft:
You can also install other generators. :)

PR #1083, opened 2024-07-03
docs(blade): Add inn link for laravel-dev-generators

Uzaaft:
Avoids cases like #1081  where people are too lazy to search themselves.


luxus (approved):
easy pr scoring

PR #1090, opened 2024-07-07
fix(nvim-lsp-endhints): Fix issue with loading of plugin due to event

Uzaaft:
We tried AstroLsp, but seems that the event loading proposed by the plugin maintainer is the way to go.


mehalter (approved):
Yeah this is the correct event

PR #1094, opened 2024-07-08
fix(python): better basedpyright defaults

Uzaaft:


mehalter (approved):

awerty-noob:
the default doesn't work on my astroNvim. When I open a python source code, a bunch of errors `Type annotation is missing` comes up, I checked Mason and the LSP server configuration schema shows that `basedpyright.analysis.typeCheckingMode = "all"`

Uzaaft:
> the default doesn't work on my astroNvim. When I open a python source code, a bunch of errors `Type annotation is missing` comes up, I checked Mason and the LSP server configuration schema shows that `basedpyright.analysis.typeCheckingMode = "all"`

You can change the settings if you dont like the defaults.

PR #1095, opened 2024-07-08
feat(java)!: move to `nvim-java` for `jdtls` setup

### LINKED ISSUE #1062: Replace nvim-jdtls with nvim-java

_No response_

### Describe the new feature

Which should replace nvim-jdtls with https://github.com/nvim-java/nvim-java, which provides more functionality like spring-boot support etc 

### Additional context

_No response_

### LINKED ISSUE #1064: Add a spring-boot pack

_No response_

### Describe the new feature

Right now, the current java pack doesn't support spring-boot. 
We should add this functionality in a seperate pack. 
https://github.com/JavaHello/spring-boot.nvim/tree/main is a candidate plugin. 

### Additional context

_No response_

mehalter:
Closes #1062
Closes #1064 

## 📑 Description

This moves the Java pack to use [`nvim-java`](https://github.com/nvim-java/nvim-java) which greatly decreases the complexity of the pack as well as adds stuff like spring-boot support out of the box.

I did some testing and it spins up totally fine, the only downside I have found so far is that the debugger in my testing doesn't seem to get configured correctly. A debugging adapter is created but no DAP configurations are inserted. There seems to be an upstream issue regarding this: https://github.com/nvim-java/nvim-java/issues/169

Someone might want to run this down or verify if it's a me problem before we merge this in.

## TODO

- [ ] Confirm debugger and make sure that works to not introduce a big regression

### DIFF
--- a/lua/astrocommunity/pack/java/README.md
+++ b/lua/astrocommunity/pack/java/README.md
@@ -7,17 +7,8 @@ Requires:
 This plugin pack does the following:
 
 - Adds `java` treesitter parsers
-- Adds `html` treesitter parsers (XML)
 - Adds `jdtls` language server
-- Adds `lemminx` language server (XML)
-- Adds [nvim-jdtls](https://github.com/mfussenegger/nvim-jdtls) for language specific tooling
-- Enables [hot reloading](https://github.com/mfussenegger/nvim-jdtls/issues/80)
-- Adds `javadbg` and `javatest` debug adapters for nvim-dap
-
-## Note
-
-We require that the root folder of your projects include one of these files/folders([Reference](https://github.com/AstroNvim/astrocommunity/blob/49e9a3961bba079d7f413b8d5567382dd6f55392/lua/astrocommunity/pack/java/java.lua#LL37C96-L37C96)):  
-`.git`, `mvnw`, `gradlew`, `pom.xml`, `build.gradle` or `.project`
+- Adds [nvim-java](https://github.com/nvim-java/nvim-java) for language specific tooling
 
 ## Tips
 
@@ -25,30 +16,29 @@ We require that the root folder of your projects include one of these files/fold
 Here is a simple example:
 
 ```lua
+return {
   "AstroNvim/astrocommunity",
   { import = "astrocommunity.pack.java" },
   {
-    "mfussenegger/nvim-jdtls",
+    "AstroNvim/astrolsp",
     opts = {
-      settings = {
-        java = {
-          configuration = {
-            runtimes = {
-              {
-                name = "JavaSE-11",
-                path = "/usr/lib/jvm/java-11-openjdk/",
+      config = {
+        jdtls = {
+          settings = {
+            java = {
+              configuration = {
+                runtimes = {
+                  {
+                    name = "JavaSE-11",
+                    path = "/usr/lib/jvm/java-11-openjdk/",
+                  },
+                },
               },
             },
           },
         },
-        format = {
-          enabled = true,
-          settings = { -- you can use your preferred format style
-            url = "https://raw.githubusercontent.com/google/styleguide/gh-pages/eclipse-java-google-style.xml",
-            profile = "GoogleStyle",
-          },
-        },
       },
     },
-  }
+  },
+}
 ```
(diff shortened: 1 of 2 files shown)


Juniar-Rakhman:
I work with the java pack on a daily basis. Can I test this out first?

mehalter:
Please do

Juniar-Rakhman:
Tried your changes, lombok isn't working (using java 21):
image

Give me sometimes to test nvim-java out to find the optimal configs for both Spring and non-Spring project (Quarkus)

Juniar-Rakhman:
I was able to solve the lombok issue by doing full reset on nvim and also deleting my ~/.m2 folder. I couldn't get debugging to work because of this upstream issue: https://github.com/nvim-java/nvim-java/issues/125. I tried disabling spring-boot-tools via config, but it does not work:
```lua
return {
  { import = "astrocommunity.pack.java" },
  {
    "AstroNvim/astrolsp",
    opts = {
      config = {
        jdtls = {
          settings = {
            java = {
              configuration = {
                java_debug_adapter = {
                  enable = true,
                },
                spring_boot_tools = {
                  enable = false,
                },
                jdk = {
                  auto_install = false,
                },
                runtimes = {
                  {
                    name = "JavaSE-21",
                    path = os.getenv "HOME" .. "/.sdkman/candidates/java/21.0.1-tem",
                  },
                },
              },
            },
          },
        },
      },
    },
  },
}
```

Juniar-Rakhman:
I don't think nvim-java is mature enough yet to replace nvim-jdtls. 
@mehalter How about putting this into a separate pack, i.e. pack/spring-boot? That way, we can keep the already working java pack and experiment with nvim-java using spring-boot pack?

mehalter:
@Juniar-Rakhman the options for disabling spring boot are not part of jdtls. They should be passed to the `opts` for `nvim-java`

mehalter:
I think if we do that, we might as well just make a plugin for `spring-boot.nvim` and just use it with the current Java pack. No need to use `nvim-java` at all. I also don't care at all. I don't use Java nor care what is in this pack. Feel free to open a PR with whatever you want.

I do think configuring spring boot directly rather than using `nvim-java` makes the most sense

mehalter:
https://github.com/JavaHello/spring-boot.nvim

I don't even know if it needs to be a pack, it can just be a plugin in the `lsp` section or whatever makes the most sense. Provides a pack of things

Juniar-Rakhman:
> @Juniar-Rakhman the options for disabling spring boot are not part of jdtls. They should be passed to the `opts` for `nvim-java`

thanks! the following config works for me:
```lua
return {
  { import = "astrocommunity.pack.java" },
  {
    "nvim-java/nvim-java",
    opts = {
      jdk = {
        auto_install = false,
      },
      spring_boot_tools = {
        enable = false,
      },
    },
  },
  {
    "AstroNvim/astrolsp",
    opts = {
      config = {
        jdtls = {
          settings = {
            java = {
              configuration = {
                runtimes = {
                  {
                    name = "JavaSE-21",
                    path = os.getenv "HOME" .. "/.sdkman/candidates/java/21.0.1-tem",
                  },
                },
              },
            },
          },
        },
      },
    },
  },
}

```
Debugging works on test files at least, further testing is needed. You need to turn off the `spring-boot-tools` to get around the [bug](https://github.com/nvim-java/nvim-java/issues/125).

Juniar-Rakhman:
@mehalter I did some further testing. Debugging and testing works, but not as intuitive as nvim-jdtls. Nvim-jdtls functionalities that I often use, such as generating tests, go to tests, attaching debugger to a running process are still missing. Debugging experience is not very nice, when running test in debug mode, the dap-ui does not open automatically.

So in conclusion I would rather stick with nvim-jdtls, maybe we could upgrade the current java pack and integrate nvim-jdtls with [spring-boot-tool](https://github.com/JavaHello/spring-boot.nvim).

I know that nvim-java is much easier to config, but I think the current java pack is good enough in hiding the complexity of nvim-jdtls. I don't think moving the pack to nvim-java is worth the missing functionalities and the regression in user experience.

Uzaaft:
Thanks for taking the time to test this PR @Juniar-Rakhman. This is exactly what the community near more off! 🫶🏽

PR #1105, opened 2024-07-13
fix(trouble-nvim): prevent clashes with existing `<Leader>x` mappings

rami3l:
## 📑 Description

This is to avoid clashes with the [recent addition of `<Leader>x` to AstroNvim](https://github.com/AstroNvim/AstroNvim/commit/99c2b13df7b4943b3f82ca772db3dfb8f13f42e7).

This PR has also simplified some mapping descriptions.

### DIFF
--- a/lua/astrocommunity/diagnostics/trouble-nvim/init.lua
+++ b/lua/astrocommunity/diagnostics/trouble-nvim/init.lua
@@ -8,21 +8,15 @@ return {
       opts = function(_, opts)
         local maps = opts.mappings
         local prefix = "<Leader>x"
-        maps.n[prefix] = { desc = require("astroui").get_icon("Trouble", 1, true) .. "Trouble" }
-        maps.n[prefix .. "X"] = { "<Cmd>Trouble diagnostics toggle<CR>", desc = "Workspace Diagnostics (Trouble)" }
+        maps.n[prefix .. "X"] = { "<Cmd>Trouble diagnostics toggle<CR>", desc = "Trouble Workspace Diagnostics" }
         maps.n[prefix .. "x"] =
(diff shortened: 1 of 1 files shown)


mehalter:
We can leave all the mappings as is, just change `<Leader>xq` to `<Leader>xQ` and `<Leader>xl` to `<Leader>xL`

mehalter:
No need to make them all 3 key presses when they can share the Quickfix/Lists menu

rami3l:
> We can leave all the mappings as is, just change `<Leader>xq` to `<Leader>xQ` and `<Leader>xl` to `<Leader>xL`

@mehalter Thanks for your suggestion! I'll modify my PR accordingly.

rami3l:
> We can leave all the mappings as is, just change `<Leader>xq` to `<Leader>xQ` and `<Leader>xl` to `<Leader>xL`

@mehalter What if other types of Quickfix/Lists show up later though? Looks like we are prioritizing `trouble` somehow, although that would be beneficial for my own usecase...

mehalter:
Yeah I think if the user adds several list plugins it can just be up to them the priority they want

rami3l:
> Yeah I think if the user adds several list plugins it can just be up to them the priority they want

@mehalter Okay, I've applied your suggestions. One problem I have is that the original descriptions are too long, so I've tried my best to shorten them a bit.

mehalter (changes requested):
We should leave the original messages no? That mention Trouble?

mehalter:
@rami3l this should be a 2 character change PR

mehalter:
Ah I see your most recent comment now

mehalter (approved):
Sure! Looks good

rami3l:
> Sure! Looks good

@mehalter Thanks a lot, and sorry for the previous confusion! 🙇

PR #1107, opened 2024-07-13
fix(main): Swap out usage of `name = *` in mapping declaration with `desc = *` to conform with which-key v3

Uzaaft:


mehalter (approved):
This vimtex one will definitely still not work at all. I can take a closer look on Monday thi

PR #1110, opened 2024-07-15
feat(lsp): add `nvim-java` plugin

mehalter:
## 📑 Description

Supersedes #1095 . This just adds `nvim-java` as an individual plugin in the LSP category. Seems like a good way to provide it in AstroCommunity for those who want to try it but it's opt-in.


Uzaaft (approved):
:shipit:

PR #1118, opened 2024-07-23
feat(colorscheme): Add vim-moonfly-colors

Std-Enigma:


Uzaaft:
@Std-Enigma  Does the plugin work with `lazy = true`. If not, we should explicitly set it to false

ALameLlama:
LGTM

Uzaaft (approved):

PR #1121, opened 2024-07-23
fix(codesnap-nvim): Add default saving path

### LINKED ISSUE #1120: [codesnap-nvim] doesn't save the pictures to the drive
### Checklist

- [X] I have searched through the AstroNvim documentation
- [X] I have searched through the existing issues of this project
- [X] I have searched the existing issues of plugins related to this issue
- [X] I can replicate the bug with the minimal `repro.lua` provided below

### Neovim version (nvim -v)

0.10.0 release

### Operating system/version

ArchLinux

### Terminal/GUI

kitty

### Describe the bug

Since the save_path options isn't specified in the plugin the taken pictures don't get saved.

### Steps to Reproduce

1. Select a part of your code
2. run the CodeSnapSave command

### Expected behavior

CodeSnap should the save the picture to your system's hard drive

### Screenshots

_No response_

### Additional Context

this is the error you get:
If you want to save snapshot in somewhere, you should config the save_path before, refer: https://github.com/mistricky/codesnap.nvim?tab=readme-ov-file#save-the-snapshot

### Minimal configuration

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
end

local lazypath = vim.fn.stdpath "data" .. "/lazy/lazy.nvim"
if not (vim.uv or vim.loop).fs_stat(lazypath) then
  -- stylua: ignore
end
vim.opt.rtp:prepend(lazypath)

require("lazy").setup {
  { "AstroNvim/astrocommunity", { import = "astrocommunity.media.codesnap-nvim" } },
}
```

Std-Enigma:
Closes #1120

### DIFF
--- a/lua/astrocommunity/media/codesnap-nvim/init.lua
+++ b/lua/astrocommunity/media/codesnap-nvim/init.lua
@@ -20,6 +20,7 @@ return {
     },
   },
   opts = {
+    save_path = "~/Pictures",
     mac_window_bar = false,
     watermark = "",
   },
--- a/lua/astrocommunity/media/codesnap-nvim/README.md
+++ b/lua/astrocommunity/media/codesnap-nvim/README.md
@@ -4,7 +4,9 @@
 
 **Repository:** <https://github.com/mistricky/codesnap.nvim>
 
-**Notes:** This plugin spec defaults the `save_path` to `~/Pictures`.
+This plugin spec defaults the `save_path` to `~/Pictures`.
+
+
 
 > [!NOTE]
 >
--- a/lua/astrocommunity/media/codesnap-nvim/README.md
+++ b/lua/astrocommunity/media/codesnap-nvim/README.md
@@ -3,3 +3,5 @@
 📸 Snapshot plugin with rich features that can make pretty code snapshots for Neovim
 
 **Repository:** <https://github.com/mistricky/codesnap.nvim>
+
+**Notes:** This plugin spec defaults the `save_path` to `~/Pictures`.


lua/astrocommunity/media/codesnap-nvim/init.lua
```diff
@@ -20,6 +20,7 @@ return {
     },
   },
   opts = {
+    save_path = "~/Pictures",
```
Uzaaft, line 23: This should rather use `os.getenv( "HOME" )`

Std-Enigma: can i know why? doesn't tilda rely on the HOME env internally?
or you just mean we should scrap the Pictures folder part?

Uzaaft: Not sure if `~` get resolved correctly on windows.

Std-Enigma: oh for that actually this plugin doesn't support windows at all unless you use a very early stage version
so no worries in that regard

**But I'm working on it to use the correct version based on the system you have in a different branch i will make a pr when i tested it**

Uzaaft:
Hold off on merging this, I’m gonna try and upstream something instead

Uzaaft:
https://github.com/mistricky/codesnap.nvim/pull/124 This will make this redundant

lua/astrocommunity/media/codesnap-nvim/init.lua
mistricky: Hi guys @Std-Enigma @Uzaaft , I just noticed this thread. 

CodeSnap has already parsed `~` and `$HOME` to the `$HOME` environment value in the lower level, thus in Lua level, it can pass `~` or `$HOME` directly, these values should be parse correctly by `generator` module of CodeSnap.nvim.

image

Uzaaft: The discussion wasn't only limited to this plugin, but also for astrocommunity as a whole.

Uzaaft:
Re-opened since the maintainer is slow to merge, which is fair enough. Let's get this merged and then we'll revert the changes whenever the upstream PR is merged.

Uzaaft (approved):

PR #1122, opened 2024-07-23
fix(quarto): Lazy load on command

Uzaaft:


Uzaaft:
Wait, should actually scratch this. Something is wrong with the lazy loading it seems.

PR #1123, opened 2024-07-23
feat(codesnap-nvim): support for windows users

Std-Enigma:
## ℹ Additional Information

This plugin at least on the latest version doesn't support windows by any means but the "0.0.x" version does however but it's not so performant and works kinda differently 

This PR make sures that windows users get the latest working version that supports their OS

### DIFF
--- a/lua/astrocommunity/media/codesnap-nvim/README.md
+++ b/lua/astrocommunity/media/codesnap-nvim/README.md
@@ -4,8 +4,8 @@
 
 **Repository:** <https://github.com/mistricky/codesnap.nvim>
 
-> [!NOTE] **Windows Support:**
+> [!NOTE]
 >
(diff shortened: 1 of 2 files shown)


ALameLlama:
This looks good to me but i think @Uzaaft was looking into an upstream change

Std-Enigma:
i mean sure but at least for now we can hack our way to get it to work for every user until they support it in the later versions

Std-Enigma:
> upstream change 
so you mean i should I've just pushed this changes to the same pr i made a while ago?

Uzaaft:
> > upstream change
> > so you mean i should I've just pushed this changes to the same pr i made a while ago?

What?

Uzaaft:
@Std-Enigma can you add a note into the readme to document this behaviour

ALameLlama (approved):

ALameLlama:
Lgtm

PR #1128, opened 2024-07-24
fix(pack): fix stevearc/conform.nvim ansible setup

lucazz:
## 📑 Description

This PR updates stevearc/conform.nvim's setup, addressing the deprecation warning that comes up when opening any yaml file while using the yaml pack:
```
deprecated[conform]: The nested {} syntax to run the first formatter has been replaced by the stop_after_first option (see :help conform.format).
Support for the old syntax will be dropped on 2025-01-01.
```


Uzaaft:
We should fix this across all the different usages at once.

PR #1132, opened 2024-07-26
feat(split-and-window): add colorful-winsep.nvim plugin

Std-Enigma:
## 📑 Description

This plugin adds beautiful colors to the borders of your split windows. Additionally, it includes a smooth and pleasant animation when switching between them.

## ℹ Additional Information

This PR also adds colorful-winsep.nvim integration to the catppuccin colorscheme
_Note_: For some reason, the integration doesn't work when specified as an optional spec, so I had to include it in the Catppuccin plugin specification.

### DIFF
--- a/lua/astrocommunity/split-and-window/colorful-winsep-nvim/init.lua
+++ b/lua/astrocommunity/split-and-window/colorful-winsep-nvim/init.lua
@@ -1,0 +1,5 @@
+return {
+  "nvim-zh/colorful-winsep.nvim",
+  event = "WinLeave",
+  config = true,
+}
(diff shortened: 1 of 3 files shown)


lua/astrocommunity/split-and-window/colorful-winsep-nvim/init.lua
```diff
@@ -0,0 +1,5 @@
+return {
+  "nvim-zh/colorful-winsep.nvim",
+  event = "WinLeave",
+  config = true,
```
Uzaaft, line 4: Use opts instead.

Uzaaft (commented):
See comment.

Uzaaft (changes requested):
See comment.

Uzaaft (approved):

PR #1133, opened 2024-07-26
feat(rust): update `rustaceanvim` to v5

mehalter:
## 📑 Description

This updates to the new [Rustaceanvim v5 release](https://github.com/mrcjkb/rustaceanvim/releases/tag/5.0.0). This drops support for neovim v0.9. To add backwards compatibility I simply copied the entired of the spec. This protects against if the `opts` change between the two or something the previous version will always be correct.


ALameLlama (approved):
LGTM, but do we want to include a note in the readme saying it'll get pinned based on your nvim version?

PR #1135, opened 2024-07-27
feat(haskell): update `haskell-tools` to v4

mehalter:
## 📑 Description

This updates to the new [haskell-tools v4 release](https://github.com/mrcjkb/haskell-tools.nvim/releases/tag/4.0.0). This drops support for neovim v0.9. To add backwards compatibility I simply copied the entired of the spec. This protects against if the `opts` change between the two or something the previous version will always be correct.


ALameLlama (approved):
LGTM, just pinning versions for nvim versions

ALameLlama:
@mehalter would we want to add something to the readme saying the version will get pinned to v3 if you're on nvim 0.9?

PR #1140, opened 2024-08-01
feat(completion): add supermaven-nvim

taskylizard:
## 📑 Description

Adds configuration for https://github.com/supermaven-inc/supermaven-nvim.

### DIFF
--- a/lua/astrocommunity/completion/supermaven-nvim/init.lua
+++ b/lua/astrocommunity/completion/supermaven-nvim/init.lua
@@ -1,16 +1,14 @@
 return {
   "supermaven-inc/supermaven-nvim",
   event = "VeryLazy",
-  config = function()
-    require("supermaven-nvim").setup {
-      keymaps = {
-        accept_suggestion = "<C-l>",
-        clear_suggestion = "<C-h>",
-        accept_word = "<C-w>",
-      },
-      log_level = "warn",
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/completion/supermaven-nvim/init.lua
Uzaaft, line 1: Does the keymaps appear in which key?

Uzaaft: Ah true, they are ctrl based.

lua/astrocommunity/completion/supermaven-nvim/init.lua
```diff
@@ -0,0 +1,16 @@
...
+    require("supermaven-nvim").setup {
+      keymaps = {
+        accept_suggestion = "<C-l>",
+        clear_suggestion = "<C-h>",
+        accept_word = "<C-w>",
+      },
+      log_level = "warn",
+      disable_inline_completion = false, -- disables inline completion for use with cmp
+      disable_keymaps = false, -- disables built in keymaps for more manual control
+    }
```
Uzaaft, line 14: These can be moved to an opts instead.

taskylizard: Indeed yes, I just followed the documentation for this but this is better, fixed.

Uzaaft (approved):

PR #1146, opened 2024-08-08
feat(debugging): add `quicker.nvim`

mehalter:
## 📑 Description

This adds the new [`quicker.nvim`](https://github.com/stevearc/quicker.nvim) plugin by the legendary `stevearc`

This also makes a new category `quickfix` since `debugging` isn't super clear. It deprecates the old `nvim-bqf` location and warns the user to migrate their configuration so it's non-breaking atm.


ALameLlama (approved):
LGTM

PR #1147, opened 2024-08-08
feat(cs): add neotest integration with `neotest-dotnet`

ahmtsen:
## 📑 Description

adds `neotest-dotnet` package optionally if "astrocommunity.testing.neotest" pack and "astrocommunity.pack.cs" pack are enabled at the same time testing support for dotnet will be enabled


Uzaaft (approved):
:shipit:

PR #1149, opened 2024-08-09
fix(transparent-nvim): make `transparent.nvim` actually work with Heirline (also makes the `opts` table work as well)

mehalter:
## 📑 Description

This makes `transparent.nvim` actually work with Heirline now that features have been added that allow it to support it. This also fixes the fact that the `transparent.nvim` spec apparently wasn't using the `opts` table. So it's actually configurable by the user now.


Uzaaft (approved):
:shipit:

Uzaaft:
I fucking love that hamster

PR #1151, opened 2024-08-11
feat(pack): remove pinned version of `typst-preview.nvim` in `typst` pack

rami3l:
## 📑 Description

New releases of this plugin is quite frequent now (https://github.com/chomosuke/typst-preview.nvim/releases), so I don't think we should limit ourselves to `0.1.*`.

## ℹ Additional Information

N/A


ALameLlama (approved):
LGTM, I can't see anything on the plugin about minimum requirements or anything that will break compatibility.

Mostly just bumping typst-preview

PR #1153, opened 2024-08-12
feat(markdown): added markdown-preview.nvim plugin.

11D-Beyonder:
Use command `MarkdownPreview` to open the browser for real-time preview.

### DIFF
--- a/lua/astrocommunity/pack/markdown/README.md
+++ b/lua/astrocommunity/pack/markdown/README.md
@@ -5,3 +5,4 @@ This plugin pack does the following:
 - Adds `markdown` and `markdown_inline` Treesitter parsers
 - Adds `marksman` language server
 - Adds `prettierd` formatter
+- Adds `markdown-preview.nvim` for real-time preview
--- a/lua/astrocommunity/pack/markdown/init.lua
+++ b/lua/astrocommunity/pack/markdown/init.lua
@@ -1,5 +1,21 @@
 return {
   {
+    "iamcco/markdown-preview.nvim",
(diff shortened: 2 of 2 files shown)


11D-Beyonder:
> ### Review Checklist
> Does this PR follow the [Contribution Guidelines](development guidelines)? Following is a _partial_ checklist:
> 
> Proper conventional commit scoping:
> 
> * If you are adding a new plugin, the scope would be the name of the category it is being added into. ex. `feat(utility): added noice.nvim plugin`
> * If you are modifying a pre-existing plugin or pack, the scope would be the name of the plugin folder. ex. `fix(noice-nvim): fix LSP handler error`
> * [x]  Pull request title has the appropriate conventional commit type and **scope** where the scope is the name of the _pre-existing_ directory in the project as described above
> * [x]  `README` is properly formatted and uses fenced in links with `<url>` unless they are inside a `[title](url)`
> * [x]  Entry returns a single plugin spec with the _new plugin_ as the only top level spec (not applicable for recipes or packs).
> * [x]  Proper usage of `opts` table rather than setting things up with the `config` function.
> * [x]  Proper usage of `specs` table for all specs that are not dependencies of a given plugin (not applicable for recipes or packs).

ALameLlama:
Personally, I don't think this should be included in the pack

There are other plugins that will work for both SSH and desktop such as 
https://github.com/OXY2DEV/markview.nvim
https://github.com/ellisonleao/glow.nvim

but I am happy with this being a standalone plugin.

PR #1154, opened 2024-08-12
feat(markdown): added markdown-preview.nvim plugin.

11D-Beyonder:
## 📑 Description

Use command `MarkdownPreview` to open browser for real-time preview.


mehalter:
Thanks for the recommendation! This is actually already available in AstroCommunity: https://github.com/AstroNvim/astrocommunity/tree/main/lua/astrocommunity/markdown-and-latex/markdown-preview-nvim

There are many markdown preview plugins that tons of people have different opinions on. So it is left up to the user to pick one that they like.

PR #1155, opened 2024-08-12
feat(markdown-and-latex): add `markview.nvim`

mehalter:
## 📑 Description

Adds [`markview.nvim`](https://github.com/OXY2DEV/markview.nvim)

### DIFF
--- a/lua/astrocommunity/markdown-and-latex/markview-nvim/init.lua
+++ b/lua/astrocommunity/markdown-and-latex/markview-nvim/init.lua
@@ -1,0 +1,9 @@
+return {
+  "OXY2DEV/markview.nvim",
+  ft = { "markdown", "markdown.mdx" },
+  dependencies = { "nvim-treesitter/nvim-treesitter" },
+  opts = {
+    hybrid_modes = { "n" },
+    headings = { shift_width = 0 },
+  },
+}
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/markdown-and-latex/markview-nvim/init.lua
```diff
@@ -0,0 +1,9 @@
+return {
+  "OXY2DEV/markview.nvim",
+  ft = { "markdown", "markdown.mdx" },
+  dependencies = { "nvim-treesitter/nvim-treesitter" },
```
ALameLlama, line 4: on their docs, they also have `nvim-tree/nvim-web-devicons` as a dep, should we include this here too?

ALameLlama (changes requested):

lua/astrocommunity/markdown-and-latex/markview-nvim/init.lua
mehalter: It's fine without it because it will load on module

ALameLlama: sweet

ALameLlama (approved):
LGTM

PR #1160, opened 2024-08-13
fix: add `astrocore` dependency to `helpview-nvim` and `markview-nvim`

manuuurino:


mehalter:
This isn't necessary. AstroCore is a requirement of AstroNvim so it's already going to be present

PR #1162, opened 2024-08-16
fix(go): automatically install gofumpt and goimports

pgr0ss:
## 📑 Description

The go configuration tells conform to use these tools for formatting, but they were not automatically installed. This fix adds them to the mason-tool-installer list so they will be installed with the rest of the go tools.

## ℹ Additional Information

These tools are referenced here:
https://github.com/AstroNvim/astrocommunity/blob/bb8089b72bbb11ad850902ce58a9f1fc64277563/lua/astrocommunity/pack/go/init.lua#L134-L142

### DIFF
--- a/lua/astrocommunity/pack/go/init.lua
+++ b/lua/astrocommunity/pack/go/init.lua
@@ -87,7 +87,7 @@ return {
     opts = function(_, opts)
       opts.ensure_installed = require("astrocore").list_insert_unique(
         opts.ensure_installed,
-        { "delve", "gopls", "gomodifytags", "gotests", "iferr", "impl" }
+        { "delve", "gopls", "gofumpt", "goimports", "gomodifytags", "gotests", "iferr", "impl" }
       )
     end,
   },


Uzaaft (changes requested):
Could you add them to the docs as Well

Uzaaft:
actually, we should do the opposite

Uzaaft:
We should remove these from conform.

pgr0ss:
Why do you think we should remove them from conform instead of installing them?

Uzaaft:
> Why do you think we should remove them from conform instead of installing them?

Because we removed them from the null-ls portion. We should remove them from conform as Well.

pgr0ss:
Sorry, I don't think I understand. `goimports` in particular is extremely useful, and I don't think we want to remove it. I think we want auto imports by default. I have less of an opinion on`gofumpt`, however. I have been using it and like it, but I think it's also ok if the go pack defaults to using the standard formatter. Thanks.

Uzaaft:
> think

It's already removed in the pack. The conform stuff is just dangling because it wasn't removed with the null-ls stuff

mehalter:
Opened up a new PR that amends this commit. Thanks for catching this! `goimports` was removed on accident :D

Uzaaft:
> Opened up a new PR that amends this commit. Thanks for catching this! `goimports` was removed on accident :D

I wonder who removed them...👀

PR #1163, opened 2024-08-18
fix(grug-far-nvim): fix icon disabled options

dannyfritz:
## 📑 Description

Additional options that should be set for grug-far.nvim if icons are disabled.

## ℹ Additional Information

- https://github.com/MagicDuck/grug-far.nvim/blob/main/lua/grug-far/opts.lua#L150-L150
- https://github.com/MagicDuck/grug-far.nvim/blob/main/lua/grug-far/opts.lua#L156-L170

Before:
Screenshot from 2024-08-17 19-02-54
After:
Screenshot from 2024-08-17 19-01-47


Uzaaft (approved):
Grug it

PR #1164, opened 2024-08-18
fix(go): Remove conform table

Uzaaft:
This PR removes the conform table in the go back, since conform doesn't support these OOB. They were removed from the original go pack when migrating away from go.nvim.

If merged, it superseeds #1162


Uzaaft:
> Nah, the other PR is correct

So that means that the initial PR removing those were wrong then?

PR #1166, opened 2024-08-19
fix(go): install missing `goimports`

mehalter:
## 📑 Description

This supersedes #1162. `gopls` uses `gofumpt` for formatting so we can leverage that after running `goimports`

### DIFF
--- a/lua/astrocommunity/pack/go/README.md
+++ b/lua/astrocommunity/pack/go/README.md
@@ -7,6 +7,7 @@ This plugin pack does the following:
 - Adds `go` Treesitter parsers
 - Adds `gopls` language server
 - Adds the following go packages:
+  - [goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports?utm_source=godoc)
   - [gomodifytags](https://github.com/fatih/gomodifytags)
   - [gotests](https://github.com/cweill/gotests)
   - [iferr](https://github.com/koron/iferr)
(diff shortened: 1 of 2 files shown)


Uzaaft (approved):

Uzaaft:
Fixes mistake done in #1059 I think?

mehalter:
Yup!

pgr0ss:
Don't you still need to install `gofumpt` via `mason-tool-installer`? In my local testing, it didn't run `gofumpt` until I added this:
```lua
  {
    "WhoIsSethDaniel/mason-tool-installer.nvim",
    opts = function(_, opts)
      opts.ensure_installed = require("astrocore").list_insert_unique(opts.ensure_installed, { "gofumpt" })
    end,
  },
```

mehalter:
It doesn't need to, it uses `gopls` for formatting since that has `gofumpt` built in

mehalter:
I can take a look in a bit, but `gopls` _should_ be formatting after `goimports` 😅

mehalter:
Just tested @pgr0ss and it works great! no `gofumpt` necessary :D `gopls` handles the formatting itself

pgr0ss:
@mehalter Do we need to add `gofumpt = true` to the `gopls` config, then? This says the default is `false`:

https://github.com/golang/tools/blob/master/gopls/doc/settings.md#gofumpt-bool

When I added `gofumpt = true` here it started working for me:

https://github.com/AstroNvim/astrocommunity/blob/76f4f81b5da0f1573a2d7040d7909e11f1239fac/lua/astrocommunity/pack/go/init.lua#L11-L51

mehalter:
oh yeah maybe! I see, without it it's just running `go fmt`. Yup that would be a great addition. One second and I'll add it

PR #1169, opened 2024-08-20
feat(editing-support): add `CopilotChat.nvim`

ALameLlama:
## 📑 Description

Add [CopilotChat.nvim](https://github.com/CopilotC-Nvim/CopilotChat.nvim) with some keybinds

### DIFF
--- a/lua/astrocommunity/editing-support/copilotchat-nvim/README.md
+++ b/lua/astrocommunity/editing-support/copilotchat-nvim/README.md
@@ -4,4 +4,4 @@ Chat with GitHub Copilot in Neovim
 
 **Repository:** https://github.com/CopilotC-Nvim/CopilotChat.nvim
 
-Note: This will use the same auth as `:Copilot setup`
+Note: This will to have Copilot setup, this uses the same auth as `:Copilot setup`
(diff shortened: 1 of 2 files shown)


Uzaaft:
@ALameLlama  In what way is the lazy loading broken?

ALameLlama:
@Uzaaft The way i understand using cmd = {} will lazy load the plugin until one of the commands if used. currently it still getting loaded on startup and taking about 100ms for this plugin alone

Uzaaft:
Ref this from my lazy.nvim:

ALameLlama:
Ok, so the lazy loading is breaking when I require within astrocore, I'll need to move the requires within the functions instead of having them at the top. which honestly makes complete sense when I am getting it to load them right away lol instead of when the function is called for the keys

ALameLlama:
We can also style this a little bit, by default it creates a vertical split, but if you already have splits, then you get it in the middle and it's a little awkward.

By default it looks like this
image

but we can make it a floating window always on the right like this.
image

Uzaaft:
I think most of these makes sense. Some of the window stuff might me opinionated, but lets see what people say.

Uzaaft (approved):
:shipit:

PR #1171, opened 2024-08-21
fix(helm): better file type detection

Zebradil:
## 📑 Description

This PR improves filetype detection for helm files.

The current filetype detection often enables yaml-ls for helm templates. yaml-ls then marks helm specific syntax as errors. To avoid this, a few specific rules are added to detect helm files.

## ℹ Additional Information

The new code is mainly based on the following sources:

- https://github.com/mandos/nvim-helm/blob/7b5b5945bf9c84b17dceb2f9a6cabfac7a6d12f1/lua/helm/init.lua
- https://github.com/mandos/nvim-helm/blob/7b5b5945bf9c84b17dceb2f9a6cabfac7a6d12f1/lua/helm/utils.lua

They are under the MIT license. How should I attribute them?

- [x] Add attribution to the source code
- [x] Test the changes (I have tested them but as a separate module, outside of astrocommunity)

### DIFF
--- a/lua/astrocommunity/pack/helm/init.lua
+++ b/lua/astrocommunity/pack/helm/init.lua
@@ -1,53 +1,44 @@
-local M = {}
-
-function M.is_helm_file(path)
-  local check = vim.fs.find("Chart.yaml", { path = vim.fs.dirname(path), upward = true })
-  return not vim.tbl_isempty(check)
-end
-
---@private
---@return string
-function M.yaml_filetype(path, _) return M.is_helm_file(path) and "helm.yaml" or "yaml" end
-
---@private
---@return string
-function M.tmpl_filetype(path, _) return M.is_helm_file(path) and "helm.tmpl" or "template" end
-
---@private
---@return string
-function M.tpl_filetype(path, _) return M.is_helm_file(path) and "helm.tmpl" or "smarty" end
-
 return {
   {
     "AstroNvim/astrocore",
-    ---@type AstroCoreOpts
-    opts = {
-      autocmds = {
-        helm_commentstring = {
-          {
-            event = "FileType",
-            pattern = "helm",
-            callback = function() vim.opt_local.commentstring = "{{/* %s */}}" end,
+    ---@param opts AstroCoreOpts
+    opts = function(_, opts)
+      local is_helm_file = function(path)
+        local check = vim.fs.find("Chart.yaml", { path = vim.fs.dirname(path), upward = true })
+        return not vim.tbl_isempty(check)
+      end
+      local yaml_filetype = function(path, _) return is_helm_file(path) and "helm.yaml" or "yaml" end
+      local tmpl_filetype = function(path, _) return is_helm_file(path) and "helm.tmpl" or "template" end
+      local tpl_filetype = function(path, _) return is_helm_file(path) and "helm.tmpl" or "smarty" end
+
+      return require("astrocore").extend_tbl(opts, {
+        autocmds = {
+          helm_commentstring = {
+            {
+              event = "FileType",
+              pattern = "helm",
+              callback = function() vim.opt_local.commentstring = "{{/* %s */}}" end,
+            },
           },
         },
-      },
-      filetypes = {
-        extension = {
-          gotmpl = "helm",
-          yaml = M.yaml_filetype,
-          yml = M.yaml_filetype,
-          tmpl = M.tmpl_filetype,
-          tpl = M.tpl_filetype,
+        filetypes = {
+          extension = {
+            gotmpl = "helm",
+            yaml = yaml_filetype,
+            yml = yaml_filetype,
+            tmpl = tmpl_filetype,
+            tpl = tpl_filetype,
+          },
+          filename = {
+            ["Chart.yaml"] = "yaml",
+            ["Chart.lock"] = "yaml",
+          },
+          pattern = {
+            ["helmfile.*%.ya?ml"] = "helm",
+          },
         },
-        filename = {
-          ["Chart.yaml"] = "yaml",
-          ["Chart.lock"] = "yaml",
-        },
-        pattern = {
-          ["helmfile.*%.ya?ml"] = "helm",
-        },
-      },
-    },
+      })
+    end,
   },
   {
     "nvim-treesitter/nvim-treesitter",
--- a/lua/astrocommunity/pack/helm/init.lua
+++ b/lua/astrocommunity/pack/helm/init.lua
@@ -1,20 +1,5 @@
 local M = {}
 
-function M.setup()
-  vim.filetype.add {
-    extension = {
-      yaml = M.yaml_filetype,
-      yml = M.yaml_filetype,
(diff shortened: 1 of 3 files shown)


lua/astrocommunity/pack/helm/init.lua
```diff
@@ -1,3 +1,37 @@
...
+      yml = M.yaml_filetype,
+      tmpl = M.tmpl_filetype,
+      tpl = M.tpl_filetype,
+    },
+    filename = {
+      ["Chart.yaml"] = "yaml",
+      ["Chart.lock"] = "yaml",
+    },
+  }
+end
```
Uzaaft, line 16: Why isn't this just moved into the astrocore spec?

lua/astrocommunity/pack/helm/init.lua
```diff
@@ -1,3 +1,37 @@
...
+--@return string
+function M.yaml_filetype(path, _) return M.is_helm_file(path) and "helm.yaml" or "yaml" end
+
+--@private
+--@return string
+function M.tmpl_filetype(path, _) return M.is_helm_file(path) and "helm.tmpl" or "template" end
+
+--@private
+--@return string
+function M.tpl_filetype(path, _) return M.is_helm_file(path) and "helm.tmpl" or "smarty" end
```
Uzaaft, line 33: These can be moved into the opts, and convert the opts stuff to the function notation.

lua/astrocommunity/pack/helm/init.lua
Zebradil: Sorry, I actually meant to do that.

lua/astrocommunity/pack/helm/init.lua
Zebradil: Are there any examples on how to do that? My lua proficiency level is "copy someone else's code around".

Zebradil: Ah, wait a second, I think I got it.

Zebradil: Did you mean something like this?

```lua
  {
    "AstroNvim/astrocore",
    ---@type AstroCoreOpts
    opts = {
      autocmds = {
        helm_commentstring = {
          {
            event = "FileType",
            pattern = "helm",
            callback = function() vim.opt_local.commentstring = "{{/* %s */}}" end,
          },
        },
      },
      filetypes = {
        extension = {
          gotmpl = "helm",
          yaml = function(path, _) return M.is_helm_file(path) and "helm.yaml" or "yaml" end,
          yml = function(path, _) return M.is_helm_file(path) and "helm.yaml" or "yaml" end,
          tmpl = function(path, _) return M.is_helm_file(path) and "helm.tmpl" or "template" end,
          tpl = function(path, _) return M.is_helm_file(path) and "helm.tmpl" or "smarty" end,
        },
        filename = {
          ["Chart.yaml"] = "yaml",
          ["Chart.lock"] = "yaml",
        },
        pattern = {
          ["helmfile.*%.ya?ml"] = "helm",
        },
      },
    },
  },
```

I still need to keep the `is_helm_file` function somewhere. And the function for yaml and yml extensions is repeated.

lua/astrocommunity/pack/helm/init.lua
Zebradil: Removed.

lua/astrocommunity/pack/helm/init.lua
Uzaaft: No. More like:
```lua
opts = function(_, opts)
-- Do opts stuff here
end
```

And define stuff inside the function.

Zebradil: I have updated the code. Now opts is a function which defines local helper functions inside. Did I get your idea right?

Uzaaft:
I'm just gonna ask for a second set of eyes from @AstroNvim/astrocommunity-maintainers

mehalter (commented):
One thing we may need to start thinking about (and may require fixing other packs) is how we will have these compound file type detection functions play nicely with each other. The filetype detection conflicts with other language packs like the audible pack.

Also is it true that all gotmpl files are helm files 🤔 that seems wrong

Zebradil:
> have these compound file type detection functions play nicely with each other.

I don't think it's possible to do this in an automated way. In cases of conflicts, users will have to resolve them on a per-case basis. With time and reports from other users, we can figure out better rules to minimize conflicts.

> Also is it true that all gotmpl files are helm files 🤔 that seems wrong

I don't know, this was in the pack [from the start](https://github.com/AstroNvim/astrocommunity/commit/f8fbe8404559e8314b4096cffa20e20c1f08fbfe), added by @Sacro.

I don't work a lot with non-standard Helm templates. So far, I have seen the following files that are Helm templates:

- `NOTES.txt` (`helm.tmpl`)
- `**/*.yaml` in a chart directory (`helm.yaml`)
- `helmfile.yaml` (`helm.yaml`)
- `something.yaml.gotmpl` (`helm.yaml`)

With the current rules in this PR, I have false positives where `values.yaml` is detected as `helm.yaml` instead of `yaml`. So I was thinking whether the rules should be changed to be more precise, but I probably haven't seen all possible cases so far to make a good guess about the best set of rules.

mehalter:
> I don't think it's possible to do this in an automated way. In cases of conflicts, users will have to resolve them on a per-case basis. With time and reports from other users, we can figure out better rules to minimize conflicts.

I can fix this, I will push a fix to this branch on Monday when I get to a computer again. I will also fix the rest of the filetype additions throughout AstroCommunity

Sacro:
> I don't know, this was in the pack [from the start](https://github.com/AstroNvim/astrocommunity/commit/f8fbe8404559e8314b4096cffa20e20c1f08fbfe), added by @Sacro.

I've slept since then, I've looked through Discord and I can't remember where that came from or why I left it in.

mehalter:
@Zebradil I just pushed a commit that adds filetype merging as a utility in AstroCommunity and uses it for both the ansible pack and the helm pack. I have tested them together and they play nicely together! Also I reverted the change that all of these should map to the `helm` filetype and not `helm.yaml`/`helm.tmpl`. This breaks the language server and I can't find any example of those filetypes being used anywhere. Let me know what you think of the changes!

Zebradil:
> I just pushed a commit that adds filetype merging as a utility in AstroCommunity and uses it for both the ansible pack and the helm pack.

Awesome, that's really neat 👍 

> I reverted the change that all of these should map to the `helm` filetype and not `helm.yaml`/`helm.tmpl`. This breaks the language server and I can't find any example of those filetypes being used anywhere.

Yeah, sorry for that. I inspected the origins of this idea (to use `helm.yaml` instead of `helm`) more carefully and it seems like it was not finished and it requires more configuration (to parser and to language server) to work. I'm puzzled why it didn't break when I was testing this, but I'd leave it for now. Maybe I figure that out and file another PR.

It'd be nice to merge this if there are no more suggestions. The question about code attribution is not relevant anymore, as there is nothing left from the original code.

mehalter (approved):

PR #1172, opened 2024-08-21
fix(tint-nvim): update neo-tree → NeoTree (ignore_patterns)

jondkinney:
## 📑 Description

In the `highlight_ignore_patterns`, `neo-tree` wasn't working at all, so I tried `NeoTree`. That worked except for the icons still dim. That's an acceptable limitation for now IMO. Perhaps in the future we can fix it so the icon hl group doesn't dim, either. But this is less broken than it was.

### DIFF
--- a/lua/astrocommunity/color/tint-nvim/init.lua
+++ b/lua/astrocommunity/color/tint-nvim/init.lua
@@ -2,7 +2,7 @@ return {
   "levouh/tint.nvim",
   event = "User AstroFile",
   opts = {
-    highlight_ignore_patterns = { "WinSeparator", "neo-tree", "Status.*" },
+    highlight_ignore_patterns = { "WinSeparator", "NeoTree", "Status.*" },
     tint = -45, -- Darken colors, use a positive value to brighten
     saturation = 0.6, -- Saturation to preserve
   },


Uzaaft (approved):

Uzaaft:
neo-tree is the filetype. 
@jondkinney The icons are in the highlights group NeoTreeFileIcon. Perhaps that's useful info for you :)

jondkinney:
@Uzaaft I actually found a more effective way to do this.

```lua
return {
  "levouh/tint.nvim",
  event = "User AstroFile",
  opts = function()
    local transforms = require "tint.transforms"

    -- Custom function to determine if a window is NeoTree
    local function is_neotree(buf)
      local ft = vim.bo[buf].filetype
      return ft == "neo-tree"
    end

    return {
      highlight_ignore_patterns = {
        "WinSeparator",
        "Status.*",
      },
      window_ignore_function = function(winid)
        local buf = vim.api.nvim_win_get_buf(winid)
        return is_neotree(buf)
      end,
      transforms = {
        transforms.tint_with_threshold(-45, "#2E3440", 135),
        transforms.saturate(0.7),
      },
    }
  end,
}

```

The transforms are specific to my colorscheme (nord), so those aren't probably helpful in the community version.

PR #1175, opened 2024-08-22
feat(split-and-window): add `neominimap.nvim`

mehalter:
## 📑 Description

This adds the [`neominimap.nvim`](https://github.com/Isrothy/neominimap.nvim) plugin


Uzaaft (approved):
Impeccable as always Signore :shipit:

PR #1177, opened 2024-08-23
docs(python): Swap mentions of `pyright` with `basedpyright`

jcromero:
fix(python): Updated info about language server used in pack


Uzaaft (approved):
🤓

PR #1183, opened 2024-08-31
fix(python-ruff): Ensure formatters are called for python-ruff pack.

gwww:
## 📑 Description

Using a freshly cloned [template](https://github.com/AstroNvim/template), changing only the file `lua/astrocommunity.lua` to: 
```
---@type LazySpec
return {
  "AstroNvim/astrocommunity",
  { import = "astrocommunity.pack.python-ruff" },
}
```
then only the `ruff_format` formatter is called when saving a Python file. The `ruff_organize_imports` formatter is not called.

This PR adds `conform` configuration that ensures that both formatters are called.

### DIFF
--- a/lua/astrocommunity/pack/python-ruff/init.lua
+++ b/lua/astrocommunity/pack/python-ruff/init.lua
@@ -40,6 +40,9 @@ return {
     "stevearc/conform.nvim",
     optional = true,
     opts = {
+      format_on_save = {
+        lsp_format = "fallback",
+      },
       formatters_by_ft = {
         python = { "ruff_organize_imports", "ruff_format" },
       },


mehalter:
For clarity, without conform installed, by default it uses ruff's language server for formatting which _doesn't_ support organizing imports on format. If you install conform it sets up to just use the ruff command line to do the formatting and not the language server.

gwww:
I understand,  but with conform installed, the command line for sorting the imports on write is not called without this PR. To me formatting is code formatting and sorting, no matter what command is doing it. Is that the difference?

mehalter:
Import sorting works without this PR. What this PR is saying is "fallback to the LSP if no formatter are configured. It would change no behavior because there _are_ formatters configured. This is not something that should be in this pack.

gwww:
Import sorting on file save does not work for me with the base template, conform enabled, and this pack installed. I understand this PR may not be the correct way to solve it.

mehalter:
What does your StatusLine show is attached to the buffer when you open a Python file with the pack and conform added

mehalter:
Also how are you adding conform

mehalter:
The good news is that I was trying to say, is if this PR fixes the problem, then without this PR it would also be working since this change wouldn't affect any of the behavior

gwww:
Just tested again from scratch.

Conform enabled using this:
```
return {
  "stevearc/conform.nvim",
}
```

I have exactly two changes to the default template. The above and enabling the community pack. (In the PR I said one change, which is incorrect).

Heirline:
Screenshot 2024-09-02 at 09 43 39

(Sorry it is cut off)

Without PR sorting on save does not get called. With PR sorting on save works.

mehalter:
Oh you just are just not setting up conform at all. I would recommend installing the conform plugin from AstroCommunity as it sets up its interaction with things.

mehalter:
https://github.com/AstroNvim/astrocommunity/blob/main/lua%2Fastrocommunity%2Fediting-support%2Fconform-nvim%2Finit.lua

gwww:
I had no idea about that pack! Now I understand. 

Does it make sense to add a line to the readme recommending that pack be installed too?

mehalter:
No, it's just an option for the user if they want to use conform. We recommend not using it for the out of the box experience of the LSP rather than recommending users the much more manual configuration if conform. It's just up to them to opt into it

PR #1185, opened 2024-09-02
feat(neotest): add watch keybindings

cristobalgvera:
## 📑 Description

Add mappings to simplify the usage of watch-related commands when using `neotest`.

It also refactor the keybindings to follow current standards using an `opts` function instead of a table.

## ℹ Additional Information

The API for the usage of `neotest.watch.stop()` is not well defined in terms of the type information for the usage.

This creates the necessity to use `@diagnostic disable-next-line: missing-parameter`.

For a closer look, see https://github.com/nvim-neotest/neotest/blob/master/doc/neotest.txt#L626

### DIFF
--- a/lua/astrocommunity/test/neotest/init.lua
+++ b/lua/astrocommunity/test/neotest/init.lua
@@ -1,4 +1,6 @@
 local prefix = "<Leader>T"
+local watch_prefix = prefix .. "W"
+
 return {
   "nvim-neotest/neotest",
   lazy = true,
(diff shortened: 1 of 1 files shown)


cristobalgvera:
I'm thinking to refactor this after the PR approval, in order to follow the standard process to map keybingings used in [here](https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/code-runner/overseer-nvim/init.lua#L38-L53)

Uzaaft:
> I'm thinking to refactor this after the PR approval, in order to follow the standard process to map keybingings used in [here](https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/code-runner/overseer-nvim/init.lua?rgh-link-date=2024-09-02T13%3A55%3A32Z#L38-L53)

Refactor first please. :)

lua/astrocommunity/test/neotest/init.lua
Uzaaft, line 1: Just as a general comment @cristobalgvera. Try and keep the diff to a minimum when submitting PR's. A big chunk of the diff in this PR could have been avoided if you just added the astroui table above the existing astrocore table.

Uzaaft (approved):

PR #1186, opened 2024-09-02
feat(nvim-coverage): add keybindings

cristobalgvera:
## 📑 Description

Add mappings to `nvim-coverage`.

Also, it was added the option to reload the coverage report if it changes, a useful option complemented with the ease to show the coverage added with the mappings.

## ℹ Additional Information

The mapping was added as a second layer behind `Tests` map. This is because `neotest` and `vim-test` use the `<Leader>T` map and name it like that, allowing to complement the usage of those plugins without depending on it.

### DIFF
--- a/lua/astrocommunity/test/nvim-coverage/init.lua
+++ b/lua/astrocommunity/test/nvim-coverage/init.lua
@@ -2,7 +2,9 @@
 return {
   "andythigpen/nvim-coverage",
   event = "User AstroFile",
-  opts = {},
+  opts = {
+    auto_reload = true,
+  },
   dependencies = {
     "nvim-lua/plenary.nvim",
     {
--- a/lua/astrocommunity/test/nvim-coverage/init.lua
+++ b/lua/astrocommunity/test/nvim-coverage/init.lua
@@ -1,6 +1,42 @@
+---@type LazySpec
 return {
   "andythigpen/nvim-coverage",
   event = "User AstroFile",
-  dependencies = { "nvim-lua/plenary.nvim" },
   opts = {},
+  dependencies = {
+    "nvim-lua/plenary.nvim",
+    {
+      "AstroNvim/astroui",
+      opts = {
+        icons = {
+          Tests = "󰗇",
+          Coverage = "",
+        },
+      },
+    },
+    {
+      "AstroNvim/astrocore",
+      optional = true,
+      opts = function(_, opts)
+        local astroui = require "astroui"
+        local maps = opts.mappings
+
+        local tests_prefix = "<Leader>T"
+        local coverage_prefix = tests_prefix .. "C"
+
+        -- INFO: Compatibility with `neotest` and `vim-test`
+        maps.n[tests_prefix] = { desc = astroui.get_icon("Tests", 1, true) .. "Tests" }
+
+        maps.n[coverage_prefix] = { desc = astroui.get_icon("Coverage", 1, true) .. "Coverage" }
+        maps.n[coverage_prefix .. "t"] = { function() require("coverage").toggle() end, desc = "Toggle coverage" }
+        maps.n[coverage_prefix .. "s"] =
+          { function() require("coverage").summary() end, desc = "Show coverage summary" }
+        maps.n[coverage_prefix .. "c"] = { function() require("coverage").clear() end, desc = "Clear coverage" }
+        maps.n[coverage_prefix .. "l"] = {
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/test/nvim-coverage/init.lua
```diff
@@ -1,6 +1,33 @@
...
+              function() require("coverage").load(true) end,
+              desc = "Load and show coverage",
+            },
+            [coverage_prefix .. "t"] = { function() require("coverage").toggle() end, desc = "Toggle coverage" },
+            [coverage_prefix .. "s"] = { function() require("coverage").summary() end, desc = "Show coverage summary" },
+            [coverage_prefix .. "c"] = { function() require("coverage").clear() end, desc = "Clear coverage" },
+          },
+        },
+      },
+    },
```
ALameLlama, line 31: The key mapping should be using this style
https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/code-runner/overseer-nvim/init.lua#L38-L53

opts will be a function and you can move the local prefix to the function

lua/astrocommunity/test/nvim-coverage/init.lua
```diff
@@ -1,6 +1,44 @@
...
+      opts = function(_, opts)
+        local coverage = require "coverage"
+        local astroui = require "astroui"
+        local maps = opts.mappings
+
+        local tests_prefix = "<Leader>T"
+        local coverage_prefix = tests_prefix .. "C"
+
+        -- INFO: Compatibility with `neotest` and `vim-test`
+        maps.n[tests_prefix] = { desc = astroui.get_icon("Tests", 1, true) .. "Tests" }
```
Uzaaft, line 31:
```suggestion
        maps.n["<Leader>T"] = { desc = astroui.get_icon("Tests", 1, true) .. "Tests" }
```

lua/astrocommunity/test/nvim-coverage/init.lua
```diff
@@ -1,6 +1,44 @@
...
+    },
+    {
+      "AstroNvim/astrocore",
+      optional = true,
+      opts = function(_, opts)
+        local coverage = require "coverage"
+        local astroui = require "astroui"
+        local maps = opts.mappings
+
+        local tests_prefix = "<Leader>T"
```
Uzaaft, line 27:
```suggestion
```

lua/astrocommunity/test/nvim-coverage/init.lua
```diff
@@ -1,6 +1,44 @@
...
+        local coverage_prefix = tests_prefix .. "C"
+
+        -- INFO: Compatibility with `neotest` and `vim-test`
+        maps.n[tests_prefix] = { desc = astroui.get_icon("Tests", 1, true) .. "Tests" }
+
+        maps.n[coverage_prefix] = { desc = astroui.get_icon("Coverage", 1, true) .. "Coverage" }
```
Uzaaft, line 33:
```suggestion
        maps.n[coverage_prefix] = { desc = require("astroui").get_icon("Coverage", 1, true) .. "Coverage" }
```

Uzaaft:
Ah now I see the other usage of test_prefix. My bad. 2s

lua/astrocommunity/test/nvim-coverage/init.lua
```diff
@@ -1,6 +1,44 @@
...
+          Tests = "󰗇",
+          Coverage = "",
+        },
+      },
+    },
+    {
+      "AstroNvim/astrocore",
+      optional = true,
+      opts = function(_, opts)
+        local coverage = require "coverage"
```
ALameLlama, line 23: I'll need to double check but having this `require "coverage"` out here instead of in the functions e.g

`maps.n[coverage_prefix .. "t"] = { function() require("coverage").toggle() end, desc = "Toggle coverage" }`

Will force this to always load instead of lazy loading it, since astrocore is always and will always require it right away.

cristobalgvera: Ok, tell me if we need to change it, I understand the issue.

ALameLlama: Just tried it out it's preventing lazy loading and always loading it in, can you inline the require coverage? thanks :)

ALameLlama (changes requested):

ALameLlama (approved):

PR #1187, opened 2024-09-02
fix(markdown-preview-nvim): build using yarn when possible

cristobalgvera:
## 📑 Description

Fix the installation process of `markdown-preview-nvim`.

This PR uses the recommendations given in the [comment](https://github.com/iamcco/markdown-preview.nvim/issues/690#issuecomment-2254280534) related with the reported issue about installing the plugin using `lazy.nvim`.

## ℹ Additional Information

The plugin [Markdown Preview](https://github.com/iamcco/markdown-preview.nvim) seems to be unmaintained and the installation process provided in his documentation [fails when installed through `lazy.nvim`](https://github.com/iamcco/markdown-preview.nvim/issues/690).

Also, I added the variable `COREPACK_ENABLE_AUTO_PIN=0` to avoid the modification of the `package.json` ([described here](https://github.com/nodejs/corepack/blob/main/README.md#environment-variables)) used by the plugin, in case the user is using `Corepack`.

### DIFF
--- a/lua/astrocommunity/markdown-and-latex/markdown-preview-nvim/init.lua
+++ b/lua/astrocommunity/markdown-and-latex/markdown-preview-nvim/init.lua
@@ -1,7 +1,22 @@
 ---@type LazySpec
 return {
   "iamcco/markdown-preview.nvim",
-  build = function() vim.fn["mkdp#util#install"]() end,
+  build = function(plugin)
+    local package_manager = vim.fn.executable "yarn" and "yarn" or vim.fn.executable "npx" and "npx -y yarn" or false
+
+    --- HACK: Use `yarn` or `npx` when possible, otherwise throw an error
+    ---@see https://github.com/iamcco/markdown-preview.nvim/issues/690
+    ---@see https://github.com/iamcco/markdown-preview.nvim/issues/695
+    if not package_manager then error "Missing `yarn` or `npx` in the PATH" end
+
+    local cmd = string.format(
+      "!cd %s && cd app && COREPACK_ENABLE_AUTO_PIN=0 %s install --frozen-lockfile",
+      plugin.dir,
+      package_manager
+    )
+
+    vim.cmd(cmd)
+  end,
   ft = { "markdown", "markdown.mdx" },
   cmd = { "MarkdownPreviewToggle", "MarkdownPreview", "MarkdownPreviewStop" },
   init = function()
--- a/lua/astrocommunity/markdown-and-latex/markdown-preview-nvim/README.md
+++ b/lua/astrocommunity/markdown-and-latex/markdown-preview-nvim/README.md
@@ -1,5 +1,8 @@
 # markdown-preview.nvim
 
+> This plugin requires Node.js to be installed on your system.
+> Without it, the installation process will fail, and the plugin won't work.
+
 markdown preview plugin for (neo)vim
 
 **Repository:** <https://github.com/iamcco/markdown-preview.nvim>


lua/astrocommunity/markdown-and-latex/markdown-preview-nvim/init.lua
```diff
@@ -1,6 +1,15 @@
...
-  build = function() vim.fn["mkdp#util#install"]() end,
+  build = function(plugin)
+    -- HACK: Install using `npx` (using `yarn`) if available
+    ---@see https://github.com/iamcco/markdown-preview.nvim/issues/690#issuecomment-2254280534
+    if vim.fn.executable "npx" then
+      vim.cmd("!cd " .. plugin.dir .. " && cd app && COREPACK_ENABLE_AUTO_PIN=0 npx -y yarn install --frozen-lockfile")
+    else
+      vim.cmd [[Lazy load markdown-preview.nvim]]
+      vim.fn["mkdp#util#install"]()
+    end
```
Uzaaft, line 11: I'm not sure if this should be the default. Any thoughts? @AstroNvim/astrocommunity-maintainers

ALameLlama: I'm down for this if it means a better out-of-the-box experience for the community, and most of the time I have yarn installed but could this cause issues if people don't have yarn?

If we add this we probs want to include something in the readme

cristobalgvera: This change doesn't require you to have Yarn installed, just npx (NodeJS), and if you doesn't have it, it just go with the old approach.

mehalter: I feel like the default should probably be `yarn` and fall back to `npx`. It also might make sense to contribute this upstream to the plugin itself to improve its installation function. The. We could revert this change and just call their install function

cristobalgvera: > I feel like the default should probably be `yarn` and fall back to `npx`. It also might make sense to contribute this upstream to the plugin itself to improve its installation function. The. We could revert this change and just call their install function

The Markdown Preview repository seems to be unmaintained. I'll make the PR, tho.

mehalter: Ok! We can definitely merge this in the meantime, just like promoting contributions to more projects 🤩

cristobalgvera: > I feel like the default should probably be `yarn` and fall back to `npx`.

I'll make this change to, give me a sec.

Uzaaft: > lt should probably be `yarn` and fall back to `npx`. It also might make sense to contribute this upstream to the plugin itself to improve its installation function. The. We could revert this change and just call their install function

I agree with this.

mehalter (changes requested):
If yarn is installed it should use the default installation function. If it's not then it should do the `npx` hack.

cristobalgvera:
> If yarn is installed it should use the default installation function. If it's not then it should do the `npx` hack.

Trying this solution I found another problem that was reported [as an issue](https://github.com/iamcco/markdown-preview.nvim/issues/695).

I think that the best out-of-the-box experience that we can give is install the plugin using the explicit installation method that this PR gives.

@mehalter, let me know what do you think.

cristobalgvera:
@mehalter, maybe we could put this in the `README` of the plugin, something like _"Node is required in order to use this plugin"_ and block the build process provided by the plugin, at least while the project itself doesn't solve it.

lua/astrocommunity/markdown-and-latex/markdown-preview-nvim/init.lua
```diff
@@ -1,7 +1,24 @@
...
+      local cmd = string.format(
+        "!cd %s && cd app && COREPACK_ENABLE_AUTO_PIN=0 %s install --frozen-lockfile",
+        plugin.dir,
+        package_manager
+      )
+
+      vim.cmd(cmd)
+    else
+      vim.cmd [[Lazy load markdown-preview.nvim]]
+      vim.fn["mkdp#util#install"]()
```
mehalter, line 19:
```suggestion
    local yarn = vim.fn.executable "yarn" == 1 and "yarn" or vim.fn.executable "npx" == 1 and "npx -y yarn" or false

    if yarn then
      vim.cmd(("!cd %s && cd app && COREPACK_ENABLE_AUTO_PIN=0 %s install --frozen-lockfile"):format(plugin.dir, yarn))
    else
      vim.notify("`yarn` or `npx` must be in the PATH to build `markdown-preview.nvim`", vim.log.levels.ERROR)
```

If we are checking for `yarn` and `npx` then we really don't have a reason to fall back on the install util at all. We can just throw an error to the user and move on

mehalter (changes requested):
Ah I see what you mean with those issues. Yeah, I think it's a good idea to just manually do it like such. Recommended a bit of a refactor, if we are checking for `yarn` and `npx` manually, then there really is not reason to fall back to the default installation function since it wouldn't work anyway. Instead we can just throw an error to the user :)

lua/astrocommunity/markdown-and-latex/markdown-preview-nvim/init.lua
cristobalgvera: Yeah, I was thinking exactly in that. I'll implement it.

cristobalgvera: I was not able _(or I don't know how)_ to render the message in the screen as a notification if the `build` process fails. So, I decided to simply thrown an exception, this way the user can see the actual error.

image

Also, I included this conflicting behavior in the `README`.

mehalter (approved):

PR #1189, opened 2024-09-02
feat(typescript): add neotest-jest support

cristobalgvera:
## 📑 Description

In `typescript` pack, add support for [`neotest-jest`](https://github.com/nvim-neotest/neotest-jest) if [`neotest`](https://github.com/nvim-neotest/neotest) is installed.

## ℹ Additional Information

The usage of `table#insert` function allows to use the same pattern in other language packs without overriding the values (e.g.: use `rustaceanvim` integration for `neotest` in the same `neovim` configuration).

This usage was inspired from the `rust` [language pack](https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/pack/rust/init.lua#L83-L91).

### DIFF
--- a/lua/astrocommunity/pack/typescript/init.lua
+++ b/lua/astrocommunity/pack/typescript/init.lua
@@ -236,7 +236,7 @@ return {
   {
     "nvim-neotest/neotest",
     optional = true,
-    dependencies = { "nvim-neotest/neotest-jest" },
+    dependencies = { { "nvim-neotest/neotest-jest", config = function() end } },
     opts = function(_, opts)
       if not opts.adapters then opts.adapters = {} end
       table.insert(opts.adapters, require "neotest-jest"(require("astrocore").plugin_opts "neotest-jest"))
--- a/lua/astrocommunity/pack/typescript/init.lua
+++ b/lua/astrocommunity/pack/typescript/init.lua
@@ -233,4 +233,13 @@ return {
       },
     },
   },
+  {
+    "nvim-neotest/neotest",
+    optional = true,
+    dependencies = { "nvim-neotest/neotest-jest" },
+    opts = function(_, opts)
+      if not opts.adapters then opts.adapters = {} end
+      table.insert(opts.adapters, require "neotest-jest"(require("astrocore").plugin_opts "neotest-jest"))
+    end,
+  },
 }
--- a/lua/astrocommunity/pack/typescript/README.md
+++ b/lua/astrocommunity/pack/typescript/README.md
@@ -30,10 +30,7 @@ return {
 
 ## How do I customize `neotest-jest`?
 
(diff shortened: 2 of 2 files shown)


lua/astrocommunity/pack/typescript/init.lua
```diff
@@ -233,4 +233,13 @@ return {
       },
     },
   },
+  {
+    "nvim-neotest/neotest",
+    optional = true,
+    dependencies = { "nvim-neotest/neotest-jest" },
+    opts = function(_, opts)
+      if not opts.adapters then opts.adapters = {} end
+      table.insert(opts.adapters, require "neotest-jest" {})
```
mehalter, line 242:
```suggestion
      table.insert(opts.adapters, require "neotest-jest"(require("astrocore").plugin_opts "neotest-jest"))
```

This allows the neotest adapter to be configured easily through it's `opts` table while still getting set up in `neotest`

mehalter (changes requested):

lua/astrocommunity/pack/typescript/init.lua
cristobalgvera: Nice, I didn't know about it.

cristobalgvera: @mehalter I also added a documentation about how to add options in the `README`, because the plugin doesn't provide a `setup` function, so `lazy` throw an exception when trying to do something like

```lua
return {
  "nvim-neotest/neotest-jest",
  opts = function(_, opts)
    opts.jestCommand = "npm test --"
  end,
}
```

The workaround is simple, add a `config` key with an empty `function`. I don't know if there is a better approach in this case, tell me what do you think.

lua/astrocommunity/pack/typescript/init.lua
```diff
@@ -233,4 +233,13 @@ return {
       },
     },
   },
+  {
+    "nvim-neotest/neotest",
+    optional = true,
+    dependencies = { "nvim-neotest/neotest-jest" },
```
mehalter, line 239:
```suggestion
    dependencies = { { "nvim-neotest/neotest-jest", config = function() end } },
```

mehalter (changes requested):

lua/astrocommunity/pack/typescript/init.lua
mehalter: https://github.com/AstroNvim/astrocommunity/commits/main/?since=2024-09-04&until=2024-09-05

just pushed a similar fix everywhere else we have neotest adapters that can be configured

mehalter (approved):

cristobalgvera:
Thanks @mehalter, I'm busy today and will upload the changes at the afternoon.

mehalter:
No worries! Just wanted to help get this through since it was a small change :) Thanks again! This was a great addition plus exposed bugs in other packs :D

PR #1198, opened 2024-09-08
docs(sql): Document required dependency.

Uzaaft:


ALameLlama (approved):
Love me some documentation

PR #1204, opened 2024-09-10
refactor(lazydocker): replace keybindings table by function instead

cristobalgvera:
## 📑 Description

Standardize the keybinding of `lazydocker.nvim` to be a `function` instead of an opts `table`.

Also, it was removed the unnecesary `opts`.


Uzaaft:
I dont see why would we do this?

ALameLlama:
It would be nice to have it "Standardize" so all plugins use the same method imo

lua/astrocommunity/docker/lazydocker/init.lua
```diff
@@ -1,22 +1,23 @@
...
+          local prefix = "<Leader>t"
+
+          maps.n[prefix .. "d"] = {
+            function() require("lazydocker").open() end,
+            desc = "ToggleTerm LazyDocker",
+          }
+        end,
       },
     },
-    opts = {},
```
luxus, line 20: this would run, the `setup()` of `lazydocker.nvim` 
https://github.com/crnvl96/lazydocker.nvim?tab=readme-ov-file#requirements

mehalter: Yeah this is good

cristobalgvera: @luxus yeah, I also checked what happens inside `lazydocker.nvim` if opts are `nil`, and it [will replaced by `{}`](https://github.com/mgierada/lazydocker.nvim/blob/38c0fbc3cb77e38fd2cdc70b709cf841a38782f5/lua/lazydocker/config.lua#L20-L22).

mehalter: If you don't have it here the setup function is never called. This is necessary to have :) or you can change it to `config = true` which will pass in `nil` to the setup function. But it's better to just leave it untouched rather than breaking it by removing it

mehalter (approved):

PR #1210, opened 2024-09-15
fix(avante-nvim): update correct plug mappings

aarnphm:
## 📑 Description

Avante will now provide Plug mapping instead of commands for some functionalities to avoid poluting namespace from users.

### DIFF
--- a/lua/astrocommunity/completion/avante-nvim/init.lua
+++ b/lua/astrocommunity/completion/avante-nvim/init.lua
@@ -7,6 +7,9 @@ return {
     "AvanteEdit",
     "AvanteRefresh",
     "AvanteSwitchProvider",
+    "AvanteChat",
+    "AvanteToggle",
+    "AvanteClear",
   },
   dependencies = {
     "stevearc/dressing.nvim",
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/completion/avante-nvim/init.lua
```diff
@@ -4,16 +4,6 @@ return {
...
-    "AvanteConflictChooseCursor",
-    "AvanteConflictChooseNone",
-    "AvanteConflictChooseOurs",
-    "AvanteConflictChooseTheirs",
-    "AvanteConflictListQf",
-    "AvanteConflictNextConflict",
-    "AvanteConflictPrevConflict",
     "AvanteEdit",
     "AvanteRefresh",
     "AvanteSwitchProvider",
```
mehalter, line 9:
```suggestion
    "AvanteSwitchProvider",
    "AvanteChat",
    "AvanteToggle",
    "AvanteClear",
```

I think with this we should also add in some missing commands as well!

mehalter (changes requested):
Thanks! This is great, made one comment to help out

mehalter (approved):
LGTM, thanks!

PR #1212, opened 2024-09-19
feat(grug-far-nvim)!: add more features and align with recommended `spectre.nvim` mappings

mehalter:
## 📑 Description

This does a major refactor on the `grug-far.nvim` spec that adds a lot more bindings and options for searching and replacing. It also aligns the recommended keymaps with what we have been using for `spectre.nvim` which is a very nice workflow.


luxus (approved):
lgtm

PR #1213, opened 2024-09-19
feat(go): Add golangcilint as default linter

shivanthzen:
## 📑 Description

The PR adds [golangcilint](https://github.com/golangci/golangci-lint) as the default linter for go.pack. Currently no linters as part for the pack.

### DIFF
--- a/lua/astrocommunity/pack/go/README.md
+++ b/lua/astrocommunity/pack/go/README.md
@@ -15,3 +15,4 @@ This plugin pack does the following:
   - [impl](https://github.com/josharian/impl)
 - Adds [nvim-dap-go](https://github.com/leoluz/nvim-dap-go) for debugging
 - Adds [gopher.nvim](https://github.com/olexsmir/gopher.nvim) for language specific tools
+- Adds [golangcilint](https://github.com/golangci/golangci-lint) for linting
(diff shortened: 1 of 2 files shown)


shivanthzen:
Thanks for the revision
LGTM.

mehalter (approved):

zedeus:
For me it doesn't get updated in real-time unlike gopls which already uses e.g. staticcheck for linting. It results in both lingering and duplicated warnings for me:
After save:
image
After satisfying the check, before saving:
image

I'd like either of the following options:
- Improve the golangcilint integration to be real-time without having to edit just like the default LSP is (maybe by using null-ls instead of nvim-lint? I don't know very much about plugin landscape so maybe this is nonsensical)
- A way to easily disable golangcilint (not sure how to go about it, do I have to remove that element from the array in my configuration?)

mehalter:
@zedeus yeah it might be better for this not to be a default is gopls already has linting. Then users can opt into it. If you open a PR to revert this then I will merge it

zedeus:
@mehalter here you go: https://github.com/AstroNvim/astrocommunity/pull/1215

shivanthzen:
The revert makes sense. 
Is `A way to easily disable golangci_lint (not sure how to go about it, do I have to remove that element from the array in my configuration?)` this a possbility right now in Astronvim ? Can I override pack configurations ? I only know of `override.lua` which is sort of per plugin configuration.

mehalter:
you can add it anywhere in your `plugins/` folder. It doesn't have to go in `override.lua`

PR #1216, opened 2024-09-21
fix(mini-indentscope): add animation

fynnfluegge:
## 📑 Description

Adds an animation to `mini-indentscope` with faster border rendering. Personally the slow rendering is very distracting to my eyes. Raising a PR since it feels this distraction is a common perception.

### Before
ezgif-5-af0f5dddd8

### After
ezgif-5-343c8f0050

### DIFF
--- a/lua/astrocommunity/indent/mini-indentscope/init.lua
+++ b/lua/astrocommunity/indent/mini-indentscope/init.lua
@@ -26,6 +26,10 @@ return {
   event = "User AstroFile",
   opts = function()
     return {
+      draw = {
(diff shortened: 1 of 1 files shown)


luxus:
we normally stick to the default. 
u always can overwrite the setting for yourself locally.
i guess it would be unexpected have different defaults

fynnfluegge:
Alright, thought this would be a better default. Closing it, thanks for the response!

luxus:
it makes more sense to ask upstream to change the default.. but i guess mini will not change it

PR #1217, opened 2024-09-22
fix(copilotchat-nvim): use custom prefix from astro core if defined

fynnfluegge:
## 📑 Description

With this PR it is possible to define the prefix for CopilotChat.nvim via `AstroNvim/astrocore` `opts.options.g.copilot_chat_prefix` or if not defined use `<leader>P` as default.


luxus (approved):
lgtm

fynnfluegge:
Hi @luxus, somehow the custom leader I defined in `astrocore` as `opts.options.g.copilot_chat_prefix` is not present when loading this plugin. Still it uses `<leader>P`. Do you know what the issue is? Possible that `astrocore.opt.options.g` is initialised later?

PR #1220, opened 2024-09-25
fix(tabout-nvim): make `nvim-cmp` optional and add support for `vim.snippet`

mehalter:
## 📑 Description

This cleans up the `tabout.nvim` spec to make sure it doesn't require `nvim-cmp` and also adds support for native vim snippets integration


luxus (approved):
lgtm

PR #1225, opened 2024-09-27
fix(pack): pass `astrolsp.on_attach()` to `moonbit.nvim`

rami3l:
## 📑 Description

After installing the `moonbit` pack, `moonbit-lsp` still won't launch properly.
This looks like an oversight in #1224. (cc @mehalter)

Tested locally with the following override:

```lua
  {
    "tonyfettes/moonbit.nvim",
    opts = function(_, opts)
      -- only enable treesitter if the plugin is available
      if not require("astrocore").is_available("nvim-treesitter") then
        opts.treesitter = { enabled = false }
      end
      -- only enable the LSP if the lsp command is executable
      if vim.fn.executable("moonbit-lsp") == 1 then
        local astrolsp_avail, astrolsp = pcall(require, "astrolsp")
        if astrolsp_avail then
          opts.lsp = astrolsp.lsp_opts("moonbit")
          opts.on_attach = astrolsp.on_attach
        end
      else
        opts.lsp = false
      end
    end,
  },
```

### DIFF
--- a/lua/astrocommunity/pack/moonbit/init.lua
+++ b/lua/astrocommunity/pack/moonbit/init.lua
@@ -12,7 +12,10 @@ return {
       -- only enable the LSP if the lsp command is executable
       if vim.fn.executable "moonbit-lsp" == 1 then
         local astrolsp_avail, astrolsp = pcall(require, "astrolsp")
-        if astrolsp_avail then opts.lsp = astrolsp.lsp_opts "moonbit" end
+        if astrolsp_avail then
+          opts.lsp = astrolsp.lsp_opts "moonbit"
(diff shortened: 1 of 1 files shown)


mehalter:
This is incorrect, sorry!

mehalter:
There is no opts.on_attach option in moonbit.nvim

mehalter:
I can confirm with a base installation with just this pack as is the moonbit LSP attached correctly

rami3l:
> There is no opts.on_attach option in moonbit.nvim

That's right, my bad :| Maybe I just forgot to restart Neovim...

PR #1228, opened 2024-10-02
fix(cpp): tbl extend on opts.config instead of opts

theo97490:
## 📑 Description

Fix tbl extend on opts.config instead of opts

I believe this was a typo because assigning opts.config to a modified table of opts doesn't make sense


theo97490:
Basicly i noticed there was an issue when i copied the pack.ccp to modify it in my own config ( by removing from the astrocommunity file and adding the init into the plugins folder ).  I saw it resets all configurations. I noticed when i was using pack.python and the type strictness was enabled where it shouldn't be

luxus (approved):
lgtm

PR #1237, opened 2024-10-09
fix(neorg): Add default workspace for Neorg

haphamdev:
With default workspace, Neorg can start with that workspace. The user does not have to run the command to load the workspace after starting nvim.


ALameLlama (approved):
LGTM, seems to match what other configs look like and if `notes` isn't your default it then you'd still need to enter the command just like before.

PR #1238, opened 2024-10-10
feat(completion): add magazine.nvim

ALameLlama:
## 📑 Description

Adds [magazine.nvim](https://github.com/iguanacucumber/magazine.nvim), This replaces nvim-cmp and acts as a "beta" with performance and feature improvements before they get up streamed.

### DIFF
--- a/lua/astrocommunity/completion/magazine-nvim/README.md
+++ b/lua/astrocommunity/completion/magazine-nvim/README.md
@@ -1,0 +1,8 @@
+# magazine.nvim
+
+> [!NOTE]
+> Magazine is a fork of `nvim-cmp`. The way this plugin is setup is to hijack into an existing `nvim-cmp` installation. So make sure that `nvim-cmp` is still installed and enabled for this plugin to work and all configuration is still done through `nvim-cmp` as normal.
+
+Magazine.nvim is a "beta" nvim-cmp to fix bugs & implement new features early
+
+**Repository:** <https://github.com/iguanacucumber/magazine.nvim>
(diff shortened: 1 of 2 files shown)


Uzaaft:
@ALameLlama  Shouldnt this replce nvim-cmp?

ALameLlama:
@Uzaaft Since this is using the same name as nvim-cmp it is replacing it, and you can't use specs to disable nvim-cmp since by that point it's already taken the name.

image

Unless I've misunderstood your comment or how we can uninstall nvim-cmp programmatically first before installing magazine.nvim

Uzaaft:
> @Uzaaft Since this is using the same name as nvim-cmp it is replacing it, and you can't use specs to disable nvim-cmp since by that point it's already taken the name.
> 
> image
> 
> Unless I've misunderstood your comment or how we can uninstall nvim-cmp programmatically first before installing magazine.nvim

I guess the way to go is to set the url of nvim-cmp, like this:

```lua
  { "hrsh7th/nvim-cmp", url = "https://github.com/iguanacucumber/magazine.nvim.git" },
``` 
Gonna cc in @mehalter in case he has an opinion.

mehalter:
I think I have a good solution for this. I've been also mulling over how to achieve this nicely. I'll give a recommendation on Monday

mehalter:
@ALameLlama I revised the commit to set up `magazine.nvim` to hijack an existing `nvim-cmp` installation and just be there to update the runtime path. So all the configuration for users is still done through `nvim-cmp` as normal and this plugin spec only exists to update the runtime path for where the `cmp` module lives.

Let me know what you think!

mehalter (approved):

PR #1239, opened 2024-10-11
feat(auto-save-nvim)!: swap to okuuva repo

### LINKED ISSUE #1233: use okuuva/auto-save.nvim instead of Pocco81/auto-save.nvim

not maintained

https://github.com/AstroNvim/astrocommunity/blob/c87c236dce71107637d2fc30e9dcc116d5d60097/lua/astrocommunity/editing-support/auto-save-nvim/init.lua#L2

### Describe the new feature

https://github.com/okuuva/auto-save.nvim is more maintained

### Additional context

use https://github.com/okuuva/auto-save.nvim?tab=readme-ov-file#%EF%B8%8F--events--callbacks 

instead of before_saving, after_saving?

but I cannot test - dont use formatters and `save on insertLeave`

my config is 

```
  {
    "okuuva/auto-save.nvim",
    opts = {
      trigger_events = { -- See :h events
        immediate_save = { 'BufLeave', 'TabLeave', 'FocusLost', 'VimLeavePre' }, -- vim events that trigger an immediate save
        -- defer_save = { "InsertLeave", "TextChanged" }, -- vim events that trigger a deferred save (saves after `debounce_delay`)
        defer_save = { }, -- vim events that trigger a deferred save (saves after `debounce_delay`)
        cancel_defered_save = { "InsertEnter" }, -- vim events that cancel a pending deferred save
      },
    },
  },
```

ALameLlama:
Closes #1233

## 📑 Description

Updates to use actively maintained okuuva repo, I also needed to update the way the callbacks work. These are currently broken since it's using autoformat_enabled instead of autoformat

### DIFF
--- a/lua/astrocommunity/editing-support/auto-save-nvim/init.lua
+++ b/lua/astrocommunity/editing-support/auto-save-nvim/init.lua
@@ -1,30 +1,40 @@
-return {
-  "Pocco81/auto-save.nvim",
-  event = { "User AstroFile", "InsertEnter" },
-  opts = {
-    callbacks = {
-      before_saving = function()
-        -- save global autoformat status
-        vim.g.OLD_AUTOFORMAT = vim.g.autoformat_enabled
+local group = vim.api.nvim_create_augroup("autosave", {})
 
-        vim.g.autoformat_enabled = false
-        vim.g.OLD_AUTOFORMAT_BUFFERS = {}
-        -- disable all manually enabled buffers
-        for _, bufnr in ipairs(vim.api.nvim_list_bufs()) do
-          if vim.b[bufnr].autoformat_enabled then
-            table.insert(vim.g.OLD_BUFFER_AUTOFORMATS, bufnr)
-            vim.b[bufnr].autoformat_enabled = false
-          end
-        end
-      end,
-      after_saving = function()
-        -- restore global autoformat status
-        vim.g.autoformat_enabled = vim.g.OLD_AUTOFORMAT
-        -- reenable all manually enabled buffers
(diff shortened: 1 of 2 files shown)


ALameLlama:
I'll still need to do a little more testing, but this should be a drop in replacement to the old callbacks and actually work since before they used the wrong vim options

lua/astrocommunity/editing-support/auto-save-nvim/init.lua
```diff
@@ -1,30 +1,40 @@
...
+  group = group,
+  callback = function()
+    -- restore global autoformat status
+    vim.g.autoformat = vim.g.OLD_AUTOFORMAT
+    -- re-enable all manually enabled buffers
+    for _, bufnr in ipairs(vim.g.OLD_AUTOFORMAT_BUFFERS or {}) do
+      vim.b[bufnr].autoformat = true
+    end
+  end,
+})
```
Uzaaft, line 34: this should be done using astrocore

Uzaaft: Also why was this moved outside the original plugin spec?

Uzaaft (dismissed):
see comments

Uzaaft (changes requested):
See comments

ALameLlama:
@Uzaaft Done 🫡.  Moved to use astro core autocmds and everything still seems to be working

ALameLlama:
I also might need to update the title of this PR to include that it's a breaking change.

the forked repo is a drop-in replacement if it's just the defaults but if people have extended it, it might break things

Uzaaft (approved):

PR #1240, opened 2024-10-11
docs(copilot-cmp): Add override example 

Ding-Fan:
I've tried many place to put the `opts`, only worked this way.


Uzaaft:
Our docs shouldnt compensate for the plugin docs.

PR #1241, opened 2024-10-13
feat(pack): add csharp support with `omnisharp`

ahmtsen:
## 📑 Description
This PR adds C# language support with `omnisharp` as LSP

### DIFF
--- a/lua/astrocommunity/pack/cs-omnisharp/README.md
+++ b/lua/astrocommunity/pack/cs-omnisharp/README.md
@@ -1,0 +1,12 @@
+# C# Language Pack
+
+**Requirements:** `dotnet` must be in your `PATH` and executable
+
+This plugin pack does the following:
+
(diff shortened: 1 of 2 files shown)


ahmtsen:
@mehalter when will this PR be merged?

ALameLlama:
@AstroNvim/astrocommunity-maintainers This looks fine to me, it's similar to the CS pack but the LSP is different, should this be a separate pack or is there a way to can make it option and only have 1 CS pack?

Uzaaft:
> @AstroNvim/astrocommunity-maintainers This looks fine to me, it's similar to the CS pack but the LSP is different, should this be a separate pack or is there a way to can make it option and only have 1 CS pack?

i think this is fine as is tbh

ALameLlama (approved):

ALameLlama:
Ship it

PR #1245, opened 2024-10-17
feat(nvim-lint): make autocmd events configurable

nicholasxjy:
## 📑 Description

make `nvim-lint` autocmd events(`"BufWritePost", "BufReadPost", "InsertLeave", "TextChanged"` now) can be configured from `opts.events`

### DIFF
--- a/lua/astrocommunity/lsp/nvim-lint/init.lua
+++ b/lua/astrocommunity/lsp/nvim-lint/init.lua
@@ -9,7 +9,7 @@ return {
   opts = {},
   config = function(_, opts)
     local lint, astrocore = require "lint", require "astrocore"
-
+    local autocmd_events = opts.events or { "BufWritePost", "BufReadPost", "InsertLeave", "TextChanged" }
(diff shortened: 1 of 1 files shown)


Uzaaft (commented):
opts.events doesnt seem to be a valid field for the options that should be passed to this plugin. Not sure if we should add it, so I defer to another maintainer. But this would open the floodgates in terms scope for astrocommunity.
Also, the current plugin spec should be updated to use astrocore.

mehalter:
Yeah we should avoid trying to modify the `opts`. I think what this should be doing is using the AstroCore autocommand creation feature which would make it configurable to the user

mehalter:
Okay, tested and it looks good to go! Here is how the user would configure the autocommand events:

```lua
return {
  "AstroNvim/astrocore",
  opts = function(_, opts)
    -- customize the autocommand events for nvim-lint
    opts.autocmds.auto_lint[1].event = { "BufWritePost" }
  end
}
```

mehalter (approved):
Thanks for this! It is a great idea to make this configurable

PR #1250, opened 2024-10-23
feat(neogit): clear completion sources when adding a git commit message

windowsrefund:
## 📑 Description

When entering a commit message, I noticed completion was suggesting from the buffer source which doesn't seem very valuable in that context. 

I'm looking for feedback on this PR. For example, maybe I should check if cmp-nvim is installed first (someone may be using blink).

### DIFF
--- a/lua/astrocommunity/git/neogit/init.lua
+++ b/lua/astrocommunity/git/neogit/init.lua
@@ -41,4 +41,8 @@ return {
       integrations = { telescope = utils.is_available "telescope.nvim" },
     })
   end,
+  config = function()
+    local cmp = require("cmp")
+    cmp.setup.filetype("NeogitCommitMessage", { sources = {} })
+  end,
 }


mehalter:
This probably isn't a good idea to keep in AstroCommunity. It will break other plugins that pointedly set up completion for the NeogitCommitMessage filetype. If you want better completion I would definitely recommend checking out the `cmp-git` plugin in AstroCommunity:

https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/completion/cmp-git/init.lua

If you prefer no completion then it is probably best to add this to your user configuration

mehalter:
Let me know what you think or if you want us to help put together a highly robust setup that can try to detect these things, but it might get very messy very quickly. I'll go ahead and close this but please feel free to respond and let me know if you want it reopened!

windowsrefund:
@mehalter Thank you for the feedback. Essentially, I just  don't see the value in having any completion available when entering a commit message and figured basing it on the `NeogitMessage` filetype to be the appropriate jumping off point to remove sources. 

commit

PR #1264, opened 2024-11-08
feat: adding snacks.nvim plugins

luxus:
adding snacks from our friend folke. i just copied the install instruction and modified it a little to make them fit into astrocommunity, i'm sure im missing something and collide with some keybindings

### DIFF
--- a/lua/astrocommunity/utility/snacks-nvim/init.lua
+++ b/lua/astrocommunity/utility/snacks-nvim/init.lua
@@ -1,21 +1,37 @@
 return {
   "folke/snacks.nvim",
-  priority = 1000,
+  priority = 10000,
   lazy = false,
-  ---@type snacks.Config
-  opts = {
-    bigfile = { enabled = false },
-    notifier = {
-      enabled = true,
-      timeout = 3000,
-    },
-    quickfile = { enabled = true },
-    statuscolumn = { enabled = true },
-    words = { enabled = true },
-    styles = {
-      notification = {
-        wo = { wrap = true }, -- Wrap notifications
+  ---@param opts snacks.Config
+  opts = function(_, opts)
+    local astrocore = require "astrocore"
+    return astrocore.extend_tbl(opts, {
+      bigfile = { enabled = not vim.tbl_get(astrocore.config, "autocmds", "large_buf_settings") },
+      notifier = {
+        enabled = not astrocore.is_available "nvim-notify",
+        timeout = 3000,
       },
+      quickfile = { enabled = true },
+      statuscolumn = { enabled = true },
+      words = { enabled = not astrocore.is_available "vim-illuminate" },
+      styles = {
+        notification = {
+          wo = { wrap = true }, -- Wrap notifications
+        },
+      },
+    } --[[ @type snacks.Config ]])
+  end,
+  specs = {
+    { "rcarriga/nvim-notify", enabled = false },
+    { "RRethy/vim-illuminate", enabled = false },
+    {
+      "rebelot/heirline.nvim",
+      optional = true,
+      opts = function(_, opts)
+        if vim.tbl_get(require("astrocore").plugin_opts "snacks.nvim", "statuscolumn", "enabled") then
+          opts.statuscolumn = false
+        end
+      end,
     },
   },
   dependencies = {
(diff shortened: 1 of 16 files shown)


lua/astrocommunity/utility/snacks-nvim/init.lua
```diff
@@ -0,0 +1,84 @@
...
+        maps.n["<leader>gb"] = { function() Snacks.git.blame_line() end, desc = "Git Blame Line" }
+        maps.n["<leader>gB"] = { function() Snacks.gitbrowse() end, desc = "Git Browse" }
+        maps.n["<leader>gf"] = { function() Snacks.lazygit.log_file() end, desc = "Lazygit Current File History" }
+        maps.n["<leader>gl"] = { function() Snacks.lazygit.log() end, desc = "Lazygit Log (cwd)" }
+        maps.n["<leader>cR"] = { function() Snacks.rename() end, desc = "Rename File" }
+        maps.n["<c-/>"] = { function() Snacks.terminal() end, desc = "Toggle Terminal" }
+        maps.n["<c-_>"] = { function() Snacks.terminal() end, desc = "which_key_ignore" }
+        maps.n["]]"] = { function() Snacks.words.jump(vim.v.count1) end, desc = "Next Reference" }
+        maps.n["[["] = { function() Snacks.words.jump(-vim.v.count1) end, desc = "Prev Reference" }
+        maps.n["<leader>N"] = {
```
Uzaaft, line 65: Does this override an existing binding.

luxus: yeah i need some time to convert the toggleterm stuff to snacks
https://github.com/AstroNvim/AstroNvim/blob/main/lua/astronvim/plugins/toggleterm.lua

luxus:
i think this News `<Leader>N` is just a PoC and not really useful? :D

Uzaaft:
> i think this News `<Leader>N` is just a PoC and not really useful? :D

Agreed.

lua/astrocommunity/utility/snacks-nvim/init.lua
```diff
@@ -0,0 +1,101 @@
...
+                spell = false,
+                wrap = false,
+                signcolumn = "yes",
+                statuscolumn = " ",
+                conceallevel = 3,
+              },
+            }
+          end,
+          desc = "Neovim News",
+        }
```
Uzaaft, line 97:
```suggestion
```

luxus:
should not toggleterm be fully replaced with that? @mehalter

mehalter:
Yeah I think it should, I think you mentioned above you were gonna do it 😂

mehalter:
It should probably be default override toggleterm but if the user manually re-enables it then it should go back to the defaults in AstroNvim

luxus:
if i wait long enough, hopefully it will fix by itself :-D

Uzaaft:
Ready for review?

mehalter:
There still needs to be work done to replace toggleterm stuff

luxus:
i checked snacks, dunno how to do a float without a cmd
not sure how to emulate <Leader>tf

RayJameson:
> i checked snacks, dunno how to do a float without a cmd not sure how to emulate `<Leader>tf`

I peeked into source code, this snippet works for me
```lua
        maps.n["<Leader>tf"] = {
          function() Snacks.terminal(nil, { win = { position = "float" } }) end,
          desc = "Toggle Terminal Float",
        }
```

lua/astrocommunity/utility/snacks-nvim/init.lua
```diff
@@ -0,0 +1,109 @@
...
+        if vim.fn.executable(gdu) == 1 then
+          maps.n["<Leader>tu"] = { function() Snacks.terminal { "gdu" } end, desc = "ToggleTerm gdu" }
+        end
+        maps.n["<Leader>gB"] = { function() Snacks.gitbrowse() end, desc = "Git Browse" }
+        maps.n["<Leader>gf"] = { function() Snacks.lazygit.log_file() end, desc = "Lazygit Current File History" }
+        maps.n["<Leader>gl"] = { function() Snacks.lazygit.log() end, desc = "Lazygit Log (cwd)" }
+        maps.n["<Leader>cR"] = { function() Snacks.rename() end, desc = "Rename File" }
+        maps.n["<Leader>th"] = { function() Snacks.terminal() end, desc = "Toggle Terminal" }
+        maps.n["<Leader>tf"] = { function() Snacks.terminal() end, desc = "Toggle Terminal Float" }
+        maps.n["<Leader>to"] = { function() Snacks.terminal "<Cmd>execute v:count ." end, desc = "Toggle Terminal" }
```
RayJameson, line 103: This seems incorrect

Uzaaft: Why/How?

RayJameson: Maybe I'm using it wrong, but if I press the keymap, then I get error, and if I understand the `Snacks.terminal` api correctly, use supposed to provide shell command, not vim command

lua/astrocommunity/utility/snacks-nvim/init.lua
```diff
@@ -0,0 +1,109 @@
+return {
+  "folke/snacks.nvim",
+  priority = 10000,
+  lazy = false,
+  ---@param opts snacks.Config
+  opts = function(_, opts)
+    local astrocore = require "astrocore"
+    return astrocore.extend_tbl(opts, {
```
RayJameson, line 8: I think those options would make snacks windows fall in line with other current float windows style
```suggestion
    return astrocore.extend_tbl(opts, {
    win = {
      border = "rounded",
      backdrop = false,
    },
```

lua/astrocommunity/utility/snacks-nvim/init.lua
luxus: > Maybe I'm using it wrong, but if I press the keymap, then I get error, and if I understand the `Snacks.terminal` api correctly, use supposed to provide shell command, not vim command

do you have a example, before i will search myself

liujoey:
Wish one day Snacks can be part of the core and replaces a ton of core plugins.

mehalter:
They don't really provide much and I'm not a fan of the structure personally. They also aren't super full featured and aren't particularly good replacements for the masses

luxus:
yeah i have mixed feelings too, i like the dashboard and the lazygit with nicer colors.
the issue is that many of the stuff needs configuration and people should configure it themself instead of a oppionated version in astrocommunity

mehalter:
Also lazygit nice colors had been implemented similarly into AstroUI

ALameLlama:
If we are splitting them into individual plugins, would adding a pack for snacks be handy to enable them all?

mehalter:
@ALameLlama yeah this is what I imagined would be the case

Uzaaft:
> I'd like to know what stage it's at now.

This PR is still in draft. There  are lots of plugins from snacks.nvim still missing.

acdcbyl:
> > I'd like to know what stage it's at now.
> 
> This PR is still in draft. There are lots of plugins from snacks.nvim still missing.

Thank you very much

Uzaaft:
since v5 is launched, this seems like its not needed. Wdyt? @luxus @mehalter

mehalter:
I think there are still some good stuff in here that are useful like if the user wants to easily opt into the `snacks.words` or `snacks.statuscol`

PR #1265, opened 2024-11-10
fix(auto-save-nvim): not restoring buffer settings

ALameLlama:
## 📑 Description
When testing this more, I noticed that the `table.insert` directly into `vim.g.OLD_AUTOFORMAT_BUFFERS` didn't seem to work, at the end of the for loop if you `table.inspect` it, it will be `{}`.

This means, that if you went into insert mode, did something, went to normal mode and then `:w` it wouldn't format even if you had it enabled before.


wert2all:
tested. work as expected.

Uzaaft (approved):

PR #1268, opened 2024-11-14
feat(blink-cmp): disable cmp-git

windowsrefund:
Ensure cmp-git is no longer enabled.

### DIFF
--- a/lua/astrocommunity/completion/blink-cmp/init.lua
+++ b/lua/astrocommunity/completion/blink-cmp/init.lua
@@ -93,6 +93,7 @@ return {
     -- disable built in completion plugins
     { "hrsh7th/nvim-cmp", enabled = false },
     { "rcarriga/cmp-dap", enabled = false },
+    { "petertriho/cmp-git", enabled = false },
     { "L3MON4D3/LuaSnip", enabled = false },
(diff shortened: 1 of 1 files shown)


Uzaaft:
@windowsrefund Do you have a reference or a link as to why this plugin needs to be disabled?

windowsrefund:
https://github.com/AstroNvim/astrocommunity/blob/aaaa844e45420cd7b5f11b7c399bee919513d1d5/lua/astrocommunity/completion/cmp-git/init.lua#L10

Uzaaft:
> https://github.com/AstroNvim/astrocommunity/blob/aaaa844e45420cd7b5f11b7c399bee919513d1d5/lua/astrocommunity/completion/cmp-git/init.lua#L10

Doesnt answer the questiion. Why does it have to be disabled? doesnt it work with the plugin?

windowsrefund:
How would it work if cmp itself is disabled?

Uzaaft:
> How would it work if cmp itself is disabled?

This one isn’t built in.

Uzaaft:
All the other plugins being disabled are built ins.

Uzaaft (approved):

PR #1275, opened 2024-11-23
fix(proto): replace deprecated `bufls` with `buf_ls`

imsakg:
Update language server references from `bufls` to `buf_ls` as `bufls` is deprecated and will be removed in lspconfig 0.2.1.

### DIFF
--- a/lua/astrocommunity/pack/proto/README.md
+++ b/lua/astrocommunity/pack/proto/README.md
@@ -3,5 +3,5 @@
 This plugin pack does the following:
 
 - Adds `proto` Treesitter parsers
-- Adds `bufls` language servers
+- Adds `buf_ls` language servers
 - Adds `buf` formatter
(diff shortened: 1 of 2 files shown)


Uzaaft (approved):

Galex88:
Hello !
I have an issue and I don't know where to put it. So here we go :

> [mason-lspconfig.nvim] Server "buf_ls" is not a valid entry in ensure_installed. Make sure to only provide lspconfig server names.

I am a newbie on nvim (used a lot of Clion and VSC before). Please help me !

Uzaaft:
@Galex88  Make sure your packages are up to date.

Galex88:
How can I do that ?
My nvim is at 0.10.2 (which is latest), and AstroNvim is at latest too (using "<Leader> p a" to update all Plugins and Mason packages)
image

Galex88:
Ok, after searching a bit, it looks like it's not an issue here sorry.
It looks like it is a [mason-lspconfig.nvim plugin issue](https://github.com/williamboman/mason-lspconfig.nvim/issues/488)

Uzaaft:
marking them as offtopic then

PR #1277, opened 2024-11-25
docs: Add pandoc theme to docs site

aslak01:
## 📑 Description

Added a  nicer theme (with dark mode support) to the pandoc markdown-html conversion, tweaked colors to match astrotheme.

Credit to https://github.com/jez/pandoc-markdown-css-theme

Preview: https://aslak01.github.io/astrocommunity/

## ℹ Additional Information

Moved bash script and resources to /docs/


Kamilcuk:
Hi. The page looks just incredibly stunning. Check review checklist - follow "proper conventional commit scoping", see https://www.conventionalcommits.org/en/v1.0.0/ . Your commits should start with `build(docs):` for example. The page visual is just 5 out of 5 stars.

Uzaaft:
> Hi. The page looks just incredibly stunning. Check review checklist - follow "proper conventional commit scoping", see [conventionalcommits.org/en/v1.0.0](https://www.conventionalcommits.org/en/v1.0.0/) . Your commits should start with `build(docs):` for example. The page visual is just 5 out of 5 stars.

Wrong. Its docs.

Uzaaft:
See https://github.com/AstroNvim/astrocommunity/commit/f5f344292e51252513b2316214d7c65038465814

Uzaaft:
Chrome works fine. Firefox as well.

Fuck Arc, less than 1% uses it. LGTM. 

@mehalter U merge pookie?

Uzaaft:
@aslak01 any reason why you added a shitton of fonts? Cant they be fetched at runtime?

ALameLlama:
Pretty sure we can remove most of these just by doing
```
@import url('https://fonts.googleapis.com/css2?family=Inter:wght@100;200;300;400;500;600;700;800;900&display=swap');
```

should be able to remove the font folder and font.css if we include that import in the theme.css

aslak01:
Yep, that'd work but I prefer to self-host fonts rather than depend on google fonts where I can personally. It's the same pattern that's in use on astronvim.com. 

If you look in the network tab, only two fonts files are loaded, the rest are just there in case we need them in the future, since Inter is an open source font and provides the entire set of weights and styles as a set.

aslak01:
@Uzaaft I can change it to fetch the font remotely if you want to make that a requirement but in my opinion "locally" hosted fonts are better.

Uzaaft:
@mehalter Ready to be merged

Uzaaft (approved):

Uzaaft:
The local fonts get a OK from me. Mehalter may have a differing opinion.

mehalter:
Hm I think maybe in this case we should do remote fonts. Mainly because we don't really need to distribute the font on every clone of the community  repo... Could also add it to the build process to pull them and include them in the deployed GitHub pages 🤷

aslak01:
OK, I've changed it to download the font files from the github repo at build time.

mehalter:
Looks great! @aslak01 could we also use JetBrains Mono for the code font to match the docs as well?

aslak01:
Sure, I can add that too.

aslak01:
Added @mehalter. Just wanted to plug where I got the previous font definition: https://modernfontstacks.com, pretty neat project to avoid downloading fonts.

mehalter:
Awesome! Yeah we are using fontsource for all of the fonts on AstroNvim.com and the docs

mehalter (approved):
Looks great! Thanks so much!

PR #1279, opened 2024-11-29
fix(cs-omnisharp): Fix bug in handler registration caused by wrong lsp name

### LINKED ISSUE #1278: Omnisharp doesn't work on M1 Mac
### Checklist

- [X] I have searched through the AstroNvim documentation
- [X] I have searched through the existing issues of this project
- [X] I have searched the existing issues of plugins related to this issue
- [X] I can replicate the bug with the minimal `repro.lua` provided below

### Neovim version (nvim -v)

v0.10.2 

### Operating system/version

macOS 15.0

### Terminal/GUI

iTerm2

### Describe the bug

Omnisharp simply does not work on the M1 Mac. Attempting to "go to definition" yields the error seen in the screencap. What's puzzling is that autocomplete suggestions seem to be present, but accepting a suggestion will yield the same error.

Screenshot 2024-11-26 at 2 47 01 PM

### Steps to Reproduce

1. Create directory
2. Add repro.lua to directory
3. Make a new dotnet project (`dotnet new mvc` via terminal)
4. Run `nvim -u repro.lua`
5. Open any cs file in the project
6. Attempt to "Find References", "Go to definition", or any LSP-supported action.

### Expected behavior

LSP should work.

### Screenshots

_No response_

### Additional Context

Looks like someone else has this exact problem on the Neovim repo: 
https://github.com/neovim/neovim/issues/30908

### Minimal configuration

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
end

-- bootstrap lazy
	-- stylua: ignore
	vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", "--branch=stable",
		lazypath })
end

-- install plugins
local plugins = {
	{ "AstroNvim/AstroNvim",      import = "astronvim.plugins" },
	{ "AstroNvim/astrocommunity", import = "astrocommunity.pack.mdx" },
        { "AstroNvim/astrocommunity", import = "astrocommunity.pack.cs-omnisharp" },
}
	root = root .. "/plugins",
})

-- bootstrap lazy
  -- stylua: ignore
end

-- install plugins
local plugins = {

}
  root = root .. "/plugins",
})

```

ahmtsen:
## 📑 Description
`omnisharp-extended-lsp.nvim` handlers are not registered correctly.

Language server name changed from `csharp_ls` to `omnisharp`.

Closes #1278


Uzaaft (approved):

gepbird:
AFAIK `csharp_ls` and omnisharp are 2 different projects. Does `omnisharp-extended-lsp.nvim` work with `csharp_ls`? Have you tested for example the go to definition function on a symbol that requires decompilation?

Uzaaft:
> ~AFAIK `csharp_ls` and `omnisharp-roslyn` are 2 different projects. Does `omnisharp-extended-lsp.nvim` work with `csharp_ls`? Have you tested for example the go to definition function on a symbol that requires decompilation?~
> 
> Oops I misread the changes, sorry. However I'm still surprised that this fix works. What about using custom commands instead of handlers?

Waiting on a reply on this before I merge.  @ahmtsen

ahmtsen:
> > ~AFAIK `csharp_ls` and `omnisharp-roslyn` are 2 different projects. Does `omnisharp-extended-lsp.nvim` work with `csharp_ls`? Have you tested for example the go to definition function on a symbol that requires decompilation?~
> > Oops I misread the changes, sorry. However I'm still surprised that this fix works. What about using custom commands instead of handlers?
> 
> Waiting on a reply on this before I merge. @ahmtsen

Custom handlers extend the functionality of Neovim's built-in LSP functions, such as `vim.lsp.buf.references()`. The reason for me to avoid custom commands provided by `omnisharp-extended-lsp.nvim` is the need to map those commands to specific keys, which can create inconsistencies.

For example, if we map `require('omnisharp_extended').lsp_references()` to the `gr` key when `omnisharp` lsp is attached, but the user has mapped `vim.lsp.buf.references()` to another key, such as `fr`, they would expect the `fr` key to trigger the `"go to references"` functionality. However, this won't work without additional setup.

By using custom handlers instead of custom commands, this pack extends the functionality of Neovim's built-in functions directly. This approach ensures consistency and avoids conflicts with user-defined key mappings.

I hope this explanation makes things clear.

gepbird:
@ahmtsen Yes, that make sense, thanks for the explanation.

In my config I was using handlers for a long time, then after a neovim nightly update I ran into this problem (https://github.com/neovim/neovim/issues/30908) and that's why I switched to custom commands which fixed it. And custom commands are marked as optimal while handlers are marked as not suboptimal (https://github.com/Hoffs/omnisharp-extended-lsp.nvim#how-to-use) due to many [similar issues](https://github.com/Hoffs/omnisharp-extended-lsp.nvim/issues?q=cursor+position+outside+of+buffer).

Thanks for pointing out that in a distro you can't just change default keybindings as people may have remapped that, that's a very good point! However if this issue comes back, maybe some hacky solution has to be implemented, like overriding `vim.lsp.buf.defition` and others instead of changing keybindings, but this sounds like a bad idea.

If you've properly tested this change, no objections from me, feel free to merge, I see this fixes an issue :) But I don't see how does it fix the 'Cursor position outside of buffer' issue.

(Also sorry for the noise, I came from the neovim issue and I'm not an AstroNvim user)

Uzaaft (approved):

PR #1287, opened 2024-12-13
fix(avante)!: Remap prefix to A

### LINKED ISSUE #1257: Avante bad keymappings

I've started to use Avante. It's really great. But I have some troubles with key bindings.

They start on [c] key that already used by vim motions [c]ange. E.g. I cannot use standard vim [cc] binding. I know, that I can use [S] instead of [cc], but it is occupied by another popular vim sneaky plugin.

FYI @aarnphm 

### Describe the new feature

I'd like to consider to change first letter or add leader key to all AvanteConflict* commands

### Additional context

https://github.com/AstroNvim/astrocommunity/tree/main/lua/astrocommunity/completion/avante-nvim

Uzaaft:
Closes #1257

### DIFF
--- a/lua/astrocommunity/completion/avante-nvim/init.lua
+++ b/lua/astrocommunity/completion/avante-nvim/init.lua
@@ -19,7 +19,7 @@ return {
       "AstroNvim/astrocore",
       opts = function(_, opts)
         local maps = assert(opts.mappings)
-        local prefix = "<Leader>a"
+        local prefix = "<Leader>A"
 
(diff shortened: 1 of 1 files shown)


Uzaaft:
R? @AstroNvim/astrocommunity-maintainers

mehalter:
I'll fix this

mehalter:
Also this doesn't fix the original issue at all, opened a new PR that resolves the issue and applies this suggestion as well.

PR #1288, opened 2024-12-13
fix(avante-nvim)!: improve Avante keymaps to not conflict with with changing text

### LINKED ISSUE #1257: Avante bad keymappings

mehalter:
Closes #1257 

## 📑 Description

This does a few things to fix issues in the Avante plugin:

- Makes it so the conflict resolution keymaps do not conflict with core vim changing text bindings
- Fixes building for windows
- Changes prefix to `<Leader>A` and makes Avante ask binding more ergonomic
- Adds pretty icon for Avante menu title

### DIFF
--- a/lua/astrocommunity/completion/avante-nvim/init.lua
+++ b/lua/astrocommunity/completion/avante-nvim/init.lua
@@ -1,6 +1,7 @@
 return {
   "yetone/avante.nvim",
-  build = ":AvanteBuild",
+  build = vim.fn.has "win32" == 1 and "powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource false"
+    or "make",
   cmd = {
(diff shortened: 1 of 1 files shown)


Uzaaft (approved):
LGTM  pookie :shipit:

PR #1289, opened 2024-12-14
feat(session): add vim-workspace plugin and session directory

Kamilcuk:
## 📑 Description

Add vim-workspace plugin and session directory.

If a new `session` directory is not ok... I guess `workflow` or `project` dir? I wasn't able to decide, please guide me.

### DIFF
--- a/lua/astrocommunity/session/vim-workspace/init.lua
+++ b/lua/astrocommunity/session/vim-workspace/init.lua
@@ -1,0 +1,30 @@
+return {
+  { "resession.nvim", enabled = false },
+  {
+    "thaerkh/vim-workspace",
+    enabled = true,
+    priority = 10000,
+    lazy = false,
+    init = function()
+      vim.cmd [[
+      let g:workspace_autosave_ignore = ['gitcommit', "neo-tree", "nerdtree", "qf", "tagbar"]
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/session/vim-workspace/init.lua
```diff
@@ -0,0 +1,30 @@
...
+          \ if exists(":NERDTreeClose") | execute 'NERDTreeClose' | endif
+	    let g:workspace_create_new_tabs = 0
+	    " Because a bug, these two populate search / history, just disable them.
+	    let g:workspace_autosave_untrailtabs = 0
+	    let g:workspace_autosave_untrailspaces = 0
+	    let g:workspace_nocompatible = 0
+	    let g:workspace_session_disable_on_args = 1
+	    " https://github.com/thaerkh/vim-workspace/issues/11
+	    set sessionoptions-=blank
+      ]]
```
mehalter, line 27: These should be using `AstroCore`

mehalter: I refactored this PR so that the configuration is actually configurable by the user rather than slapping it into a black box

mehalter (approved):
Thanks for the contribution!

mehalter (approved):

PR #1290, opened 2024-12-14
feat(project): Add in linear-nvim

Uzaaft:
Unsure about the deps.

They seem to be required. 

Added in keymaps that will conflict with overseer.


ALameLlama (approved):
LGTM

PR #1293, opened 2024-12-18
fix(leap-nvim): Fix cursor invisible bug on nvim 0.10+

### LINKED ISSUE #1219: Leap cursor visibility workaround makes cursor invisible on nvim 0.10+
https://github.com/AstroNvim/astrocommunity/blob/f4a19c410fdcf778d44a8b9ece55e529df7a0274/lua/astrocommunity/motion/leap-nvim/init.lua#L9-L25

These should be conditional on nvim < 0.10

ALameLlama:
Closes #1219 

## 📑 Description

The workaround breaks on 0.10+ so I've made it only apply when the minor version is less than 10.

### DIFF
--- a/lua/astrocommunity/motion/leap-nvim/init.lua
+++ b/lua/astrocommunity/motion/leap-nvim/init.lua
@@ -5,26 +5,28 @@ return {
     {
       "AstroNvim/astrocore",
       opts = {
-        autocmds = {
-          leap_cursor = { -- https://github.com/ggandor/leap.nvim/issues/70#issuecomment-1521177534
-            {
-              event = "User",
-              pattern = "LeapEnter",
-              callback = function()
-                vim.cmd.hi("Cursor", "blend=100")
-                vim.opt.guicursor:append { "a:Cursor/lCursor" }
-              end,
-            },
-            {
-              event = "User",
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/motion/leap-nvim/init.lua
```diff
@@ -5,26 +5,28 @@ return {
...
+                  event = "User",
+                  pattern = "LeapLeave",
+                  callback = function()
+                    vim.cmd.hi("Cursor", "blend=0")
+                    vim.opt.guicursor:remove { "a:Cursor/lCursor" }
+                  end,
+                },
+              },
+            }
+          or {},
```
Uzaaft, line 29: Is this vim.version() api available on 0.9?

ALameLlama: Was added in 0.9 by the looks of it https://neovim.io/doc/user/news-0.9.html

> Added [vim.version](https://neovim.io/doc/user/lua.html#vim.version) for parsing and comparing version strings conforming to the semver specification.

Uzaaft (approved):

Uzaaft (approved):

PR #1297, opened 2024-12-20
feat(blink.cmp): load lazy when enter cmdline mode.

chaozwn:
sometimes i enter neovim and use cmdline directly. i can't get cmdline hints.


mehalter:
This has been resolved in the latest commit

PR #1299, opened 2024-12-23
fix: fix vue template complete cursor problem.

chaozwn:


mehalter:
This is resolved on `main`

PR #1301, opened 2024-12-25
fix(mini-surround): chanage prefix from gz to gs

windowsrefund:
Feel free to reject if `gz` was or still is desired but it strikes me as an oversight.


Uzaaft:
letting this one stay open for now until someone else raise this concern.

PR #1306, opened 2025-01-04
docs: fix relative font path

aslak01:
Forgot about the leading path in github pages `/astrocommunity/` so we need to add a dot to the font preload to avoid flash of unstyled text.

Capture d’écran 2025-01-04 à 14 07 17


mehalter (approved):
Thanks so much!

PR #1313, opened 2025-01-09
feat(snippet): add `mini.snippets`

mehalter:
## 📑 Description

This adds [`mini.snippets`](https://github.com/echasnovski/mini.snippets) as an option for configuring snippets with a fairly fully featured configuration and integration setup.

### DIFF
--- a/lua/astrocommunity/snippet/mini-snippets/init.lua
+++ b/lua/astrocommunity/snippet/mini-snippets/init.lua
@@ -1,0 +1,104 @@
+return {
+  "echasnovski/mini.snippets",
+  dependencies = "rafamadriz/friendly-snippets",
+  opts_extend = { "snippets" },
+  opts = function(_, opts)
+    local mini_snippets = require "mini.snippets"
+    local gen_loader = mini_snippets.gen_loader
+    opts.snippets = vim.list_extend({
+      -- global snippets file
+      gen_loader.from_file(vim.fn.stdpath "config" .. "/snippets/global.json"),
+      -- load language specific snippets from runtime
+      gen_loader.from_lang(),
+      -- load global, project local snippets
+      gen_loader.from_file ".vscode/project.code-snippets",
+      -- load language specific, project local snippets
+      function(context)
+        local rel_path = ".vscode/" .. context.lang .. ".code-snippets"
+        return vim.fn.filereadable(rel_path) == 1 and mini_snippets.read_file(rel_path)
+      end,
+    }, opts.snippets or {})
+  end,
+  specs = {
+    {
+      "AstroNvim/astrocore",
+      ---@param opts AstroCoreOpts
+      opts = function(_, opts)
+        local snippet_mode_change
+        local stop_all_sessions = function()
+          local mini_snippets = require "mini.snippets"
+          while mini_snippets.session.get(false) do
+            mini_snippets.session.stop()
+          end
+          snippet_mode_change = nil
+        end
+        opts.autocmds.mini_snippets_stop = {
+          {
+            event = "User",
+            pattern = "MiniSnippetsSessionStart",
+            desc = "Set up snippet stopping when going to normal mode autocmd",
+            callback = function()
+              if not snippet_mode_change then
+                snippet_mode_change = vim.api.nvim_create_autocmd(
+                  "ModeChanged",
+                  { pattern = "*:n", once = true, callback = stop_all_sessions }
+                )
+              end
+            end,
+          },
+        }
+      end,
+    },
+    {
+      "hrsh7th/nvim-cmp",
+      optional = true,
+      dependencies = { { "abeldekat/cmp-mini-snippets", lazy = true } },
+      opts = function(_, opts)
+        local mini_snippets, cmp = require "mini.snippets", require "cmp"
+
+        if not opts.snippet then opts.snippet = {} end
+        opts.snippet.expand = function(args)
+          (mini_snippets.config.expand.insert or mini_snippets.default_insert) { body = args.body }
+        end
+
+        if not opts.sources then opts.sources = {} end
+        table.insert(opts.sources, { name = "mini_snippets", priority = 750 })
+
+        local function has_words_before()
+          local line, col = (unpack or table.unpack)(vim.api.nvim_win_get_cursor(0))
+          return col ~= 0 and vim.api.nvim_buf_get_lines(0, line - 1, line, true)[1]:sub(col, col):match "%s" == nil
+        end
+        local function is_visible(CMP) return CMP.core.view:visible() or vim.fn.pumvisible() == 1 end
+
+        if not opts.mappings then opts.mappings = {} end
+        opts.mapping["<Tab>"] = cmp.mapping(function(fallback)
+          if is_visible(cmp) then
+            cmp.select_next_item()
+          elseif vim.api.nvim_get_mode().mode ~= "c" and mini_snippets.session.get(false) then
+            mini_snippets.session.jump "next"
+          elseif has_words_before() then
+            cmp.complete()
+          else
+            fallback()
+          end
+        end)
+        opts.mapping["<S-Tab>"] = cmp.mapping(function(fallback)
+          if is_visible(cmp) then
+            cmp.select_prev_item()
+          elseif vim.api.nvim_get_mode().mode ~= "c" and mini_snippets.session.get(false) then
+            mini_snippets.session.jump "prev"
+          else
+            fallback()
+          end
+        end)
+      end,
+    },
+    {
+      "Saghen/blink.cmp",
+      dependencies = "echasnovski/mini.snippets",
+      optional = true,
+      opts = { snippets = { preset = "mini_snippets" } },
+    },
+    { "L3MON4D3/LuaSnip", enabled = false },
+  },
+}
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/snippet/mini-snippets/init.lua
```diff
@@ -0,0 +1,104 @@
+return {
+  "echasnovski/mini.snippets",
+  dependencies = "rafamadriz/friendly-snippets",
+  opts_extend = { "snippets" },
+  opts = function(_, opts)
+    local mini_snippets = require "mini.snippets"
+    local gen_loader = mini_snippets.gen_loader
+    opts.snippets = vim.list_extend(opts.snippets or {}, {
```
echasnovski, line 8: I *think* the order here should be different: first "default" locations and then the ones from the `opts`. This way snippets read from the user supplied locations will be preferred if there are snippets with the same prefix ("last one wins"). This should allow users to override snippets from "default" locations.

lua/astrocommunity/snippet/mini-snippets/init.lua
```diff
@@ -0,0 +1,104 @@
...
+    {
+      "hrsh7th/nvim-cmp",
+      optional = true,
+      dependencies = { { "abeldekat/cmp-mini-snippets", lazy = true } },
+      opts = function(_, opts)
+        local mini_snippets, cmp = require "mini.snippets", require "cmp"
+
+        if not opts.snippet then opts.snippet = {} end
+        opts.snippet.expand = function(args)
+          (mini_snippets.config.expand.insert or mini_snippets.default_insert) { body = args.body }
```
echasnovski, line 61: Please, use `_G.MiniSnippets` at least for `config.expand.insert` part, as it will allow more flexibility for adjusting config.

lua/astrocommunity/snippet/mini-snippets/init.lua
```diff
@@ -0,0 +1,104 @@
...
+          if is_visible(cmp) then
+            cmp.select_next_item()
+          elseif vim.api.nvim_get_mode().mode ~= "c" and mini_snippets.session.get(false) then
+            mini_snippets.session.jump "next"
+          elseif has_words_before() then
+            cmp.complete()
+          else
+            fallback()
+          end
+        end, { "i", "s" })
```
echasnovski, line 84: I am not sure if 'nvim-cmp' itself requires a `<Tab>` and `<S-Tab>` mappings in Select mode, but 'mini.snippets' does not.

lua/astrocommunity/snippet/mini-snippets/init.lua
mehalter: Ah we can flip that! The way these are imported in AstroNvim should make this a non-issue because all community specs are imported before the user's code so all of them should be added after

lua/astrocommunity/snippet/mini-snippets/init.lua
mehalter: I have tested this and it works well :) thanks so much for the review!

mehalter: We can go ahead and remove the `s` part

lua/astrocommunity/snippet/mini-snippets/init.lua
mehalter: Just tested and there is no performance difference nor problems with flexibility between using `_G.MiniSnippets` vs `require("mini.snippets")` with lua caching and the fact that the tables are the same memory location. I'll leave it as is since that makes LSP/linters more happy and promote better "type safety". Thanks again for the recommendation!

echasnovski: This is not really about performance or making linter happy.

Yes, if no funny business is being done, `require('mini.snippets')` and `_G.MiniSnippets` should point to exactly the same table. The same is true for their `config` field, thus modifying it inside either one will also affect the other. *But* if something to happen and `_G.MiniSnippets` or `_G.MiniSnippets.config` is replaced with its copy, then it would stop working.

'mini.nvim' recommends using `_G.MiniXxx` global variable for "interactive" config adjustment (among other things) mostly because it is more intuitive than adjusting `require('mini.xxx')`, aligns well with buffer-local configs (`vim.b.minixxx_config`), and serves as an extra check that module was properly and purposefully set up.

mehalter: I'm not sure if the internal use of `_G.MiniSnippets` works the way you think it does. In the internals of `mini.snippets` you are never referencing `_G.MiniSnippets`, you are referencing your local variable `MiniSnippets` which overshadows the global variable throughout the module. So say the user does do really funny business and just deepcopies all of `_G.MiniSnippets`. You are correct that `require("mini.snippets")` will not reflect these changes, but neither will those changes be made available to the internals of `mini.snippets` either.

Steps to verify:

1. Add a line to `print(global.test or "NOT FOUND")` here where you calculate the resolution of the config which is used throughout `mini.snippets`: https://github.com/echasnovski/mini.nvim/blob/890db26e37bb28d8a95f05570dd0d9fa757b698b/lua/mini/snippets.lua#L1558
2. In the configuration, right after calling `mini.snippets` setup, add the following:
    ```lua
        _G.MiniSnippets = vim.tbl_deep_extend("force", vim.deepcopy(_G.MiniSnippets), {
          config = { test = "FOUND" },
        })
    ```
3. Start the editor and use `mini.snippets`. Notice that it keeps printing `NOT FOUND`, but if you run `:lua =_G.MiniSnippets.config.test` you will get `FOUND`

Also I'm not sure if I agree with it being in alignment with the `vim.b` variable. I think it would make a lot more sense from a parallelism standpoint to have `vim.g.mini_snippets_config` and `vim.b.mini_snippets_config` instead and then in the source code access that one single global variable than trying to just export the entire module globally.

mehalter: outside of fully copying the entire module and extending the entire table, any funny business done within the table, say deep copying `_G.MiniSnippets.config` will be reflected because `_G.MiniSnippets` is referenced by name and not by value.

mehalter: in the case of how `lazy.nvim` works with lazy loading and stuff, the `require("mini.snippets")` is a way for the module to be loaded if it's not loaded yet and run through the `setup` function which helps guarantee at the time of needing it that it is available. Rather than hoping that the user did it previously

mehalter: Another way to test this without modifying the `mini.snippets` source code is to just do:

```lua
require("mini.snippets").setup(opts)
_G.MiniSnippets = vim.deepcopy(_G.MiniSnippets)
_G.MiniSnippets.config = nil
```

and see that `mini.snippets` still works completely. But if you add `require("mini.snippets").config = nil`, then you run into problems because the code does assume the configuration exists but isn't actually checking `_G.MiniSnippets.config`.

echasnovski: > I'm not sure if the internal use of `_G.MiniSnippets` works the way you think it does. In the internals of `mini.snippets` you are never referencing `_G.MiniSnippets`, you are referencing your local variable `MiniSnippets` which overshadows the global variable throughout the module. So say the user does do really funny business and just deepcopies all of `_G.MiniSnippets`. You are correct that `require("mini.snippets")` will not reflect these changes, but neither will those changes be made available to the internals of `mini.snippets` either.

Nothing in my previous two comments says that `_G.MiniSnippets` is *used* inside 'mini.snippets'. You are correct that global variable is not directly used inside 'mini.snippets', no need for multiple ways showing that. Yes, this also means that 'mini.snippets' internal usage of config will not respect any "funny business" way of adjusting config. I just thought that the same need not apply here as well. Well, your PR, your rules.

> Also I'm not sure if I agree with it being in alignment with the `vim.b` variable. I think it would make a lot more sense from a parallelism standpoint to have `vim.g.mini_snippets_config` and `vim.b.mini_snippets_config` instead and then in the source code access that one single global variable than trying to just export the entire module globally.

Reasons for general suggestion about `_G.MiniSnippets` vs `require('mini.snippets')` are outlined in [last paragraph of this comment](https://github.com/AstroNvim/astrocommunity/pull/1313#discussion_r1910120877). 

> in the case of how `lazy.nvim` works with lazy loading and stuff, the `require("mini.snippets")` is a way for the module to be loaded if it's not loaded yet and run through the `setup` function which helps guarantee at the time of needing it that it is available. Rather than hoping that the user did it previously

I forgot that this is for a distro which is locked in on specific plugin manager and uses standalone 'mini.snippets' repo. The general advice is formulated for 'mini.nvim' users for whom checking `_G.MiniSnippets` presence is crucial, as `require('mini.snippets')` will always succeed but they might have not opt-in for its usage. Opting in via this PR's 'mini.snippets' integration/recipe/spec should be enough for that.

Uzaaft (approved):
:shipit:

lua/astrocommunity/snippet/mini-snippets/init.lua
Uzaaft: feel free to open a follow up PR on any improvements you guys come up with. I have tested this the last day or so and its working fine as of now.

PR #1314, opened 2025-01-12
fix(linear-nvim)!: Change leader from `<Leader>m` to `<Leader>L`

Uzaaft:
This makes more sense semantically

### DIFF
--- a/lua/astrocommunity/project/linear-nvim/init.lua
+++ b/lua/astrocommunity/project/linear-nvim/init.lua
@@ -11,7 +11,7 @@ return {
         local prefix = "<Leader>L"
 
         maps.n[prefix] = { desc = "Linear" }
-        maps.n[prefix .. "m"] = {
+        maps.n[prefix .. "l"] = {
           function() require("linear-nvim").show_assigned_issues() end,
           desc = "Show assigned issues",
         }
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/project/linear-nvim/init.lua
```diff
@@ -8,7 +8,7 @@ return {
       "AstroNvim/astrocore",
       opts = function(_, opts)
         local maps = opts.mappings
-        local prefix = "<Leader>m"
+        local prefix = "<Leader>L"
 
         maps.n[prefix] = { desc = "Linear" }
         maps.n[prefix .. "m"] = {
```
mehalter, line 14:
```suggestion
        maps.n[prefix .. "l"] = {
```

mehalter: I think the mapping descriptions for the other bindings also should be updated like "create issue" and "show issue"

mehalter (changes requested):
This is a great idea, I left a couple comments to go ahead and do to help clean this spec up

lua/astrocommunity/project/linear-nvim/init.lua
Uzaaft: Why? Isnt the current mapping descriptions(based  on the original desc from the plugin readme) good enough?

mehalter (approved):
Looks great!

Issue #883, opened 2024-04-14
nix pack updates to nixfmt and nixd


luxus:
### Describe the new feature

nix pack could some updates
https://github.com/nix-community/nixd looks like the newest and fastest lsp right now.
but it's unstable because of a major rewrite?
i changed it locally via https://github.com/nix-community/nixd/blob/main/nixd/docs/editor-setup.md
alejandra should be replaced by  nixfmt if it makes into 2405 release. (waiting for https://github.com/NixOS/nixfmt/issues/153)
it works on unstable with `nixpkgs.nixfmt-rfc-style`

both is bleeding edge
thats my code right now. 

```lua
return {
  {
    "nvim-treesitter/nvim-treesitter",
    optional = true,
    opts = function(_, opts)
      if opts.ensure_installed ~= "all" then
        opts.ensure_installed = require("astrocore").list_insert_unique(opts.ensure_installed, { "nix" })
      end
    end,
  },
  {
    "AstroNvim/astrolsp",
    -- we must use the function override because table merging
    -- does not play nicely with list-like tables
    ---@param opts AstroLSPOpts
    opts = function(plugin, opts)
      -- safely extend the servers list
      opts.servers = opts.servers or {}
      vim.list_extend(opts.servers, {
        "nixd",
        -- add more servers as needed...
      })
    end,
  },
  {
    "nvimtools/none-ls.nvim",
    optional = true,
    opts = function(_, opts)
      local nls = require "null-ls"
      if type(opts.sources) == "table" then
        vim.list_extend(opts.sources, {
          nls.builtins.code_actions.statix,
          nls.builtins.formatting.nixfmt,
          nls.builtins.diagnostics.deadnix,
        })
      end
    end,
  },
  {
    "stevearc/conform.nvim",
    optional = true,
    opts = {
      formatters_by_ft = {
        nix = { "nixfmt" },
      },
    },
  },
  -- {
  --   "mfussenegger/nvim-lint",
  --   optional = true,
  --   opts = {
  --     linters_by_ft = {
  --       nix = { "statix", "nix" },
  --     },
  --   },
  -- },
}
```

and i guess it makes sense to add deadnix to nvim-lint too
https://github.com/bondzula/nvim/blob/7a58fd58857a9a121df9703d4ad132b9bb583e56/lua/plugins/nvim-lint.lua#L9-L41

Uzaaft:
@luxus  Feel free to add nixd now that version 2 is released.

[...]

PR #1324, opened 2025-01-19
fix(nix): drop none-ls formatter

Mic92:
nixd already provides formatting builtin:

https://github.com/nix-community/nixd/blob/2c25600cb9c91bc06fe8676c044814dc30435274/nixd/docs/configuration.md?plain=1#L62

### DIFF
--- a/lua/astrocommunity/pack/nix/README.md
+++ b/lua/astrocommunity/pack/nix/README.md
@@ -3,7 +3,6 @@
 Requires the following in your `PATH`
 
 - [nixd](https://github.com/nix-community/nixd)
-- [alejandra](https://github.com/kamadorueda/alejandra)
 - [deadnix](https://github.com/astro/deadnix)
 - [statix](https://github.com/nerdypepper/statix)
 
@@ -12,6 +11,5 @@ This plugin pack does the following:
 - Adds `nix` Treesitter parsers
 - Adds `nixd` language server
 - Adds the following `null-ls` sources:
-  - [alejandra](https://github.com/kamadorueda/alejandra)
   - [deadnix](https://github.com/astro/deadnix)
   - [statix](https://github.com/nerdypepper/statix)
--- a/lua/astrocommunity/pack/nix/init.lua
+++ b/lua/astrocommunity/pack/nix/init.lua
@@ -15,7 +15,6 @@ return {
       local builtins = require("null-ls").builtins
       opts.sources = require("astrocore").list_insert_unique(opts.sources, {
         builtins.code_actions.statix,
-        builtins.formatting.alejandra,
         builtins.diagnostics.deadnix,
       })
     end,
@@ -24,15 +23,6 @@ return {
     "AstroNvim/astrolsp",
     optional = true,
     opts = function(_, opts) opts.servers = require("astrocore").list_insert_unique(opts.servers, { "nixd" }) end,
-  },
-  {
-    "stevearc/conform.nvim",
-    optional = true,
-    opts = {
-      formatters_by_ft = {
-        nix = { "alejandra" },
-      },
-    },
   },
   {
     "mfussenegger/nvim-lint",


Uzaaft:
We are currently avoiding nixfmt until it becoms stable. See the tracking issue: #883

Mic92:
We have nixfmt in 24.11, so it is stable? Also this doesn't replace alejandra it let nixd handle the configuration. People can still configure nixd to use alejandra. It just sets the default.

Uzaaft:
> We have nixfmt in 24.11, so it is stable? Also this doesn't replace alejandra it let nixd handle the configuration. People can still configure nixd to use alejandra. It just sets the default.

My PR sets the formatter as alejandra.As for the stability of nixfmt, check their github repo. They call it not yet stable. :)

Mic92:
> > We have nixfmt in 24.11, so it is stable? Also this doesn't replace alejandra it let nixd handle the configuration. People can still configure nixd to use alejandra. It just sets the default.
> 
> My PR sets the formatter as alejandra.As for the stability of nixfmt, check their github repo. They call it not yet stable. :)

If nixd is active this doesn't work because nixd do the formatting not non-ls. So this is a no-op.

luxus:
thanks for the pr. whats will be the default formatter after the merge?

luxus (approved):
looks fine, if people want to use something else then the default nixfmt, they can configure it

but the readme needs a update

Uzaaft (approved):

Zebradil:
How to change the formatter now? I tried adding the following to the astrolsp config:

```lua
return {
  "AstroNvim/astrolsp",
  ---@type AstroLSPOpts
  opts = {
    ---@diagnostic disable: missing-fields
    config = {
      nixd = {
        formatting = { command = { "alejandra" } },
      },
    },
  },
}
```

But there seem to be no effect. Trying to format a nix file leads to the following error message in the lsp log (`nixfmt` isn't installed):

```
... "1583502: --> reply:textDocument/formatting(10) formatting nixfmt command exited with 65280, error: 3\n"
```

Though, it could be because nixd configuration is broken in general, as with vanilla astronvim configuration (from the template repo), with the nix pack enabled, I see the following error in the lsp log after opening a nix file:

```
... workspace/configuration: parse error expected object
```

Uzaaft:
> How to change the formatter now? I tried adding the following to the astrolsp config:
> 
> ```lua
> return {
>   "AstroNvim/astrolsp",
>   ---@type AstroLSPOpts
>   opts = {
>     ---@diagnostic disable: missing-fields
>     config = {
>       nixd = {
>         formatting = { command = { "alejandra" } },
>       },
>     },
>   },
> }
> ```
> 
> But there seem to be no effect. Trying to format a nix file leads to the following error message in the lsp log (`nixfmt` isn't installed):
> 
> ```
> ... "1583502: --> reply:textDocument/formatting(10) formatting nixfmt command exited with 65280, error: 3\n"
> ```
> 
> Though, it could be because nixd configuration is broken in general, as with vanilla astronvim configuration (from the template repo), with the nix pack enabled, I see the following error in the lsp log after opening a nix file:
> 
> ```
> ... workspace/configuration: parse error expected object
> ```

You can set it in your flake.nix

Zebradil:
According to https://github.com/nix-community/nixd/issues/496, nixd doesn't support getting formatter configuration from flake.nix

luxus:
what you wanna do? setting a different formatter? because nixd does have build in formating configuration and sets it to nix-fmt, if you want that, then just use the code from above

Mic92:
https://github.com/nix-community/nixd/blob/main/nixd/docs/configuration.md#configuration-overview nixd has a configuration file.

Uzaaft:
> what you wanna do? setting a different formatter? because nixd does have build in formating configuration and sets it to nix-fmt, if you want that, then just use the code from above

You are right. I gotta double check the behaviour in nixd some other day and come back to you.

Zebradil:
> what you wanna do? setting a different formatter?

Yes, I wanted to use alejandra instead of the default nixfmt.

As @Mic92 pointed out, nixd can be configured, including the formatter. The problem is that my configuration doesn't have any effect.

It feels like there is a problem in the initial configuration process somewhere between astrolsp and nixd, but I don't possess skills to debug this further. Probably, this is not the right place for me for reporting this issue, I just thought that maybe some of you happen to use non-default formatter with nixd and knows how to configure it in astronvim. Where should I create an issue (astrolsp → nixd) instead?

PR #1325, opened 2025-01-20
feat(nix): Specify alejandra as formatter in the lsp instead of nixfmt

Uzaaft:
Superseeds #1324


lua/astrocommunity/pack/nix/init.lua
```diff
@@ -23,16 +23,24 @@ return {
...
-    opts = function(_, opts) opts.servers = require("astrocore").list_insert_unique(opts.servers, { "nixd" }) end,
-  },
-  {
-    "stevearc/conform.nvim",
-    optional = true,
-    opts = {
-      formatters_by_ft = {
-        nix = { "alejandra" },
-      },
-    },
```
Uzaaft, line 35: Unsure about how we should do this. I removed conform since that made the most sense, but I am open to adding ternary logic for astrolsp to disable formatting if conform is defined in the config

PR #1326, opened 2025-01-20
docs: add extention 'autolink_bare_uris'

fronterior:
## 📑 Description

Some plugin documentation doesn't work because the links are written in just text instead of link markdown syntax.
Pandoc has an option to automatically convert this, so adding it to the site build process removes the need for contributors to check it.

I checked by running the pages_create.sh script on my local environment and only link text without braces is converted.

## ℹ Additional Information

https://pandoc.org/MANUAL.html#extension-autolink_bare_uris

### DIFF
--- a/docs/pages_create.sh
+++ b/docs/pages_create.sh
@@ -46,7 +46,7 @@ EOF
 # Use pandoc to convert from markdown to html.
 docker run -i --rm --mount type=bind,source="$PWD",target="$PWD",readonly --workdir "$PWD" \
   pandoc/core:3.5 \
-  --from markdown \
+  --from markdown+autolink_bare_uris \
   --standalone \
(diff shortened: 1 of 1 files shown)


Uzaaft:
I vote against this. For a upcoming prototype of the next website for astrocommunity I don’t use pandoc. Leaving the decision up to you @mehalter

fronterior:
However, if accept this PR, we can now add links to unlinked text on the astrocommunity website. I was uncomfortable with this, so I wrote a PR. However, as @Uzaaft says, if a new site is being created soon, I don't think it's necessary.

Uzaaft:
Actually this isnt needed. The rule in the repo is that we have fenced urls. That should stay.

PR #1336, opened 2025-01-26
feat(sql): disable sqls formatter and add sqlfluff

### LINKED ISSUE #1323: Disable the default SQL formatter

Should the default formatting be disabled for the [SQL pack](https://github.com/AstroNvim/astrocommunity/tree/main/lua/astrocommunity/pack/sql)?

It uses `sqls`, which hasn't been maintained for a while, and there is an ongoing issue that causes some SQL statements to be broken by the formatter: https://github.com/sqls-server/sqls/issues/149

### Describe the new feature

Either update the readme and add an example of how to disable the formatting, leaving a warning note that this issue occurs, or disable the formatting by default in the SQL pack.

An alternative would be to find a replacement formatter.

### Additional context

In the comments of the `sqls`, someone mentioned https://github.com/nrempel/sleek for formatting, but I couldn't find it in Mason.

azdanov:
## 📑 Description

This should disable the sqls lsp formatter and use sqlfluff instead. Might as well use sqlfluff diagnostics also, I see them beneficial.

This is the issue with sqls at the moment: https://github.com/sqls-server/sqls/issues/149

## ℹ Additional Information
This needs thorough testing, since I'm not sure if all works as expected and I lack knowledge to verify on my own due to being new to neovim ecosystem. Especially different combinations of null-ls/conform/nvim-lint/lspconfig community packs.

I disabled cwd in conform due to default settings requiring a config file to be present for the formatter to work. I didn't notice this behavior in null-ls.

### DIFF
--- a/lua/astrocommunity/pack/sql/init.lua
+++ b/lua/astrocommunity/pack/sql/init.lua
@@ -26,15 +26,56 @@ return {
     end,
   },
   {
-    "WhoIsSethDaniel/mason-tool-installer.nvim",
+    "jay-babu/mason-null-ls.nvim",
     optional = true,
     opts = function(_, opts)
-      opts.ensure_installed = require("astrocore").list_insert_unique(opts.ensure_installed, { "sqls" })
+      opts.ensure_installed = require("astrocore").list_insert_unique(opts.ensure_installed, { "sqlfluff" })
+      opts.handlers = {
+        sqlfluff = function()
+          local null_ls = require "null-ls"
+          null_ls.register(null_ls.builtins.diagnostics.sqlfluff.with {
+            extra_args = { "--dialect", "ansi" },
+          })
+          null_ls.register(null_ls.builtins.formatting.sqlfluff.with {
+            extra_args = { "--dialect", "ansi" },
+          })
+        end,
+      }
     end,
   },
   {
+    "WhoIsSethDaniel/mason-tool-installer.nvim",
+    optional = true,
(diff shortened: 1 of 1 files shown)


azdanov:
Seems to be working. One thing is that the default conform timeout in community pack is 500ms, which might not be enough for sqlfluf :man_shrugging:

lua/astrocommunity/pack/sql/init.lua
```diff
@@ -45,7 +86,11 @@ return {
               desc = "Load sqls.nvim with sqls",
               callback = function(args)
                 local client = assert(vim.lsp.get_client_by_id(args.data.client_id))
-                if client.name == "sqls" then require("sqls").on_attach(client, args.buf) end
+                if client.name == "sqls" then
+                  -- Disable formatting due to bugs: https://github.com/sqls-server/sqls/issues/149
+                  client.server_capabilities.documentFormattingProvider = false
```
mehalter, line 91: should this be done here or should this be done in AstroLSP?

azdanov: I couldn't figure out how to do it in LSP 😄

mehalter (approved):
LGTM!

lua/astrocommunity/pack/sql/init.lua
mehalter: I resolved this

mehalter (approved):

PR #1333, opened 2025-01-25
feat(copilotchat-nvim): select a picker based on installed one

azdanov:
## 📑 Description

This removes mandatory dependency on telescope and selects the picker based on which one is installed instead with a default to the telescope.

## ℹ Additional Information
Can see different integrations here:
https://github.com/CopilotC-Nvim/CopilotChat.nvim/tree/main/lua/CopilotChat/integrations

PR #1337, opened 2025-01-26
refactor(copilotchat-nvim): improve picker selection logic and error handling

tku137:
This is an attempt to further improve on https://github.com/AstroNvim/astrocommunity/pull/1333.

First, a separate function to check for the currently active pickers name or `nil` if it has no CopilotChat.integrations module. Since I did not find a better solution to check if snacks.nvim picker is actually used as the picker (solely checking for snacks.nvim availability does not mean that it's picker is actually configured and used, contrary to telescope or fzf-lua), I tried finding something to find out, which is up for debate. Second, the `create_mapping()` function now checks for availability and compatibility of the active picker with CopilotChat integrations and gracefully warn users if it's not.

### DIFF
--- a/lua/astrocommunity/editing-support/copilotchat-nvim/init.lua
+++ b/lua/astrocommunity/editing-support/copilotchat-nvim/init.lua
@@ -69,16 +69,20 @@ return {
 
         -- Determine the currently active picker
         local function get_active_picker()
-          -- Snacks integration check
-          -- this is tricky because we can't just assume that if snacks.nvim is
-          -- installed, it's being used as the picker, too
-          local snacks = require "snacks.picker"
-          if snacks.config and snacks.config.ui_select then return "snacks" end
-
-          -- Check other CopilotChat compatibel pickers
+          -- Check CopilotChat compatibel pickers
           -- this is also a mapping between picker module names and their
           -- counterparts in the CopilotChat integrations
-          if require("astrocore").is_available "fzf-lua" then
+          -- snacks.picker is tricky because we can't just assume that if snacks.nvim is
+          -- installed, it's being used as the picker, too
+          if
+            require("astrocore").is_available "snacks.nvim"
+            and (function()
+              local snacks = require "snacks.picker"
+              return snacks.config and snacks.config.ui_select
+            end)()
+          then
+            return "snacks"
+          elseif require("astrocore").is_available "fzf-lua" then
             return "fzflua"
           elseif require("astrocore").is_available "telescope.nvim" then
             return "telescope"
--- a/lua/astrocommunity/editing-support/copilotchat-nvim/init.lua
+++ b/lua/astrocommunity/editing-support/copilotchat-nvim/init.lua
@@ -67,17 +67,62 @@ return {
           desc = "Load Chat",
         }
 
+        -- Determine the currently active picker
+        local function get_active_picker()
+          -- Snacks integration check
+          -- this is tricky because we can't just assume that if snacks.nvim is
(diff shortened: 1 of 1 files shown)


azdanov:
Looks good, but I don't understand most of it (low vim skills on my side) :smile:

lua/astrocommunity/editing-support/copilotchat-nvim/init.lua
```diff
@@ -67,17 +67,62 @@ return {
...
+          if snacks.config and snacks.config.ui_select then return "snacks" end
+
+          -- Check other CopilotChat compatibel pickers
+          -- this is also a mapping between picker module names and their
+          -- counterparts in the CopilotChat integrations
+          if require("astrocore").is_available "fzf-lua" then
+            return "fzflua"
+          elseif require("astrocore").is_available "telescope.nvim" then
+            return "telescope"
+          end
```
Uzaaft, line 85: Why not use the same logic in the whole function? The snack require would fail at runtime if I am thinking right.

tku137: I'm not sure. If snacks is not installed, line 75 should be `nil`, right? My reasoning for different logic is that even though one has installed snacks, it does not mean that its picker is configured, too. Other then fzf-lua and telescope. I am not sure either how we could solve getting the actually configured picker. LazyVim has `LazyVim.pick.want()` which is exactly the name of the picker the user has chosen. Does AstroNVim has something similar?

mehalter: no if snacks is not installed then it will error

lua/astrocommunity/editing-support/copilotchat-nvim/init.lua
```diff
@@ -69,16 +69,20 @@ return {
...
           -- counterparts in the CopilotChat integrations
-          if require("astrocore").is_available "fzf-lua" then
+          -- snacks.picker is tricky because we can't just assume that if snacks.nvim is
+          -- installed, it's being used as the picker, too
+          if
+            require("astrocore").is_available "snacks.nvim"
+            and (function()
+              local snacks = require "snacks.picker"
+              return snacks.config and snacks.config.ui_select
+            end)()
```
tku137, line 82: What about this? Utilizing astrocore and snacks config...

mehalter: The easiest way to find out is to test it :D give it a shot. It will still error

mehalter: ah I see you did the is_available first, that is a very wacky way of doing this 😂 fine by me tho

tku137: Yeah I agree ... but I don't see a way to be sure about snacks.picker, tbh. Astronvim needs a `LazyVim.pick.want()` 😁

I did a quick test and it worked, of course. Altough it was just a quick with and without snacks-picker from astrocommunity and firing up the CopilotChat prompt picker. Not using AstroNVim these days 😇

mehalter: haha I want the core code of AstroNvim to be a lot less abstracted than that. LazyVim is moving towards a very highly engineered configuration which is great but it makes the codebase highly complex and a lot less like how someone would configure Neovim. There are pros and cons of each but it's good to have differences between distributions ❤️

mehalter (approved):
Looks great! Thanks so much! :shipit:

PR #1339, opened 2025-01-28
feat(fzf-lua): directly go to definition when only a single result

azdanov:
## 📑 Description
This will add behavior similar to telescope and snacks-picker, where if there is a single result then it directly goes there instead of opening the picker window.

## ℹ Additional Information
https://github.com/ibhagwan/fzf-lua/wiki#lsp-jump-to-location-for-single-result


Uzaaft (approved):
:shipit:

PR #1345, opened 2025-01-30
fix(markview-nvim): adapted config to consider breaking changes in version v25.1.

gitkeeper:
## 📑 Description

Adapted config to consider breaking changes introduced in `markview.nvim` v25.1.

## ℹ Additional Information

This fix works for `markview.nvim` v25.1 and later.


ALameLlama (approved):
Thanks!

PR #1346, opened 2025-01-31
fix(tailwindcss): add support for v4 by using tailwind.css rooter scan

### LINKED ISSUE #1328: Tailwind V4 support

Tailwindcss V4 dropped [few hours ago](https://tailwindcss.com/blog/tailwindcss-v4). And with the new major release, the way configuration are handled completely, moving from a js file to a css-firsr configuration. Which means the tailwindcss pack doesn't work anymore since it doesn't detect any JS configuration.

### Describe the new feature

The tailwindcss pack needs some tweaking in order to make it work again.

### Additional context

_No response_

azdanov:
## 📑 Description

This should hopefully allow working with tailwind v4

## ℹ Additional Information
Proper solution would require scanning all `.css` files for `@import "tailwindcss";` which I imagine would be quite intensive.

### DIFF
--- a/lua/astrocommunity/pack/tailwindcss/README.md
+++ b/lua/astrocommunity/pack/tailwindcss/README.md
@@ -4,5 +4,11 @@ This plugin pack does the following:
 
 - Adds `css` Treesitter parser
 - Adds `tailwindcss` and `cssls` language servers
+  - tailwindcss server activates when it detects any of these files:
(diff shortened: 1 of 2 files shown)


azdanov:
I don't think this is a good approach. See https://github.com/AstroNvim/astrocommunity/issues/1328

PR #1350, opened 2025-02-01
feat(copilot-lua-cmp): add blink.cmp support

azdanov:
## 📑 Description
This should hopefully add similar copilot support to blink as it was for cmp. Not sure if it's 1 to 1, since I'm not familiar with either of them :see_no_evil: 

## ℹ Additional Information
I've checked https://cmp.saghen.dev/configuration/keymap.html#example and copied `Tab` from main community pack.

### DIFF
--- a/lua/astrocommunity/completion/copilot-lua-cmp/init.lua
+++ b/lua/astrocommunity/completion/copilot-lua-cmp/init.lua
@@ -87,13 +87,13 @@ return {
             end,
             "fallback",
           },
-          ["<C-x>"] = { copilot_action "next" },
-          ["<C-z>"] = { copilot_action "prev" },
-          ["<C-right>"] = { copilot_action "accept_word" },
-          ["<C-l>"] = { copilot_action "accept_word" },
-          ["<C-down>"] = { copilot_action "accept_line" },
-          ["<C-j>"] = { copilot_action "accept_line", "select_next", "fallback" },
-          ["<C-c>"] = { copilot_action "dismiss" },
+          ["<C-X>"] = { copilot_action "next" },
+          ["<C-Z>"] = { copilot_action "prev" },
+          ["<C-Right>"] = { copilot_action "accept_word" },
+          ["<C-L>"] = { copilot_action "accept_word" },
+          ["<C-Down>"] = { copilot_action "accept_line" },
+          ["<C-J>"] = { copilot_action "accept_line", "select_next", "fallback" },
+          ["<C-C>"] = { copilot_action "dismiss" },
         }
 
         for k, v in pairs(keymap) do
--- a/lua/astrocommunity/completion/copilot-lua-cmp/init.lua
+++ b/lua/astrocommunity/completion/copilot-lua-cmp/init.lua
@@ -87,13 +87,13 @@ return {
             end,
             "fallback",
           },
-          ["<C-X>"] = { copilot_action "next" },
-          ["<C-Z>"] = { copilot_action "prev" },
-          ["<C-Right>"] = { copilot_action "accept_word" },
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/completion/copilot-lua-cmp/init.lua
```diff
@@ -56,6 +58,48 @@ return {
...
+          ["<Tab>"] = {
+            copilot_action "accept",
+            "select_next",
+            "snippet_forward",
+            function(cmp)
+              if has_words_before() or vim.api.nvim_get_mode().mode == "c" then return cmp.show() end
+            end,
+            "fallback",
+          },
+          ["<C-X>"] = { copilot_action "next" },
```
ALameLlama, line 90: should this be x instead of X to match what we do in cmp?

 I haven't swapped to using blink yet so I can't test if this functions the same but I'll see if I can test this more sometime next week

azdanov: Maybe. I don't know :smile: 

I copied from here: https://github.com/AstroNvim/astrocommunity/blob/5263d85a1dcbc8fd153ec385c1346c2ad9d95056/lua/astrocommunity/completion/blink-cmp/init.lua#L90-L96

ALameLlama: I think those are fine, those would be actions on the blink cmp pop up, but the binds here
https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/completion/copilot-lua-cmp/init.lua#L31-L57

are for the copilot suggestion and should probably be the same between the two cmp's

ALameLlama: other wise instead of `["<C-X>"] = { copilot_action "next" },` it should be `["<C-N>"] = { copilot_action "next" },` to match the blink mapping

mehalter (changes requested):
Could you update the README? I'll merge this in after

lua/astrocommunity/completion/copilot-lua-cmp/init.lua
mehalter: ah see there is an unresolved conversation. I'll let @ALameLlama figure out when this is ready 💪🏼

mehalter:
@azdanov you should revert the lowercase stuff. They should be uppercase.We should normalize all mappings on official keycodes.

lua/astrocommunity/completion/copilot-lua-cmp/init.lua
```diff
@@ -56,6 +58,48 @@ return {
...
+            end,
+            "fallback",
+          },
+          ["<C-x>"] = { copilot_action "next" },
+          ["<C-z>"] = { copilot_action "prev" },
+          ["<C-right>"] = { copilot_action "accept_word" },
+          ["<C-l>"] = { copilot_action "accept_word" },
+          ["<C-down>"] = { copilot_action "accept_line" },
+          ["<C-j>"] = { copilot_action "accept_line", "select_next", "fallback" },
+          ["<C-c>"] = { copilot_action "dismiss" },
```
mehalter, line 96:
```suggestion
          ["<C-X>"] = { copilot_action "next" },
          ["<C-Z>"] = { copilot_action "prev" },
          ["<C-Right>"] = { copilot_action "accept_word" },
          ["<C-L>"] = { copilot_action "accept_word" },
          ["<C-Down>"] = { copilot_action "accept_line" },
          ["<C-J>"] = { copilot_action "accept_line", "select_next", "fallback" },
          ["<C-C>"] = { copilot_action "dismiss" },
```

mehalter (approved):
LGTM :shipit:

Issue #1269, opened 2024-11-15
astrocommunity.pack.chezmoi - telescope mapping doesn't work


RayJameson:
### Checklist

- [X] I have searched through the AstroNvim documentation
- [X] I have searched through the existing issues of this project
- [X] I have searched the existing issues of plugins related to this issue
- [X] I can replicate the bug with the minimal `repro.lua` provided below

### Neovim version (nvim -v)

NVIM v0.10.2 

### Operating system/version

macOS 15.1

### Terminal/GUI

wezterm/neovide

### Describe the bug

Mapping for telescope in chezmoi pack doesn't work, I assume mapping for fzf-lua doesn't work either.
I would try to fix it but it's some complicated lazy magic there

### Steps to Reproduce

1. add `astrocommunity.pack.chezmoi` to `community.lua`
2. sync plugins
3. press `<Leader>f.`

### Expected behavior

Telescope picker should open

### Minimal configuration

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "runtime", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  -- stylua: ignore
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", "--branch=stable", lazypath })
end
vim.opt.rtp:prepend(vim.env.LAZY or lazypath)

-- install plugins
local plugins = {
  { "AstroNvim/AstroNvim", import = "astronvim.plugins" },
  { "AstroNvim/astrocommunity" },
  { import = "astrocommunity.pack.chezmoi" },

  -- add any other plugins/customizations here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

-- add anything else here (autocommands, vim.filetype, etc.)
```

RayJameson:
> I've updated, the repro.lua to install the chezmoi pack but I didn't have any issues trying up telescope with `<leader>ff`
> 
> 
> 
> is there anything else missing from the repro to reproduce the issue?

do you mean `<Leader>f.` ?

PR #1362, opened 2025-02-19
feat(chezmoi): Fix `specs` loading and enable watch option

Cretezy:
`fzf-lua` was defined outside of `specs`, causing none of the `specs` to load. 

Fixes https://github.com/AstroNvim/astrocommunity/issues/1269

Also enables auto-watching with notifications on apply.

### DIFF
--- a/lua/astrocommunity/pack/chezmoi/init.lua
+++ b/lua/astrocommunity/pack/chezmoi/init.lua
@@ -21,16 +21,13 @@ return {
     "xvzc/chezmoi.nvim",
     opts = {
       edit = {
-        watch = false,
+        watch = true,
         force = false,
       },
       notification = {
         on_open = true,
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/pack/chezmoi/init.lua
```diff
@@ -21,16 +21,13 @@ return {
     "xvzc/chezmoi.nvim",
     opts = {
       edit = {
-        watch = false,
+        watch = true,
```
Uzaaft, line 24: Why the change of this setting?

Uzaaft: Fixed the PR title to reflect this change.

Uzaaft (approved):

PR #1367, opened 2025-02-23
fix(noice-nvim): Use repo name for heirline plugin

Sundhar22:


luxus:
whats the reason for the change? is there are case when it is not working?

Uzaaft (approved):

Sundhar22:
> whats the reason for the change? is there are case when it is not working?

yes when I try to use Lualine it causes some error so I changed this and checked it worked perfectly

PR #1370, opened 2025-02-24
feat(editing-support): Add `quick-scope`

axelcalixte:
## 📑 Description

This pr adds the unblevable/quick-scope vim plugin to astrocommunity.

It offers a visual enhancement to f and t motions. I quite like it as I find it less intrusive than plugins like sneak or flash and just improves on a default vim workflow.

## ℹ Additional Information

I configured it to my liking to add a bit of value to its addition to astrocommunity.

### DIFF
--- a/lua/astrocommunity/editing-support/quick-scope/init.lua
+++ b/lua/astrocommunity/editing-support/quick-scope/init.lua
@@ -1,0 +1,15 @@
+return {
+  {
+    "unblevable/quick-scope",
+    keys = { "f", "F", "t", "T" },
+    dependencies = {
+      "AstroNvim/astrocore",
+      opts = function(_, opts)
+        if not opts.options then opts.options = {} end
+        opts.options.g["qs_highlight_on_keys"] = { "f", "F", "t", "T" }
+        if not opts.mappings then opts.mappings = {} end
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/editing-support/quick-scope/init.lua
Uzaaft, line 1: Please use astrocore to set the keys as well.

Uzaaft (changes requested):

lua/astrocommunity/editing-support/quick-scope/init.lua
axelcalixte: hum
I don't think I'm setting any mappings here
I just have the keys list to lazy load the plugin. The plugin could also work with `lazy = false` only.

Should I use `on_keys` table from AstroCoreConfig ?

mehalter: I think this one shouldn't be set through AstroCore because you don't actually define the mappings. I believe the plugin does. I can check more specifically later

axelcalixte: I modified the commit to include the plugin's mapping to the UI section.

PR #1374, opened 2025-02-28
feat(completion): add copilot-vim-cmp

azdanov:
## 📑 Description
This will add the official GitHub Copilot vim plugin (from tpope) and cmp support. The main reason is because copilot.vim gets updated immediately to any Copilot changes.

## ℹ Additional Information
There might be breaking changes or any other slowdowns in the copilot-lua similar to https://github.com/zbirenbaum/copilot.lua/pull/368 and it would be nice to have an alternative that's official and immediately updated by GitHub itself.


Uzaaft (changes requested):
@azdanov i think we should make it into two packs.
Copilot-vim and then a pack thats copilot-vim + cmp.
Thus people who prefer just basic copilot.vim can use it as is.

Uzaaft (approved):

PR #1378, opened 2025-02-28
refactor(completion): update AI plugins to use the new `astrocommunit…

mehalter:
…y.recipes.ai` boilerplate

## 📑 Description

This is an exploration of using the new AI recipes in AstroCommunity to improve the maintenance and implementation of AI plugins with completion. This should be tested, improved, and iterated on before merging. Someone could figure out a nice way to have the user switch between inline suggestions vs completion based suggestions using the `vim.g.ai_inline` variable that the user can set in AstroCore


azdanov:
A bit difficult to imagine how to make it work across so many different plugins, since I lack experience with many of them. :thinking: 

My main ones are just `github/copilot.vim` and `CopilotC-Nvim/CopilotChat.nvim`.

PR #1382, opened 2025-03-02
feat(docs): add auto scrollable TOC

azdanov:
## 📑 Description
This should make the docs website a little easier to use, since the TOC will be scrolling alongside the main content. And it should also scroll the TOC when a URL with an ID is opened, e.g., `https://astronvim.github.io/astrocommunity/#bigfile-nvim`.

[Screencast_20250302_181703.webm](https://github.com/user-attachments/assets/533feec6-9f8c-4dae-a3cf-34a3e6748fad)

## ℹ Additional Information
Might need some more testing, but seems to be working alright.

### DIFF
--- a/docs/assets/template.html5
+++ b/docs/assets/template.html5
@@ -104,6 +104,45 @@
           });
         });
       })();
+      document.addEventListener('DOMContentLoaded', function () {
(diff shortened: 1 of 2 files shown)


Uzaaft:
This is fucking sexy!

Uzaaft (approved):

PR #1385, opened 2025-03-02
feat(motion): add mini-jump plugin

Std-Enigma:
## 📑 Description

Add [Mini.Jump](https://github.com/echasnovski/mini.jump) plugin

### DIFF
--- a/lua/astrocommunity/motion/mini-jump/init.lua
+++ b/lua/astrocommunity/motion/mini-jump/init.lua
@@ -1,13 +1,5 @@
 return {
   "echasnovski/mini.jump",
   event = "User AstroFile",
-  opts = {
-    delay = {
-      -- Delay between jump and highlighting all possible jumps
-      highlight = 250,
-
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/motion/mini-jump/init.lua
```diff
@@ -0,0 +1,13 @@
...
+  event = "User AstroFile",
+  opts = {
+    delay = {
+      -- Delay between jump and highlighting all possible jumps
+      highlight = 250,
+
+      -- Delay between jump and automatic stop if idle (no jump is done)
+      idle_stop = 1000,
+    },
+  },
```
Uzaaft, line 12:
```suggestion
  opts = {},
```

Uzaaft: I dont see any reason to change the defaults for now. lets wait on feedback from community before we add in some defaults.

Uzaaft (approved):

PR #1388, opened 2025-03-13
feat(code-runner): Add mf-runner.nvim

burgr033:
## 📑 Description

This PR introduces [mf-runner.nvim](https://github.com/burgr033/mf-runner.nvim). A lightweight code runner plugin that uses good old Makefiles as a base.

### DIFF
--- a/lua/astrocommunity/code-runner/mf-runner-nvim/README.md
+++ b/lua/astrocommunity/code-runner/mf-runner-nvim/README.md
@@ -1,5 +1,5 @@
 # mf-runner.nvim
 
-Makefile task runner
+code runner based off of Makefiles because it's the easiest and one of the most common formats.
 
 **Repository:** <https://github.com/burgr033/mf-runner.nvim>
--- a/lua/astrocommunity/code-runner/mf-runner-nvim/README.md
+++ b/lua/astrocommunity/code-runner/mf-runner-nvim/README.md
@@ -1,0 +1,5 @@
+# mf-runner.nvim
+
+Makefile task runner
+
+**Repository:** <https://github.com/burgr033/mf-runner.nvim>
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/code-runner/mf-runner-nvim/README.md
```diff
@@ -0,0 +1,5 @@
+# mf-runner.nvim
+
+Makefile task runner
```
Uzaaft, line 3:
```suggestion
code runner based off of Makefiles because it's the easiest and one of the most common formats.
```

lua/astrocommunity/code-runner/mf-runner-nvim/init.lua
```diff
@@ -0,0 +1,19 @@
+return {
+  "burgr033/mf-runner.nvim",
+  cmd = { "MFROpen", "MFRRun", "MFREdit" },
+  dependencies = {
+    { "folke/snacks.nvim" },
```
Uzaaft, line 5: Since you are the author: You should probably make the plugin picker indepedent. @burgr033

Uzaaft (approved):

lua/astrocommunity/code-runner/mf-runner-nvim/init.lua
burgr033: thanks for the input! Switched from telescope and toggleterm to snacks the other day. should i just let the user choose  between the two, or what would be your suggested approach? @Uzaaft

Uzaaft: Well often, plugins just use vim.ui.* instead of requiring a plugin. And then its up to the user. Snacks.nvim overrides vim.ui.select

PR #1394, opened 2025-03-19
fix(docs): use correct favicon

azdanov:
## 📑 Description
Somehow I messed up copying the SVG, resulting in incorrect closing tag. This time should be good. I copied from https://astronvim.com/logo/astronvim.svg

## ℹ Additional Information
I noticed this because the SVG favicon was rendering fine in Chrome/Safari, but in Firefox it didn't work.


mehalter:
Is there a reason we are copying it here and not just referencing the external location of the logo? That way if it updates we don't need to worry about it? 🤔

azdanov:
Redundancy I guess to avoiding any network issues. I can replace with with direct links if you'd like. I don't know which way is better :)

mehalter:
Yeah idk, network issues will make it pretty hard to access the website itself 🤡

mehalter:
I don't have a strong preference

mehalter (approved):

azdanov:
> Yeah idk, network issues will make it pretty hard to access the website itself 🤡

Yeah true. I'm so used to working around weird infra problems at work that I see problems everywhere :smile:

PR #1400, opened 2025-03-24
feat(avante-nvim): integrate Avante with Mcphub

Yorizel:
- Add ravitemer/mcphub.nvim as optional plugin dependency
- Add additional cmd


Uzaaft:
This should be done the other way around.

Add mcphub as a plugin into astrocommunity, and then use the optional syntax there to integrate with the other plugins.

Yorizel:
ok, got it

PR #1408, opened 2025-03-29
feat!: astronvim v5 compatibility

Uzaaft:
TODO list:
- [x] Drop support for neovim 0.9
- [x] Alpha-nvim -> snacks.nvim
- [x] dressing.nvim -> snacks.nvim
- [x] indent-blankline.nvim	-> snacks.nvim
- [x] lspkind.nvim -> mini.icons
- [x] mini.bufremove -> snacks.nvim
- [x] nvim-cmp -> blink.cmp
- [x] nvim-colorizer.lua -> nvim-highlight-colors
- [x] nvim-notify -> snacks.nvim
- [x] nvim-web-devicons -> mini.icons
- [x] telescope.nvim -> snacks.nvim
- [x] mason-tool-installer integration 
	From release notes:
	> AstroNvim is now relying on mason-tool-installer.nvim rather than mason-lspconfig/mason-nvim-dap/mason-null-ls. This makes configuration of installed packages easier rather than managing several ensure_installed tables and also adds support for installing packages automatically that are not part of any of those usecase specific plugins. mason-tool-installer.nvim also allows the user to specify specific package versions and other options. Check out the [mason-tool-installer.nvim Documentation](https://github.com/WhoIsSethDaniel/mason-tool-installer.nvim) for details on it’s configuration.
For migration, anywhere in your configuration where you are configuring ensure_installed for mason-lspconfig/mason-nvim-dap/mason-null-ls, you should move those to an ensure_installed table in mason-tool-installer.nvim. You must also change the name of the package where necessary to properly reflect the name of the package in Mason. For instance, if you have lua_ls specified for mason-lspconfig, this would be lua-language-server in mason-tool-installer.nvim.

### DIFF
--- a/lua/astrocommunity/pack/haskell/init.lua
+++ b/lua/astrocommunity/pack/haskell/init.lua
@@ -76,7 +76,6 @@ return {
     "mrcjkb/haskell-tools.nvim",
     ft = haskell_ft,
     dependencies = {
-      { "nvim-lua/plenary.nvim", optional = true },
       { "nvim-telescope/telescope.nvim", optional = true },
       { "mfussenegger/nvim-dap", optional = true },
       {
--- a/lua/astrocommunity/pack/haskell/init.lua
+++ b/lua/astrocommunity/pack/haskell/init.lua
@@ -4,7 +4,7 @@ local is_available = function(plugin)
 end
 local haskell_ft = { "haskell", "lhaskell", "cabal", "cabalproject" }
 
-local pack = {
+return {
   { import = "astrocommunity.pack.yaml" }, -- stack.yaml
   { import = "astrocommunity.pack.json" }, -- hls.json
   { import = "astrocommunity.test.neotest" }, -- neotest-haskell
@@ -72,16 +72,11 @@ local pack = {
       table.insert(opts.adapters, require "neotest-haskell"(require("astrocore").plugin_opts "neotest-haskell"))
     end,
   },
-}
-
-if vim.fn.has "nvim-0.10" == 1 then
-  -- haskell-tools v4 supports neovim v0.10+
-  table.insert(pack, {
+  {
     "mrcjkb/haskell-tools.nvim",
     ft = haskell_ft,
     dependencies = {
-      -- vim.fn.has >= nvim 0.9 removes plenary dependency
-      { "nvim-lua/plenary.nvim", optional = vim.fn.has "nvim-0.9" == 1 },
+      { "nvim-lua/plenary.nvim", optional = true },
       { "nvim-telescope/telescope.nvim", optional = true },
       { "mfussenegger/nvim-dap", optional = true },
       {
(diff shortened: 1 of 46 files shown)


azdanov:
Should also figure out something about astrocommunity.lsp.nvim-lsp-file-operations, since in 0.11 it's built-in.
E.g. here: https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/pack/typescript/init.lua#L77

Uzaaft:
@azdanov sick! Didn’t know that, you got a link to the docs for that(on the phone rn)

azdanov:
I think this is the commit that enabled it: https://github.com/AstroNvim/AstroNvim/commit/870ddb205a145a1d48bbc13ec448bccd9ee29881

And this is the plugin, I guess it's not part of 0.11, but something in AstroNvim: https://github.com/AstroNvim/astrolsp/blob/main/lua/astrolsp/file_operations.lua

Uzaaft:
Ah got it, now I follow. Not sure if there is anything to be done? We can just leave the plugin there as is, and add in a notice about it being built in.

azdanov:
I was wondering if it's possible to add a condition check or something so it doesn't get imported if the user has these defaults turned on? https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/lsp/nvim-lsp-file-operations/init.lua

lua/astrocommunity/startup/fsplash-nvim/init.lua
```diff
@@ -28,7 +28,9 @@ return {
     },
   },
   specs = {
-    { "goolord/alpha-nvim", optional = true, enabled = false },
```
mehalter, line 31: This should be left

lua/astrocommunity/startup/mini-starter/init.lua
```diff
@@ -3,7 +3,9 @@ return {
   version = "*", -- stable version
   config = function() require("mini.starter").setup() end,
   specs = {
-    { "goolord/alpha-nvim", optional = true, enabled = false },
```
mehalter, line 6: This should be left

lua/astrocommunity/pack/nvchad-ui/init.lua
```diff
@@ -99,7 +99,12 @@ return {
     -- Disable unnecessary plugins
     { import = "astrocommunity.recipes.disable-tabline" },
     { "rebelot/heirline.nvim", opts = { statusline = false } },
-    { "goolord/alpha-nvim", enabled = false },
```
mehalter, line 102: This should be left

lua/astrocommunity/pack/haskell/init.lua
```diff
@@ -72,16 +72,11 @@ local pack = {
...
-if vim.fn.has "nvim-0.10" == 1 then
-  -- haskell-tools v4 supports neovim v0.10+
-  table.insert(pack, {
+  {
     "mrcjkb/haskell-tools.nvim",
     ft = haskell_ft,
     dependencies = {
-      -- vim.fn.has >= nvim 0.9 removes plenary dependency
-      { "nvim-lua/plenary.nvim", optional = vim.fn.has "nvim-0.9" == 1 },
+      { "nvim-lua/plenary.nvim", optional = true },
```
mehalter, line 79:
```suggestion
```

this should be removed

mehalter (changes requested):
a few review comments

mehalter:
Okay, I did most of the checklist and cleaned up some errors. Someone will need to go through all the mentions of `telescope` and `nvim-cmp` and make sure we add necessary setup for `snacks.picker` and `blink.cmp` where  it makes sense. It's important to note that we _shouldn't_ remove the already existing `nvim-cmp` and `telescope.nvim` specs because users might opt into adding those plugins. We should just also add in configuration for `blink.cmp` and `snacks.picker` as well with the `optional = true` tag.

Also where it makes sense, if we have forgotten to add the `optional = true` tag to `nvim-cmp` and `telescope.nvim` in places it should be added. Important to note that that _doesn't_ mean to add `optional = true` to all of the mentions of both plugins because some plugins have hard requirements for these dependencies and therefore they are _not_ optional.

But overall, this is very close! Let me know when this is done and I can do a review and do the merge!

mehalter:
I also think if most of the packs are done, we can just go ahead and merge this in and people using the plugins in AstroCommunity can provide fixes as they come up and are necessary. It's not the job of the maintainers of AstroCommunity to maintain the specs themselves and their validity. It's purely to manage the PRs and make sure when people want to contribute it doesn't have malicious code and gets merged in in a timely fashion

azdanov:
By the way, since this is a breaking PR. Is there an issue with `C-x` mappings? https://github.com/search?q=repo%3AAstroNvim%2Fastrocommunity+C-x+language%3ALua&type=code&l=Lua

I remember something about neovim 0.11 having some built-in autocomplete? Not sure, though.

mehalter:
There is a core vim completion mode prefix `<C-x>` but that goes back a long time we'll before  neovim and not new in 0.11

azdanov:
Okay, well, whoever uses those packs, if they have a problem with `C-x` can create an issue later :slightly_smiling_face:

mehalter:
I believe this PR is ready to go. Just need some thumbs up by someone else and we can get this merged in.

PR #1410, opened 2025-03-31
feat(picker-lsp-mappings): add `grr` and `gri` to Snacks LSP mappings

acarl005:
## 📑 Description

There are a couple of LSP-related operations that Snacks includes which we should include in these mappings: searching references and implementations. I've added the standard `grr` and `gri` here and mapped to those in Snacks.

I haven't updated the README as the README does not enumerate the included bindings.


mehalter:
this isn't really going to work because those mappings do not get set in Neovim v0.11. So if you check for them, they will only say they are  set in 0.10. These should just be unconditionally set.

mehalter (approved):
Thanks for the contribution!

I resolved the bugs in this so that it works. Please make sure to test PRs in the future to make sure the contributed code is functioning!

PR #1412, opened 2025-04-01
fix(nvim-dap-repl-highlights): nvim-dap in dependencies breaks lazy loading

RayJameson:
## 📑 Description
nvim-dap in dependencies breaks lazy loading of the nvim-dap plugin and all related plugins

## ℹ Additional Information
before this patch dap* plugins loaded on nvim open:
image


RayJameson:
this PR is wrong, it seems the reason for breaking lazy loading is my own config...

RayJameson:
No it wasn't ... 
I just removed `opts = {}` from dap-repl config, which caused break it, but nvim-dap in dependencies is still breaking the lazy loading, so this time it should work

PR #1413, opened 2025-04-01
feat(indent): add chunk highlight like hlchunk with snacks.nvim

radiol:
Add a recipe that simulates hlchunk.nvim's chunk visualization using snacks.nvim.

## 📑 Description

This recipe adds `snacks.nvim` configuration to replicate [ shellRaining/hlchunk.nvim ](https://github.com/shellRaining/hlchunk.nvim)-like indent chunk visualization.

### Screenshot
image


radiol:
Moved the recipe to the `indent` category instead of `recipes`.

lua/astrocommunity/indent/snacks-highlight-chunk/init.lua
```diff
@@ -0,0 +1,26 @@
...
+        },
+      },
+      animate = {
+        enabled = true,
+        style = "down",
+        duration = {
+          step = 20,
+          total = 300,
+        },
+      },
```
mehalter, line 23:
```suggestion
```

I don't think this spec should modify the user's animate settings

mehalter (changes requested):
I think it might be better to call it `snacks-indent-hlchunk`, and I do agree it should go in `indent` rather than `recipes`. Also left a comment

mehalter (approved):
Applied suggestions and this looks great! Thanks so much!

PR #1416, opened 2025-04-01
feat(completion): add blink-cmp-tmux

Kamilcuk:
## 📑 Description

Add blink-cmp-tmux

## ℹ Additional Information

moving to blink

### DIFF
--- a/lua/astrocommunity/completion/blink-cmp-tmux/init.lua
+++ b/lua/astrocommunity/completion/blink-cmp-tmux/init.lua
@@ -1,0 +1,16 @@
+return {
+  "saghen/blink.cmp",
+  dependencies = { "mgalliou/blink-cmp-tmux" },
+  opts = function(_, opts)
+    opts.sources = vim.tbl_deep_extend("keep", opts.sources or {}, {
+      default = {},
+      providers = {
+        tmux = {
+          module = "blink-cmp-tmux",
+          name = "tmux",
+        },
+      },
+    })
+    table.insert(opts.sources.default, "tmux")
+  end,
+}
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/completion/blink-cmp-tmux/init.lua
```diff
@@ -0,0 +1,16 @@
...
+      providers = {
+        tmux = {
+          module = "blink-cmp-tmux",
+          name = "tmux",
+        },
+      },
+    })
+    table.insert(opts.sources.default, "tmux")
+  end,
+}
```
mehalter, line 16:
```suggestion
return {
  "mgalliou/blink-cmp-tmux",
  lazy = true,
  specs = {
    {
      "Saghen/blink.cmp",
      optional = true,
      opts = {
        sources = {
          default = { "tmux" },
          providers = {
            tmux = { name = "tmux", module = "blink-cmp-tmux" },
          },
        },
      },
    },
  },
}
```

This can be simplified to this which (1) is clearer to read and (2) makes it easier for the user to disable conditionally in their user configuration from the `plugins/` side.

mehalter (changes requested):

mehalter (approved):
Looks great! Thanks!

Kamilcuk:
Hi, wow, the `default = { "tmux" }` does not replace the list, instead appends. How is the merging done? This is some code in Lazy? Thanks.

mehalter:
`lazy.nvim` has a feature where `opts` keys can be specified as a key that should be extended rather than merged which we use for `blink.cmp`'s `sources.default` key

PR #1422, opened 2025-04-03
fix(rust): use a stable path for `codelldb` in Mason

mehalter:
## 📑 Description

Mason paths within a package are not really considered stable, so we should rely on stable paths if available. I have a PR open to resolve this stable path and then we can merge this in

Relies on https://github.com/mason-org/mason-registry/pull/9565


mehalter:
This PR got merged in, so this is ready for review! Someone should reinstall `codelldb` in Mason and test out this PR just to make sure it works!

PR #1429, opened 2025-04-04
fix(copilot-lua-cmp): make nvim-cmp optional

mikemykhaylov:
## 📑 Description

After migration to v5, when using this plugin, nvim-cmp is installed automatically, which does not necessarily need to happen if blink.cmp is installed

## ℹ Additional Information

Old behavior after v5: plugin is enabled, nvim-cmp is installed automatically despite a valid completion engine (blink.cmp) being present
New behavior after v5: plugin is enabled, nvim-cmp is not set up unless already installed


azdanov (approved):

azdanov:
Thanks!

PR #1430, opened 2025-04-06
fix(leap-nvim): Fix cursor invisible bug on nvim 0.10+

ALameLlama:
We added this autocmd to fix the cursor issue in nvim 0.9 and having around around broke 0,10, since we no longer support 0.9, I think we can safely remove it

Related Issue: https://github.com/AstroNvim/astrocommunity/issues/1219


Uzaaft (approved):
Love a cleanup.
:shipit:

PR #1431, opened 2025-04-06
feat(edgy-nvim)!: Remove <Leader>f mapping from community plugin

shivanthzen:
## 📑 Description
<leader>f is prefix for all picker actions.
The keymaps for  edgy overlaps with this. Remapping the keybindings for edgy


azdanov:
What is the reason to remove it? It conflicts with an existing mapping?

ButterSus:
image

`<leader>f` is prefix for all picker actions.
It actually overlaps. Wouldn't it be better to remap it? `<leader>ue` for example

Uzaaft:
Any updates on this? @shivanthzen

shivanthzen:
@Uzaaft

Uzaaft:
Answer the question that azdanov asked @shivanthzen before pinging me...

azdanov:
Looks good. Thanks! @Uzaaft what do you think? Seems like a breaking change, so I updated the title accordingly.

azdanov (approved):

Issue #1414, opened 2025-04-01
Failed to run `config` for nvim-cmp


shibotto:
### Checklist

- [x] I have searched through the AstroNvim documentation
- [x] I have searched through the existing issues of this project
- [x] I have searched the existing issues of plugins related to this issue
- [x] I can replicate the bug with the minimal `repro.lua` provided below

### Neovim version (nvim -v)

0.10.4

### Operating system/version

GNU/Linux

### Terminal/GUI

(originally) gnome-terminal

### Describe the bug

nvim-cmp fails to load with:

```
Failed to run `config` for nvim-cmp

...ommunity/lua/astrocommunity/completion/nvim-cmp/init.lua:205: attempt to index field 'mapping' (a nil value)

# stacktrace:
  - /astrocommunity/lua/astrocommunity/completion/nvim-cmp/init.lua:205 _in_ **values**
```

### Steps to Reproduce

1. Install nvim-cmp from astrocommunity
2. Try to trigger completion
3. See notification about failing to run config for nvim-cmp

### Expected behavior

No error.

### Additional Context

Adding

```
  { "Saghen/blink.cmp", enabled = false },
  { "Saghen/blink.compat", enabled = false },
```

similarly to https://github.com/AstroNvim/astrocommunity/issues/1403 doesn't seem to make a difference (although it's probably required).

### Minimal configuration

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "runtime", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  -- stylua: ignore
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", "--branch=stable", lazypath })
end
vim.opt.rtp:prepend(vim.env.LAZY or lazypath)

-- install plugins
local plugins = {
  { "AstroNvim/AstroNvim", import = "astronvim.plugins" },
  { "AstroNvim/astrocommunity", import = "astrocommunity.completion.nvim-cmp" },

  -- add any other plugins/customizations here
}
require("lazy").setup(plugins, {
[...]

PR #1433, opened 2025-04-07
fix(nvim-cmp): misspelled `mapping` variable

shibotto:
Fixes https://github.com/AstroNvim/astrocommunity/issues/1414


azdanov (approved):
Thanks!

PR #1434, opened 2025-04-09
fix(codeium-nvim): add symbol_map initialization in mini.icons configuration

ButterSus:
## 📑 Description

This PR fixes an issue in the Codeium plugin's mini.icons configuration where it was trying to access `opts.symbol_map.codeium` without first checking if `opts.symbol_map` exists.

- [x] Add `if not opts.symbol_map then opts.symbol_map = {} end` check before setting codeium icon in mini.icons configuration

## ℹ Additional Information

This fix prevents an error that would occur when `opts.symbol_map` is nil, similar to how it's already checked in the lspkind configuration above it.

### DIFF
--- a/lua/astrocommunity/completion/codeium-nvim/init.lua
+++ b/lua/astrocommunity/completion/codeium-nvim/init.lua
@@ -64,6 +64,7 @@ return {
       -- Adds icon for codeium using mini.icons
       opts = function(_, opts)
         if not opts.lsp then opts.lsp = {} end
+        if not opts.symbol_map then opts.symbol_map = {} end
         opts.symbol_map.codeium = { glyph = require("astroui").get_icon("Codeium", 1, true), hl = "MiniIconsCyan" }
       end,
     },


lua/astrocommunity/completion/codeium-nvim/init.lua
```diff
@@ -42,13 +42,28 @@ return {
...
+    {
+      "Saghen/blink.cmp",
+      optional = true,
+      dependencies = "Exafunction/codeium.nvim",
+      specs = { "Saghen/blink.compat", version = "*", lazy = true, opts = {} },
+      opts = {
+        sources = {
+          default = { "codeium" },
+          providers = {
+            codeium = { name = "codeium", module = "blink.compat.source", score_offset = 10000 },
```
Uzaaft, line 62: Why the score offset?

lua/astrocommunity/completion/codeium-nvim/init.lua
```diff
@@ -64,6 +79,7 @@ return {
       -- Adds icon for codeium using mini.icons
       opts = function(_, opts)
         if not opts.lsp then opts.lsp = {} end
+        if not opts.symbol_map then opts.symbol_map = {} end
```
klmr, line 82: The same fix is also required for `copilot-cmp/init.lua`.

Uzaaft: No, make a seperate PR.

lua/astrocommunity/completion/codeium-nvim/init.lua
ButterSus: sorry, accidentally pushed to main branch, it was not intended to be in pull request

ButterSus: Force pushed to revert unnecessary commit

Yorizel:
fix the title

Uzaaft (approved):

PR #1437, opened 2025-04-10
fix(picker-lsp-mappings): picker-lsp-mappings in recipers to work with telescope and picker

Mega31:
## 📑 Description
- previously, lsp mapping would not be applied to use snack.picker when telescope is installed due the use of following plugin: 

  ```lua
  { import = "astrocommunity.debugging.telescope-dap-nvim" },
  ```


Uzaaft:
This doesnt make sense in the case of the telescope-nvim-snacks pack(pr recently opened), where we would just override the settings. I think, if telescope is installed, then use that. If anything, we should swap out the order.

PR #1438, opened 2025-04-12
feat(pack): add golangci-lint

azdanov:
## 📑 Description
This pack adds [golangci-lint](https://golangci-lint.run/) support using the [golangci-lint-langserver](https://github.com/nametake/golangci-lint-langserver).

## ℹ Additional Information
Not sure If it needs to be backwards compatible between golangci-lint v1 and v2. Also not sure about AstroNvim v4 and v5. So it just uses latest version of all and requires astro v5.

A custom lsp config override is needed, but it seems to work fine https://github.com/nametake/golangci-lint-langserver/issues/51


Uzaaft (commented):
Why not add this into the go pack?

azdanov:
Can be. I thought to keep it modular. Not entirely sure how common this tool is in the Go world. As an example ESlint is kind of a standard at this point in js/ts.

ALameLlama:
If this one is dependent of the go pack maybe this one should import it so users don't need to do

```
{ import = "astrocommunity.pack.go" },
{ import = "astrocommunity.pack.golangci-lint " }
```

I also can't see a linter in the current go pack but I assume this is built into another tool. I'm not that familiar with the go eco system.

azdanov:
That's a good idea to bundle them together. I was playing around with a more modular design approach to community packs. 

If there are no hard dependencies between them then maybe it's better not to hardcore and let the people choose. To help avoid mega packs like typescript. 

In this case it does make sense since why would anyone use it without a go pack.

azdanov:
> I also can't see a linter in the current go pack but I assume this is built into another tool. I'm not that familiar with the go eco system.

gopls (lsp) has some linting built in that's quite good. This just adds a lot of community tooling that are out there. As an example can check out this config: https://gist.github.com/maratori/47a4d00457a92aa426dbd48a18776322

Uzaaft (approved):

PR #1448, opened 2025-04-15
feat(bash): add zsh support

ALameLlama:
## 📑 Description

While bashls doesn't officially support zsh, the syntax is almost the same and will get you most of the way there to make dealing with zsh files much better.

### DIFF
--- a/lua/astrocommunity/pack/bash/README.md
+++ b/lua/astrocommunity/pack/bash/README.md
@@ -7,3 +7,5 @@ This plugin pack does the following:
 - Adds `shfmt` formatter
 - Adds `shellcheck` linter
 - Adds `bash` debugger
+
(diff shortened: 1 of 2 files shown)


ALameLlama:
@Uzaaft I've updated the readme, pretty much every time I edit a file that is zsh, I end up setting the filetype to sh and it does the trick.

I'm all for suggestions on making working with zsh files better

Uzaaft:
> @Uzaaft I've updated the readme, pretty much every time I edit a file that is zsh, I end up setting the filetype to sh and it does the trick.
> 
> I'm all for suggestions on making working with zsh files better

I tried it once, found out that it doesn't work that well due to shellcheck and disabled it 😆

Uzaaft (approved):

PR #1451, opened 2025-04-18
fix(noice-nvim): don't override the lsp_progress component

meijieru:
## 📑 Description

Since disabling lsp_progress in the statusline is hardcoded, it interferes with user-configured statuslines and is difficult to debug. We'll disable it temporarily.

### DIFF
--- a/lua/astrocommunity/utility/noice-nvim/init.lua
+++ b/lua/astrocommunity/utility/noice-nvim/init.lua
@@ -57,16 +57,6 @@ return {
       end,
     },
     {
-      "rebelot/heirline.nvim",
-      optional = true,
-      opts = function(_, opts)
-        local noice_opts = require("astrocore").plugin_opts "noice.nvim"
-        if vim.tbl_get(noice_opts, "lsp", "progress", "enabled") ~= false then -- check if lsp progress is enabled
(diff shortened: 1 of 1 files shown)


Uzaaft:
> ## 📑 Description
> Since disabling lsp_progress in the statusline is hardcoded, it interferes with user-configured statuslines and is difficult to debug. We'll disable it temporarily.

I believe you have misunderstood it. We don't disable it, we check if it's enabled, and if it is, we add in a component.

meijieru:
Enabling the lsp_progress feature in Noice can conflict with user-customized status lines. Because it uses a hardcoded index, it may overwrite incorrect components if the user has changed the layout. This feature should be disabled until a reliable way to dynamically locate the lsp_progress component is implemented.

Uzaaft:
> Enabling the lsp_progress feature in Noice can conflict with user-customized status lines. Because it uses a hardcoded index, it may overwrite incorrect components if the user has changed the layout. This feature should be disabled until a reliable way to dynamically locate the lsp_progress component is implemented.

Now this makes perfect sense. Merging :)

Uzaaft (approved):
:shipit:

PR #1453, opened 2025-04-18
feat(completion): Add in codex.nvim

Uzaaft:
This might be one of the case where we need to prefix the author to the name, since I recon it could be a flood of codex.nvim plugins. WDYT? @AstroNvim/astrocommunity-maintainers

### DIFF
--- a/lua/astrocommunity/completion/codex-nvim/README.md
+++ b/lua/astrocommunity/completion/codex-nvim/README.md
@@ -1,0 +1,7 @@
+# Codex.nvim
+
+OpenAI Codex plugin for Neovim
+
+For more information, please refer to:
+
+Repository: <https://github.com/johnseth97/codex.nvim>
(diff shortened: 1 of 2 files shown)


azdanov:
Yeah, would be good idea to add the author. It's already confusing with copilot.nvim and copilot.vim for some people :smile:

johnseth97:
There is still some work to be done yet on the plugin, planning on pushing out some tests today and checking compliance with the neovim spec.

johnseth97:
Update from last comment: Plugin has been updated with an auto-install flag, which will auto install codex via npm if not available. Otherwise it will notify the user and ask them to run the command `npm i -g codex`, as the default behavior. Codex will notify the user if they need to set `$OPENAI_API_KEY` env.

Depending on how hands-off we want the default config to be for the user, I would recommend setting `autoinstall=true` in the opts table

The default plugin configuration is as follows, all can be overridden with the opts table in the plugin setup call as shown in the plugin repo


```lua
opts = {
    keymaps = {}, -- disable internal mapping
    border = 'rounded', -- or 'double'
    width = 0.8,
    height = 0.8,
    autoinstall = true,
  }
```

azdanov (approved):
:+1:

PR #1456, opened 2025-04-18
feat(editing-support): add nvim-paredit plugin

practicalli-johnny:
## 📑 Description

Adds the [julienvincent/nvim-paredit](https://github.com/julienvincent/nvim-paredit) plugin to provide structural editing support for Lisp dialects.

## ℹ️  Additional Information

nvim-paredit plugin supports the changes in Neovim 0.11 and will be included in the Clojure pack in a future PR (once this PR has been approved)

PR #1457, opened 2025-04-18
feat(editing-support): add parpar-nvim plugin

practicalli-johnny:
## 📑 Description

Plugin ensures nvim-paredit and nvim-parinfer work well together

Imports both plugins from astrocommunity rather than including the plugins as dependencies (please comment if this is not the correct approach)

- [x] PR depends on acceptance of #1456

## ℹ️  Additional Information

Parpar provides the two common approaches in one plugin and ensures the plugins do not conflict with each other when editing.

Parpar can be imported into the Clojure pack and simplify its code (in a future PR)

### DIFF
--- a/lua/astrocommunity/editing-support/parpar-nvim/README.md
+++ b/lua/astrocommunity/editing-support/parpar-nvim/README.md
@@ -1,0 +1,13 @@
+## parpar-nvim
+
+A plugin for Neovim that combines two alternative approaches for structural editing with Lisp dialects.
+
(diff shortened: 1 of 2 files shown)


Uzaaft:
Marking it as a draft until https://github.com/AstroNvim/astrocommunity/pull/1456 is merged.

Uzaaft (approved):

Uzaaft (approved):

PR #1466, opened 2025-04-24
Resolving key mapping conflicts

msdone-lwt:

### DIFF
--- a/lua/astrocommunity/completion/avante-nvim/init.lua
+++ b/lua/astrocommunity/completion/avante-nvim/init.lua
@@ -36,7 +36,7 @@ return {
       toggle = {
         default = prefix .. "t",
         debug = prefix .. "d",
-        hint = prefix .. "h",
+        hint = prefix .. "H",
         suggestion = prefix .. "s",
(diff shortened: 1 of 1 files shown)


Yorizel:
what was the conflict?

msdone-lwt:
> what was the conflict?

history

lua/astrocommunity/completion/avante-nvim/init.lua
```diff
@@ -36,7 +36,7 @@ return {
       toggle = {
         default = prefix .. "t",
         debug = prefix .. "d",
-        hint = prefix .. "h",
+        hint = prefix .. "H",
```
Uzaaft, line 39: Whats this conflicting with?

Yorizel: i'm asking the same question, i tested on my local environment, and couldn't find anything

Uzaaft:
@msdone-lwt As a general rule, please provide a description of what your PR fixes. that way things go smoother.

PR #1470, opened 2025-04-25
feat(clojure): add other-nvim

practicalli-johnny:
## 📑 Description

Other-nvim now supports Clojure

Upstream PR: https://github.com/rgroli/other.nvim/pull/67

related: #1469

### DIFF
--- a/lua/astrocommunity/pack/clojure/README.md
+++ b/lua/astrocommunity/pack/clojure/README.md
@@ -17,6 +17,7 @@ The Clojure language pack includes:
 - [clojure-lsp](https://clojure-lsp.io/) language server
 - Clojure REPL client: [Olical/conjure](https://github.com/Olical/conjure)
 - Structural editing: [parpar-nvim](https://github.com/dundalek/parpar.nvim) which provides both [nvim-parinfer](https://github.com/gpanders/nvim-parinfer) & [nvim-paredit](https://github.com/julienvincent/nvim-paredit)
+- Alternative files: switch between src and test using [other.nvim](https://github.com/rgroli/other.nvim)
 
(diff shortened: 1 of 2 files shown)


Uzaaft:
Whats the advantage of having this in the pack?

practicalli-johnny:
> Whats the advantage of having this in the pack?

Switching between an implementation file and its associated test file will be done very frequently (many times an hour) for Clojure development.  Each implementation file has a unit test file associated with it.

For example: Other-nvim provides a way to navigate from `src/practicalli/gameboard/service.clj` to `src/practicalli/gameboard/service_test.clj` (and back again) with one command, rather than having to navigate via neogit each time or via open buffers (once both files are open).

Uzaaft:
I dont think its needed in the clojure pack. If the end-user needs this, they can just add it themselves. 

> I think it would be enough to mention the other-nvim pack in the clojure readme, and let the user decide if they wish to add it.

I second this.

practicalli-johnny:
> I dont think its needed in the clojure pack. If the end-user needs this, they can just add it themselves.
> 
> > I think it would be enough to mention the other-nvim pack in the clojure readme, and let the user decide if they wish to add it.
> 
> I second this.

Please explain why is it not needed?  Have you discussed this with other Clojure developers?

I clearly stated why I believe it is valuable to include other-nvim with a specific example.  Was this explanation not clear how much of a benefit this provides.  

It certainly is something I've sorely missed in Neovim and switch back to Emacs for serious development to be more effective.

azdanov:
It is a very cool plugin that I didn't even know existed. But it's not something Clojure can't function without. If this is included by default, then some developers who don't wish to use it will have to disable it. It's easier to opt in than opt out.

Uzaaft:
It seems to me to be more of a workflow thing, rather than a clojure spesific workflow. 
You are free to add it into your own, spesific setup. azdanov said it perfectly. 

Going off by your logic, this should also be added into typescript, ocaml, and rust packs. Cuz the workflow could be similar.

PR #1475, opened 2025-05-02
docs(typescript): add section about double formatting issue

azdanov:
## 📑 Description
This adds a section about double formatting due to bad configuration of eslint/prettier which might result in odd behavior in some projects.

### DIFF
--- a/lua/astrocommunity/pack/typescript/README.md
+++ b/lua/astrocommunity/pack/typescript/README.md
@@ -36,6 +36,18 @@ return {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Uzaaft (commented):
Is this really needed? User should read the prettier/eslint docs which also mentions this behaviour.

azdanov:
I guess not needed. I wonder how vscode handles this. I never had this behavior on intellij but they use their own proprietary LSP which are probably faster than the ESlint one.

PR #1482, opened 2025-05-07
fix(nvim-spider): provide UTF-8 support by default

ButterSus:
## 📑 Description

According to [README.md](https://github.com/chrisgrieser/nvim-spider?tab=readme-ov-file#utf-8-support), `luautf8` rocks package is required to correctly handle UTF-8 layouts (CJK, cyrillic, etc.).
Otherwise user won't be able to jump word-wise in layout other than english.

I added [luarocks.nvim](https://github.com/vhyrro/luarocks.nvim) dependency to install it.

### DIFF
--- a/lua/astrocommunity/motion/nvim-spider/init.lua
+++ b/lua/astrocommunity/motion/nvim-spider/init.lua
@@ -31,6 +31,13 @@ return {
         },
       },
     },
+    {
+      "vhyrro/luarocks.nvim",
+      priority = 1000,
+      opts = {
+        rocks = { "luautf8" },
+      },
+    },
   },
   opts = {},
 }


lua/astrocommunity/motion/nvim-spider/init.lua
```diff
@@ -31,6 +31,13 @@ return {
         },
       },
     },
+    {
+      "vhyrro/luarocks.nvim",
+      priority = 1000,
+      opts = {
+        rocks = { "luautf8" },
+      },
+    },
```
Uzaaft, line 40: Cant we instead use the lazy.nvim builtin way of fetching luarocks?

Uzaaft (approved):
Author didnt answer. Merging it as is. Might reconsider the luarocks plugin in the future.

PR #1488, opened 2025-05-11
fix(blink-cmp-git): Remove unneeded dependency

Uzaaft:
https://github.com/Kaiser-Yang/blink-cmp-git/commit/2d9f19078c452988f13834d9ef207acfa37f1eaf


azdanov (approved):
:+1:

Issue #1501, opened 2025-05-16
Refactor typescript pack


azdanov:
### Is your feature related to a problem?

At the moment, the TypeScript pack has too many responsibilities (TypeScript, ESLint, Prettier).

### Describe the new feature

It would be good to separate it into TypeScript, ESLint, and Prettier packs. The reason is that there are other tools that can be used for formatting and linting, and different combinations of them. At the moment, it's impossible to use those combinations, e.g., pure ESLint for linting and formatting, or no ESLint and only Prettier, or even using Biome.

### Additional context

Related issue: #962

Uzaaft:
Lets target this for release 20

azdanov:
This can be done step by step.

1. Create separate Prettier pack
2. Create separate ESLint pack
3. Remove Prettier and ESLint from TypeScript pack
4. (Optional) Add [Biome](https://biomejs.dev) pack

Uzaaft:
Me like this. Seems like the right direction to move in.
Would make debugging and fixing eslint stuff easier I recon

Issue #1496, opened 2025-05-15
Improve ESLint in typescript pack


azdanov:
### Is your feature related to a problem?

At the moment it's just an auto command, which is sufficient.

### Describe the new feature

LazyVim seems to have it done in a different way. It's mentioned in conform FAQ that running LSP formatting (EslintFixAll) rewrites the whole buffer, which is very inefficient for large files.

Would be nice to figure out if it's possible to use the conform approach of just using a diff update.

### Additional context

https://github.com/LazyVim/LazyVim/blob/main/lua/lazyvim/plugins/extras/linting/eslint.lua

https://github.com/stevearc/conform.nvim?tab=readme-ov-file#faq
https://github.com/stevearc/conform.nvim/issues/502
https://github.com/stevearc/conform.nvim/issues/466
https://github.com/stevearc/conform.nvim/issues/222

Uzaaft:
I dont see any big differences in LazyVims approach tbh.
We could consider removing the eslint fix on save fix, which based on the comment in the lazyvim code is not needed bellow neovim 0.10

Uzaaft:
I think this change makes sense.
Would make maintaining them easier as well. 
@azdanov Lets target this for the 20th release

PR #1503, opened 2025-05-17
feat(pack): add eslint

azdanov:
Issue: #1501 
## 📑 Description
Adds ESLint based on the typescript pack. Needs testing and maybe potentially adding `eslint.d` support to replace `EslintFixAll` auocmd.

## ℹ Additional Information
See #1496 for discussion on EslintFixAll


azdanov:
I think eslint.d is not needed, since eslint lsp does it all.
Maybe it improves performance when using conform and prettier, so you can configure a chain "prettier -> eslint"? But I don't know.

lua/astrocommunity/pack/eslint/README.md
```diff
@@ -0,0 +1,14 @@
+# ESLint
+
+[ESLint](https://eslint.org/) is a tool for identifying and reporting on
+patterns found in ECMAScript/JavaScript code, with the goal of making code more
+consistent and avoiding bugs.
+
+ESLint is completely pluggable. Every single rule is a plugin and you can add
+more at runtime. You can also add community plugins, configurations, and
+parsers to extend the functionality of ESLint.
```
Uzaaft, line 9:
```suggestion
[ESLint](https://eslint.org/) statically analyzes your code to quickly find problems. It is built into most text editors and you can run ESLint as part of your continuous integration pipeline.
```

lua/astrocommunity/pack/eslint/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Prerequisites
+
+To use ESLint, you must have [Node.js](https://nodejs.org/en/) (^18.18.0, ^20.9.0, or >=21.1.0) installed
+and built with SSL support. (If you are using an official Node.js distribution, SSL is always built in.)
```
Uzaaft, line 14: I believe this might be wrong per the https://github.com/microsoft/vscode-eslint docs. Version 3 appears, if I read the docs right, to require nodejs 20? 
https://github.com/microsoft/vscode-eslint#version-3010---release

PR #1507, opened 2025-05-18
fix(docker): filter out docker compose filtering instead of disabling

azdanov:
## 📑 Description
This was causing an issue because of overwriting what conform was setting here: https://github.com/AstroNvim/astrocommunity/blob/b69674df8de9a083754d44fb8a96d8dac1c06860/lua/astrocommunity/editing-support/conform-nvim/init.lua#L6 and enabling back the astrolsp formatting

## ℹ Additional Information
Use filter instead: https://neovim.io/doc/user/lsp.html#vim.lsp.buf.format()

### DIFF
--- a/lua/astrocommunity/pack/docker/init.lua
+++ b/lua/astrocommunity/pack/docker/init.lua
@@ -8,11 +8,10 @@ return {
   {
     "AstroNvim/astrolsp",
     optional = true,
+    ---@type AstroLSPOpts
     opts = {
       formatting = {
-        disabled = {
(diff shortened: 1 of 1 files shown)


Uzaaft (approved):

Uzaaft:
LGTM. Have you tested it @azdanov

azdanov:
> LGTM. Have you tested it @azdanov

Yeah, it worked. Docker compose was using prettier and conform pack was disabling astrolsp formatting.

PR #1510, opened 2025-05-20
fix(avante-nvim): prevent neo-tree mapping warnings

krmcbride:
## 📑 Description

Scopes the `oa` neo tree mapping under the filesystem config to prevent `invalid mapping` warnings when viewing other sources (git, buffers, etc).


azdanov (approved):
Thanks!

Issue #1500, opened 2025-05-16
Clean up catppuccin pack


azdanov:
### Describe the new feature

I noticed that there are a lot of integrations that are enabled in the pack itself, but no longer ship with the latest AstroNvim. It would be good to disable them and move them into the specs of the respective packs if those exist.
Also, some integrations are enabled by default, so they can also be removed.

omagarwal25:
I'd be willing to work on this!

omagarwal25:
Did something go wrong with my PR? @azdanov

azdanov:
@omagarwal25 

Can check on discord https://discord.com/channels/939594913560031363/1224731564261507072/1382269567794937946

omagarwal25:
Umm so what's the next steps here, should I go back to the old approach of enabling things if present only? Or just leave everything as is? Sorry about the mishap. Weird that it breaks on LSPSaga specifically though.

azdanov:
Difficult to say what's next 😄 
I'd just forget about it. I stopped using catppuccin theme, so can't test it properly myself

azdanov:
@omagarwal25 If you feel up to it and think this is a useful contribution, then of course you can continue and fix the issue. You're part of the astrocommunity. I'm at the moment overwhelmed with my job to be of much help :)

PR #1513, opened 2025-05-20
fix(catppuccin): cleaned up catppuccin pack 

omagarwal25:
## 📑 Description
Cleaned up the Catppuccin Pack

- [x] removed unnecessary default integrations
- [x] added integrations for all supported packs
- [x] added integrations by default for all astronvim default plugins

This addresses #1500


Uzaaft:
This PR needs extensive review so gonna hold of until later today.

azdanov:
Thanks!

Oh wow, that's a lot of changes. I tried using the Catppuccin theme a while ago, but returned to my old theme.

I'll see if I can find time to check this out eventually.

omagarwal25:
Thanks!

azdanov:
I've checked it. It looks good. Thanks! But I started having doubts about catppuccin.

Maybe it would make sense to enable all those plugins by default in the catppuccin theme, instead of embedding it everywhere in specs? What do you think?

I hope that one day they implement auto integration as other modern plugins do: https://github.com/catppuccin/nvim/issues/838

omagarwal25:
I'm happy to do that too, just let me know

azdanov:
Then could you please update the theme file only? This way it's nice and in one place. I just realized that if other themes have similar setup, and need updating for other plugins then it won't be maintainable at all :slightly_smiling_face:

omagarwal25:
Of course! I only did it this way as it was recommended in the issue. I'll get on it sometime this week. Should I make a new PR or update this one?

azdanov:
Can update this one. Thanks!

Yeah it was my mistake, not thinking the issue through :) At first it seemed reasonable, but thanks to your PR I can see how much unnecessary coupling it would introduce to other packs.

omagarwal25:
Is there any downside in having all the integrations enabled by default?

azdanov:
Don't think so. Just a bit more load time probably

omagarwal25:
All done

azdanov (approved):
Thank you for your time and work!
Closes: https://github.com/AstroNvim/astrocommunity/issues/1500

tofulm:
with this update, I've an error : `catppuccin.groups.integrations.lsp_saga not found`

Uzaaft:
Its fixed now. Please update @tofulm

PR #1529, opened 2025-05-31
docs(java): add section about spring boot

azdanov:
## 📑 Description
Added a section on how to add spring-boot support to the pack.


Uzaaft:
Cant this be a seperate pack?

azdanov:
> Cant this be a seperate pack?

Yeah I think so. It does need jdtls to work, but otherwise could be useful as a separate one.

PR #1535, opened 2025-06-01
fix(ansible): improve regex pattern to avoid false matches

### LINKED ISSUE #1523: File incorrectly seen as type yaml.ansible
### Checklist

### Neovim version (nvim -v)

v0.10.4

### Operating system/version

Fedora 41

### Terminal/GUI

alacritty

### Describe the bug

Whenever a file yaml contains a nested key `host` or `tasks`, it gets incorrectly recognized as ansible playbook.
The regex responsible for this behaviour is the following:
https://github.com/AstroNvim/astrocommunity/blob/7a007c1557e8c1dd0066bdd7e5899d12784f9930/lua/astrocommunity/pack/ansible/init.lua#L11

I think it's better to only match a root-level key, it helps to avoid false matches.
```lua
local regex = vim.regex "^(hosts\\|tasks):"
```

### Steps to Reproduce

1. Import "astrocommunity.pack.ansible"
2. Open any yaml file containing a nested key `host` or `tasks`. Such as following kubernetes manifest:
  ```yaml
  apiVersion: v1
  kind: ConfigMap
  metadata:
    name: config-webapp
  data:
    host: "http://192.168.1.1"
  ```
3. File is recognized as `yaml.ansible` and ansible-lint is started

### Expected behavior

File should be recognized as `yaml`
 ```yaml
  apiVersion: v1
  kind: ConfigMap
  metadata:
    name: config-webapp
  data:
    host: "http://192.168.1.1"
  ```

### Screenshots

_No response_

### Additional Context

_No response_

### Minimal configuration

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
end

-- bootstrap lazy
  -- stylua: ignore
end

-- install plugins
local plugins = {
  { "AstroNvim/astrocommunity", import = "astrocommunity.pack.ansible" },

}
  root = root .. "/plugins",
})

```

AlePini:
Closes #1523 

## 📑 Description
The current regex is leading to multiple false matches, because the keywords `hosts` and `tasks` are not exclusively used in ansible files.
Ansible playbooks have those keywords at the start of the line, so it's just easier to search for `^(hosts\\:tasks)`

### DIFF
--- a/lua/astrocommunity/pack/ansible/init.lua
+++ b/lua/astrocommunity/pack/ansible/init.lua
@@ -8,7 +8,7 @@ local function yaml_ft(path, bufnr)
   if path_regex and path_regex:match_str(path) then return "yaml.ansible" end
 
   -- check for known ansible playbook text and if found, return yaml.ansible
-  local regex = vim.regex "hosts:\\|tasks:"
+  local regex = vim.regex "^(hosts\\|tasks):"
   if regex and regex:match_str(content) then return "yaml.ansible" end
 
   -- return yaml if nothing else


azdanov (approved):
Thanks!

PR #1537, opened 2025-06-03
fix(avante-nvim): Check for nil value in opts table for markview.nvim dependency

lucazz:
## 📑 Description
**`markview.nvim` Plugin Enhancements (Avante integration)**
   * Added a defensive check for `opts.preview` to avoid potential `nil` dereferencing.
   * Ensures compatibility and resilience when plugin options are partially defined.

## ℹ Additional Information
* This update to `markview.nvim` ensures that the plugin doesn't fail if `opts.preview` is undefined, which may occur depending on other plugin configurations.

### DIFF
--- a/lua/astrocommunity/completion/avante-nvim/init.lua
+++ b/lua/astrocommunity/completion/avante-nvim/init.lua
@@ -98,6 +98,7 @@ return {
       "OXY2DEV/markview.nvim",
       optional = true,
       opts = function(_, opts)
+        if not opts.preview then opts.preview = {} end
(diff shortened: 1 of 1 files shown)


Uzaaft:
Brother in christ that PR message and commit message is so bad.

Uzaaft (approved):

Uzaaft:
@lucazz Advice: Just keep it simple.
> just added null checks

Would have been enough with:
> `fix(avante-nvim): Check for nil value in dependency opts`

as the PR title or commit message or something.

PR #1538, opened 2025-06-06
fix(nvim-treesitter-context): fix outdated commands

ALVAROPING1:
## 📑 Description

nvim-treesitter-context changed the user commands to a single `TSContext` cmd with subcommands `enable`/`disable`/`toggle`. This patch updates the plugin spec to reflect those changes


azdanov (approved):
Thanks!

PR #1539, opened 2025-06-07
fix(kulala-nvim): fix keymaps config

YaroSpace:
## 📑 Fix for keymaps setup of Kulala.nvim

Fix `global_keymaps_prefix`
Add `lsp_keymaps` setup

### DIFF
--- a/lua/astrocommunity/programming-language-support/kulala-nvim/init.lua
+++ b/lua/astrocommunity/programming-language-support/kulala-nvim/init.lua
@@ -1,5 +1,3 @@
-local prefix = "<Leader>R"
-
 return {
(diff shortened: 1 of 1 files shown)


azdanov:
Thanks! Is `lsp_keymaps` something new? Can't find it in the docs anywhere.

YaroSpace:
The docs got updated later, after the merge. 
https://neovim.getkulala.net/docs/getting-started/keymaps#kulala-lsp-keymaps

Not really new, more of a workaround for distros that set lsp keymaps only for servers set up through lsp-config or in case of Astro, through Astrolsp.

PR #1542, opened 2025-06-12
fix(pack/json): rename schemastore to be lowercase

rawnly:
## 📑 Description

This solves a problem where lazy could not resolve dependency if it was already installed in user configuration (see video below)

https://github.com/user-attachments/assets/ab4a1bcb-119c-49ea-8027-74993fa3070e


Uzaaft:
@rawnly Could you split this PR into two? One for json and one for yaml.

PR #1548, opened 2025-06-16
fix(yaml): Normalize casing of schemastore plugin

rawnly:
## 📑 Description

This solves a problem where lazy could not resolve dependency if it was already installed in user configuration (see video below)

https://github.com/user-attachments/assets/ab4a1bcb-119c-49ea-8027-74993fa3070e

split from #1542


Uzaaft (approved):

mamachanko:
~~This issue returned after upgrading to `astrocommunity@v20.0.0`.~~

_edit: nm! it was my personal config which still had the wrong casing. sorry about that!_

PR #1551, opened 2025-06-19
fix(kulala-nvim): fix lsp keymaps

YaroSpace:
## 📑 Description

1. A cleaner fix for lsp keymaps.
2. Added lazy keymap.


azdanov:
Will be conflict with this pack: https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/editing-support/refactoring-nvim/init.lua

But dunno what other keys could be used.

Uzaaft:
@azdanov The key should have stayed as is. Or the PR should have been marked as breaking.

PR #1552, opened 2025-06-20
fix(coc-nvim): resolve <CR> conflicts and enable proper insert-mode completion

OlaleyeFisayo:
When integrating Coc.nvim into AstroNvim, I encountered issues with auto-completion — specifically the `<CR>` key not confirming completion items and the absence of expected smart formatting behavior.

After debugging, I discovered that `nvim-autopairs` was intercepting the `<CR>` mapping, which conflicted with Coc’s built-in completion and formatting logic. Since Coc already provides its own smart enter behavior and auto-pair handling, the most effective fix was to disable `nvim-autopairs` when using Coc.

This patch does the following:
- Disables the `nvim-autopairs` plugin to prevent `<CR>` mapping conflicts
- Adds default insert mode mappings for:
  - `<CR>` to confirm completion
  - `<Tab>` / `<S-Tab>` to navigate the popup menu
  - `<C-Space>` to manually trigger completion

This ensures a smoother and expected user experience when using Coc for autocompletion inside AstroNvim.

Tested with various filetypes and behaves as expected.

Let me know if you'd prefer this be added as a community plugin override instead.


lua/astrocommunity/lsp/coc-nvim/init.lua
```diff
@@ -72,6 +72,19 @@ return {
         end,
         desc = "Hover symbol details",
       }
+
+      if not maps.i then maps.i = {} end
```
Uzaaft, line 76: Is this needed? 
```suggestion
```

Uzaaft (approved):

PR #1554, opened 2025-06-20
feat(pack): add oxlint pack

Yorizel:


Yorizel:
@azdanov

taskylizard:
Looks good, the readme should probably be like the rest instead, see the other language pack for instance.

PR #1559, opened 2025-06-27
fix(nvim-cmp): Check for nil value in opts table

lixiao189:
## 📑 Description

`opts.sources` may not be intialized when `table.insert(opts.sources, { name = "lazydev", group_index = 0 })` is called
## ℹ Additional Information

image

### DIFF
--- a/lua/astrocommunity/completion/nvim-cmp/init.lua
+++ b/lua/astrocommunity/completion/nvim-cmp/init.lua
@@ -222,7 +222,10 @@ return {
       specs = {
         {
           "hrsh7th/nvim-cmp",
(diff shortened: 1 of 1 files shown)


lua/astrocommunity/completion/nvim-cmp/init.lua
```diff
@@ -222,7 +222,10 @@ return {
       specs = {
         {
           "hrsh7th/nvim-cmp",
-          opts = function(_, opts) table.insert(opts.sources, { name = "lazydev", group_index = 0 }) end,
+          opts = function(_, opts)
+            opts.sources = opts.sources or {} -- init opts.sources
```
Uzaaft, line 226:
```suggestion
            if not opts.sources then opts.sources = {} end
```

Uzaaft (approved):

PR #1563, opened 2025-06-28
feat(editing-support): adding vectorcode

Yorizel:


azdanov:
Cool that you're adding this also :slightly_smiling_face: 

I think it can even be used for other packs that we have: https://github.com/Davidyz/VectorCode/wiki/Neovim-Integrations
But that's a separate PR, if needed.

Yorizel:
i will add it

luxus:
there is some heirline support since 39 mins 
https://github.com/Davidyz/VectorCode/commit/e73bb482b33ca54dff19dff878ef8551058caf1d

Uzaaft (approved):

lua/astrocommunity/editing-support/vector-code-nvim/init.lua
```diff
@@ -0,0 +1,50 @@
+return {
+  "Davidyz/VectorCode",
+  build = function()
```
azdanov, line 3: https://github.com/Davidyz/VectorCode/blob/main/docs/neovim.md#masonnvim

Can Mason be used instead of `uv`?

It makes the setup simpler I think.

azdanov: Or maybe not good idea. Since plugin author mentions plugin pinning

Yorizel: i will take a look at it later

azdanov: @Uzaaft There is something, but author mentioned that version might be outdated. It could cause issues I guess if plugin is out of sync.
image

Uzaaft: Thats the newest one. Seems like a good PR :)

PR #1569, opened 2025-07-05
fix(obsidian-nvim): Option `use_advanced_uri` has been deprecated. Us…

wert2all:
…e in `open` module instead.

## 📑 Description

Neovim show a warning:

```
notify.warn Obsidian.nvim The config option 'use_advanced_uri' is deprecated, please use in `open` module instead

```


Uzaaft:
Rebased into #1570

PR #1574, opened 2025-07-08
refactor(tailwindcss): use root_dir function from nvim-lspconfig

azdanov:
## 📑 Description
After investigating the current root_dir setup I noticed that `neovim/nvim-lspconfig` https://github.com/neovim/nvim-lspconfig/blob/7fac9025a967a4d0846660f751cd392fac6bb788/lsp/tailwindcss.lua#L109 offers a similar (better) approach. And we don't need to maintain it.


Uzaaft:
Absolute cinema

PR #1580, opened 2025-07-17
fix(jj): properly handle cases where `neo-tree` and `telescope` are not available

mehalter:
replaces #1578

## 📑 Description

This adds proper handling of cases where the user may have disabled `neo-tree` and/or `telescope`


azdanov:
@iamralch Could you please test if this works correctly?

iamralch:
@mehalter  @azdanov it works perfectly. thank you.

PR #1593, opened 2025-07-25
chore(oxlint): hot patch `mason-lspconfig` with `oxlint` support

azdanov:
## 📑 Description
A better approach instead of always enabling oxlint.


mehalter:
fixed PR title to make more sense

azdanov:
Yep. It works

PR #1595, opened 2025-07-29
feat(quarto): dynamically set codeRunner.default_method based on avai…

iamralch:
Detects presence of "iron.core" and "molten.health" modules at runtime and sets `codeRunner.default_method` accordingly in the quarto-nvim plugin options.

### DIFF
--- a/lua/astrocommunity/pack/quarto/init.lua
+++ b/lua/astrocommunity/pack/quarto/init.lua
@@ -5,14 +5,14 @@ return {
     opts = function(_, opts)
       opts.codeRunner = opts.codeRunner or {}
 
-      local ironOk, _ = pcall(require, "iron.core")
-      if ironOk then
+      local iron_avail, _ = pcall(require, "iron.core")
+      if iron_avail then
         opts.codeRunner.default_method = "iron"
         return
       end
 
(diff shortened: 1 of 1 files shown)


azdanov:
Hey. What about `slime`? Although, I don't use this extension so no clue how it's supposed to work.

iamralch:
@azdanov slime is the default option. Do you want me to set it explicitly?

azdanov:
Nope, probably not needed if it's default. Just asking since I don't know anything about this plugin apart from reading docs.

Can there be a situation where a user has both iron and molten installed simultaneously?

iamralch:
I haven't used them in one configuration. I suspect that they are compatible. The difference is that molten is handling jupyter notebooks vs iron is a repl for python and other languages.

azdanov:
@Uzaaft Do you have any experience with these? I'm only certain of web dev stuff, this looks like data science.

Uzaaft (dismissed):
Not sure I like how one plugin is prioritized over the other, but the community will provide feedback.

Uzaaft:
After a second lookover, I feel like this code smells too much.

iamralch:
@Uzaaft @azdanov thank you for the feedback. I am closing the PR.

PR #1601, opened 2025-08-04
feat(pack): add nushell language pack

### LINKED ISSUE #1594: Add nushell pack

_No response_

### Describe the new feature

https://github.com/nushell/vscode-nushell-lang

### Additional context

https://www.nushell.sh/book/

Just a reminder to put here, there is a shell like bash and fish along with its configuration language (nuon: nu object notation).

cfbender:
Closes #1594 

## 📑 Description

Adds the nushell language pack

## ℹ Additional Information

Wasn't 100% sure if conform installs it's own `nufmt`, but I was pretty sure it doesn't - so I put it in the readme as needing a local install.

### DIFF
--- a/lua/astrocommunity/pack/nushell/README.md
+++ b/lua/astrocommunity/pack/nushell/README.md
@@ -2,6 +2,7 @@
 Requires:
 
 - A local [nufmt](https://github.com/nushell/nufmt) installation
+- A local [nushell](https://www.nushell.sh/) installation for the LSP
 
 This plugin pack does the following:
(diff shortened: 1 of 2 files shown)


azdanov:
You're correct that it needs to be manually installed. Mason doesn't seem to have nufmt, and conform doesn't install dependencies, only sets up formatting capabilities.

azdanov:
Also, would be good to mention that nushell itself is a requirement, since lsp won't work without it.

PR #1612, opened 2025-08-16
feat(file-explorer): add name to oil-nvim

alemagio:
## 📑 Description

Adding a name to the plugin makes it possible to customize it (like, for example, in the doc: "catppuccin")


Uzaaft:
This is not needed. You can customize it by using the full name: `stevearc/oil.nvim`

alemagio:
> This is not needed. You can customize it by using the full name: `stevearc/oil.nvim`

Oh boy
Sorry didn't know that 😅

PR #1628, opened 2025-08-25
feat(sql)!: Dont force ansi dialect

CongLuanTran:
## 📑 Description

The original configuration for `sqlfluff` add an extra `--dialect ansi` argument to the null_ls configuration, forcing the use of ANSI dialect. This is very annoying since it will override the dialect set in the configuration file `.sqlfluff` and leave little options for the user to use their desired dialect. With this change I remove that extra argument. However, this also means that the user must create a `.sqlfluff` file in the folder and specify the dialect there. I might adjust the README file to reflect this.

- [x] Remove `--dialect ansi` extra argument from `init.lua`
- [x] Adjust the README


Uzaaft:
Waiting on the readme changes.

CongLuanTran:
README added, but this will a breaking change since it requires a `.sqlfluff` config file. Without the config file, `sqlfluff` will not work.

Uzaaft (approved):

PR #1630, opened 2025-08-26
fix(hop-nvim): visual-selection related keymapping bug

### LINKED ISSUE #1629: Switch Hop.nvim visual mode keymaps from `v` to `x` to prevent snippet/selection conflicts
### Checklist

### Neovim version (nvim -v)

0.11.3 release

### Operating system/version

Arch Linux

### Terminal/GUI

kitty

### Describe the bug

The current `astrocommunity` hop.nvim configuration uses `v` mode mappings for visual selections.  

This has two major problems:

1. **Snippet placeholder conflicts**  
   Many snippet engines (e.g., LuaSnip, Ultisnips) temporarily place you in **characterwise visual mode** to select active placeholders.  
   With mappings in `v` mode, pressing `s` while editing a placeholder triggers Hop instead of replacing the selection as intended.  
   This breaks common workflows (e.g., typing `s` to change text in a selected placeholder) and forces you to hit `<Esc>` and re-enter the operation manually.

2. **Limited visual mode coverage**  
   `v` mode mappings apply only to characterwise visual mode, not to **visual block mode** (`Ctrl-v`).  
   As a result, block selections currently do not get the same Hop functionality.

---

## Additional note
The current configuration also passes an `extend_visual` option to `hop.nvim`, but this option does not exist in Hop and is silently ignored.  
This means there is no actual “extend selection” behavior happening; the parameter should be removed.

---

### Steps to Reproduce

1. Install the hop.nvim community pack
2. Use a snippet that has text place holder like `fromim` shortcut for python
3. You are in visual mode so try to type anything that start with 's' or 'S'
4. You will be able to see how it affects your workflow

### Expected behavior

Use `x` mode mappings instead of `v` mode.  
This applies mappings to both visual and visual block modes, and it avoids hijacking snippet placeholders because `x` mode is only active during visual selections explicitly initiated by the user.

Switching to `x` mode restores expected editing behavior, removes snippet conflicts, and broadens support to all visual modes.

### Screenshots

Below is a demonstration of this problem
https://github.com/user-attachments/assets/7b994f86-2b02-44a3-82c8-e09df636ae31

### Additional Context

_No response_

### Minimal configuration

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
end

-- bootstrap lazy
  -- stylua: ignore
end

-- install plugins
local plugins = {

}
  root = root .. "/plugins",
})

```

std-enigma:
Closes #1629

### DIFF
--- a/lua/astrocommunity/motion/hop-nvim/init.lua
+++ b/lua/astrocommunity/motion/hop-nvim/init.lua
@@ -9,7 +9,7 @@ return {
           ["s"] = { function() require("hop").hint_words() end, desc = "Hop hint words" },
           ["<S-s>"] = { function() require("hop").hint_lines() end, desc = "Hop hint lines" },
         },
-        v = {
+        x = {
           ["s"] = { function() require("hop").hint_words { extend_visual = true } end, desc = "Hop hint words" },
           ["<S-s>"] = {
             function() require("hop").hint_lines { extend_visual = true } end,


Uzaaft:
What issue does this solve? Why is it problematic to have this for visual and select mode

Uzaaft:
Ah this closes #1629.
Please use the template next time @std-enigma . Not all fields needs to be filled

PR #1635, opened 2025-08-28
refactor(main): Replace `echasnovski` with `nvim-mini`

Uzaaft:
@mehalter need a review from you.
Blocked on Astronvim/astronvim#2822


mehalter:
This shouldn't happen until AstroNvim v6

PR #1637, opened 2025-09-04
feat(vimtex): disable treesitter highlight to use vimtex highlight

CongLuanTran:
## 📑 Description

This PR disable `nvim-treesitter` highlight for `latex` to use `vimtex` highlight, which is needed for many vimtex features, including some text objects.


Uzaaft:
@CongLuanTran I moved the treesitter stuff into a dependency for the vimtex plugin. Could you test this and see if its still working as intended?

CongLuanTran:
> @CongLuanTran I moved the treesitter stuff into a dependency for the vimtex plugin. Could you test this and see if its still working as intended?

For my use cases (math environment an concealing), it works as intended.

Issue #1641, opened 2025-09-15
Venv-selector-nvim installing the wrong version


rodrigopmatias:
After the last PR on the Python pack, I realized the wrong version of venv-selector-nvim was being installed. This was noticeable because even after selecting the correct virtual environment, a new nvim session wouldn't automatically activate it.

I only noticed this issue because I forced a reinstallation of the plugins by deleting the cache and installation directories.

After extensive testing, I finally managed to fix the problem.

Uzaaft:
False report. I have been using the venv-selector code in my own config, works fine.

PR #1642, opened 2025-09-15
Hotfix/python venv selector branch fix

rodrigopmatias:
## 📑 Description

After the last Python pack update, I noticed on a secondary notebook that if the update wasn't done manually, a very old version of venv-selector-nvim was being installed.

I worked on solving this, and after recreating the environment multiple times, I realized the problem was fixed with this small change.

This resolves issue #1641.


lua/astrocommunity/pack/python/init.lua
```diff
@@ -88,8 +89,15 @@ return {
...
       },
     },
-    opts = {},
-    cmd = "VenvSelect",
+    opts = {
+      -- options = {
+      --   debug = true,
+      -- },
+    },
+    command = {
```
azdanov, line 97: You can check the docs for what options are available for lazy loading: https://lazy.folke.io/spec#spec-lazy-loading

azdanov: @rodrigopmatias Sorry for not being clear enough. But you removed the `cmd` that was used before and added `command` which doesn't exist. The `lazy` is not needed, since `ft` and `cmd` indicate already that it should be lazy.

lua/astrocommunity/pack/python/init.lua
```diff
@@ -88,8 +91,15 @@ return {
         },
       },
     },
-    opts = {},
-    cmd = "VenvSelect",
+    opts = {
+      -- options = {
```
azdanov, line 95: No commented out code should be in the plugin. If you wish, then can indicate in the readme some options or configuration.

azdanov:
PR title doesn't pass validation https://github.com/AstroNvim/astrocommunity/actions/runs/17752661693/job/50450099133?pr=1642

lua/astrocommunity/pack/python/init.lua
```diff
@@ -75,6 +75,9 @@ return {
   },
   {
     "linux-cultist/venv-selector.nvim",
+    branch = "main",
```
Uzaaft, line 78: This shouldnt be nessecary at all. I am using venv-selector without defining the branch.

Uzaaft:
To sum it up: This PR isnt needed. Feel free to open a new PR if you want to contribute just the ft changes. Make sure the code is properly written.

PR #1645, opened 2025-09-20
fix(copilot-lua-cmp): rebind dismiss keybinding to avoid keymap conflict

### LINKED ISSUE #1564: Copilot-lua-cmp: change C-c mapping

Now the completion.copilot-lua-cmp maps C-c to "dismiss", however, it is a common practice to use C-c to exit form insert mode. 

### Describe the new feature

Use another keymap for the copilot dismiss function,  e.g. C-q (for quit), or some other keymap.

### Additional context

_No response_

std-enigma:
Rebinds the dismiss keybinding to be <C-e> which is a common convention in vim/nvim to close a popup menu

Closes #1564


Uzaaft:
Isnt the dismiss more often `<C-e>` and not `<C-E>`

std-enigma:
Yes, it is — but the only reason I chose this was that the other bindings for this plugin were all in capital case, so I figured that’s what people were used to.

Uzaaft:
I think it should be lowercase e, to be inline with "defaults"

std-enigma:
Done

> I think it should be lowercase e, to be inline with "defaults"

PR #1646, opened 2025-09-21
feat(nix): add nixfmt formatter to null-ls sources

iamralch:
Include nixfmt in the list of sources for null-ls. This enables automatic formatting of Nix files within the Neovim environment.


Uzaaft:
The pack includes `nixd`, which provides formatting builtins. 
https://github.com/nix-community/nixd/blob/2c25600cb9c91bc06fe8676c044814dc30435274/nixd/docs/configuration.md?plain=1#L62

PR #1648, opened 2025-09-25
refactor(ansible): clean up ansible filetype detection

mehalter:
## 📑 Description

This just cleans up the ansible filetype detection I had written before to something a bit more manageable with less false positives.


Uzaaft (approved):
Absolutely chadtastic! :shipit:

PR #1658, opened 2025-10-04
feat(pack/typst): Add typstyle to format typst

simbafs:
## 📑 Description

Add typstyle to format typst source code.


simbafs:
What does "conform integration" mean?

Uzaaft:
you need to add support for the formatter in conform.

PR #1676, opened 2025-10-19
feat(colorscheme): add zenbones colorscheme

menisadi:
## 📑 Description

Added zenbones colorscheme collection

## 📖 Additional Information

In order to prevent further dependencies I set the compatibility mode to 1.

### DIFF
--- a/lua/astrocommunity/colorscheme/zenbones/init.lua
+++ b/lua/astrocommunity/colorscheme/zenbones/init.lua
@@ -1,5 +1,5 @@
 return {
   "zenbones-theme/zenbones.nvim",
+  dependencies = "rktjmp/lush.nvim",
   lazy = false,
-  config = function() vim.g.bones_compat = 1 end,
 }


lua/astrocommunity/colorscheme/zenbones/init.lua
```diff
@@ -0,0 +1,5 @@
+return {
+  "zenbones-theme/zenbones.nvim",
+  lazy = true,
```
Uzaaft:
```suggestion
  lazy = false,
```

menisadi:
Thanks @Uzaaft
I pulled in your suggested edits and they look good. 
I’ve synced my branch to include them. 
Let me know if anything else is needed.

Uzaaft:
There is one missing thing, which is to configure the vim.g option with astrocore(I think its possible, tbh not 100% sure ,it has been a while since I looked at the docs so I gotta refrsh my memory first).

menisadi:
@Uzaaft thanks!!
Tested. Looks good to me :)
Working as expected.

PR #1679, opened 2025-10-23
feat(trouble-nvim): integrate snacks.nvim picker

std-enigma:
## 📑 Description
Allow the integration of snacks.nvim with trouble.nvim


azdanov:
I don't think anyone can review this because it's not clear what it does.

std-enigma:
It allows the snacks picker to send the entries to trouble.nvim

Just like how with telescope you can send the items to quick-fix list or location list. This code allows you to send the items found by snacks pickers to the trouble.nvim list with all the added benefits of trouble.nvim

Uzaaft:
im taking a look at this sometime in the next 24h

PR #1680, opened 2025-10-29
fix(opencode-nvim): missing terminal provider

ALameLlama:
## 📑 Description

Currently getting this error, looks like we're missing some stuff from the default setup provided here https://github.com/NickvanDyke/opencode.nvim
```
No `provider.toggle` available — configure `vim.g.opencode_opts.provider`, or install `snacks.nvim` to use the default provider
```

### DIFF
--- a/lua/astrocommunity/ai/opencode-nvim/init.lua
+++ b/lua/astrocommunity/ai/opencode-nvim/init.lua
@@ -1,7 +1,10 @@
 return {
   "NickvanDyke/opencode.nvim",
   dependencies = {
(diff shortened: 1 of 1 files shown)


Uzaaft:
Doesnt the empty tables just override everything in snacks?

ALameLlama:
> Doesnt the empty tables just override everything in snacks?

do you think this is better?
```
 { "folke/snacks.nvim", opts = { input = { enabled = true }, picker = { enabled = true }, terminal = { enabled = true } } },
```

Uzaaft (approved):

PR #1684, opened 2025-11-05
Update mini.surround repo URL

auxym:
## 📑 Description

The repository for mini.surround has been moved from https://github.com/echasnovski/mini.surround to https://github.com/nvim-mini/mini.surround. This PR updates the URL and plugin name in the Readme and init.lua.

I also added a list of the default bindings which are set in the init.lua, as they differ from the default bindings of mini.surround (gz instead of s as prefix).


azdanov:
Duplicate of https://github.com/AstroNvim/astrocommunity/pull/1635

PR #1688, opened 2025-11-24
feat(python): Split python pack into seperate smaller packs

Uzaaft:
This allows an end-user to more easily "stack" the python packs based on the tooling they use in their d2d work. A big breaking change but with the direction python tooling is going, I think its the move.


Uzaaft:
Would appreciate some feedback @azdanov and @mehalter

Uzaaft:
Doing this would allow us to add pyrefly, ty etc with less headache.

azdanov:
Just a thought, but are nested `pack`s are supported? To have `pack.python.basedpyright`, `pack.python.isort`, etc.

mehalter:
My guess is probably. We could probably have them in a `python` folder and then the submodules would  work. Also we could have an `init.lua` file in that folder that would still allow users to do: `astrocommunity.pack.python` and get a basic base setup that could just import the submodules.

Uzaaft:
Actually `pack.python.{package}` would slap. I like that idea. I can refactor this to follow that pattern instead. 



> Also we could have an `init.lua` file in that folder that would still allow users to do: `astrocommunity.pack.python` and get a basic base setup that could just import the submodules.

Issue with this is that they might and probably will conflict.

mehalter:
It would allow people to get a "base" config or they can build their own with the pieces

Uzaaft:
I think I addressed the feedback lol @azdanov  @mehalter

mehalter:
What if we had `packs.python.base` which is what is currently `packs.python.init` and then the python init file just imports the base with basedpyright and some other relatively sane defaults. Then people who are simply import `packs.python` don't get breaking changes, but people who want to customize the pack can move to importing `packs.python.base` along with the specific ones that they want?

This would align with the other packs in the case what you can just import one line for a sane, default config to get up and running rather than pigeon holing everyone into having at least 3 lines to set up a single python environment.

EDIT: I opened a PR to give an example of what I mean (https://github.com/AstroNvim/astrocommunity/pull/1723) Feel free to re-implement it here or I can just move it into here. I haven't tested it, it's just there as an example.

Uzaaft:
> What if we had `packs.python.base` which is what is currently `packs.python.init` and then the python init file just imports the base with basedpyright and some other relatively sane defaults. Then people who are simply import `packs.python` don't get breaking changes, but people who want to customize the pack can move to importing `packs.python.base` along with the specific ones that they want?
> 
> This would align with the other packs in the case what you can just import one line for a sane, default config to get up and running rather than pigeon holing everyone into having at least 3 lines to set up a single python environment.
> 
> EDIT: I opened a PR to give an example of what I mean (#1723) Feel free to re-implement it here or I can just move it into here. I haven't tested it, it's just there as an example.

I like the design, just not sure if we should offer a base pack as I feel like its just way too opinionated and one can never win/do it right anyways 😅

mehalter:
I think you would be surprised how many people will be just fine with an opinionated pack. This gives the ability for both use cases which I think is good. I do think that 90% of people will still just use `astrocommunity.pack.python`

mehalter:
with this change, the number of lines needed to set up a specific python development environment doesn't change and the number of lines needed to set up a python development environment where you don't care about which tools decreases quite a bit with the added benefit that the people who already use the python pack who are in that camp won't need to make any changes. Then this actually doesn't become a breaking change. It's just a non-breaking feature.

Uzaaft:
> with this change, the number of lines needed to set up a specific python development environment doesn't change and the number of lines needed to set up a python development environment where you don't care about which tools decreases quite a bit with the added benefit that the people who already use the python pack who are in that camp won't need to make any changes. Then this actually doesn't become a breaking change. It's just a non-breaking feature.

in mehalter we trust!

mehalter:
Also I renamed the PR since it's no longer breaking

Uzaaft:
I'll take it for a spin tomorrow.

decko:
This is neat, folks! Any updates on it?
I was about to write a new `python-ty` pack, but I'll definitely wait for this to be merged first.

Uzaaft:
> This is neat, folks! Any updates on it? I was about to write a new `python-ty` pack, but I'll definitely wait for this to be merged first.

Sorry I got busy. I'll be doing python stuff at $DAYJOB today so I'll make sure to test it.

Uzaaft:
This worked well enough during my last days of using this at $DAYJOB. I'm just gonna let my brain brew on it a bit today before merging

mrijken:
What is needed to get this merged?

azdanov:
Uzaaft probably drowning in $job

Uzaaft:
> Uzaaft probably drowning in $job

drowning, I already drowned lol.

pgr0ss:
I opened this PR to fix the README: https://github.com/AstroNvim/astrocommunity/pull/1737

PR #1695, opened 2025-12-08
fix(rustowl): Rustowl build failure

lunaradev1:
Change lua/astrocommunity/editing-support/rustowl/init.lua build command to use cargo binstall or cargo install. Compared to running cd rustowl, which doesn't work at all.

## 📑 Description

This pull request should fix an issue with the rustowl plugin in `astrocommunity.editing-support.rustowl` spec. Originally it would try to run `cd rustowl && cargo install --path . --locked` in the build part, which fails at the first command with _(the error depends on your shell)_:

```text
zsh:cd no such file or directory: rustowl
```

This uses cargo directly by either running `cargo binstall` or `cargo install` if the `cargo-binstall` crate isn't installed.

### DIFF
--- a/lua/astrocommunity/editing-support/rustowl/README.md
+++ b/lua/astrocommunity/editing-support/rustowl/README.md
@@ -6,8 +6,6 @@ See docs for usage
 
 **Repository**: <https://github.com/cordx56/rustowl>
 
-## Note
-
-This plugin can take a while to build, and `Lazy.nvim` has a build timeout. It's recommended to **increase the build timeout**
-so this plugin can build.
+> [!NOTE]
+> This plugin can take a while to build, and `Lazy.nvim` has a build timeout. It's recommended to **increase the build timeout* so this plugin can build.
 


lunaradev1:
This is one of my first pull requests to this repo, so... if anything is wrong, please let me know.

lunaradev1:
Fixed the title, it was incorrect. My bad guys!

lunaradev1:
Gotcha, I'll get back with a change. Running `cargo install` directly may raise an issue with Lazy having a build timeout _(as when I ran it with like 16 CPU cores, it timed out and the process was killed, so the build failed)_, so I may have to **update the README to say as such.**

lunaradev1:
Attempting to find out how to increase the build timeout, stand by.

lunaradev1:
Found it, just gotta review and make sure it's right.

lunaradev1:
Done.

lunaradev1:
Forgot to use `--locked`, fixed that.

Uzaaft:
Would this work if the user doesnt have a nightly toolchain? Does cargo install respect the rust-toolchain file upstream?

lunaradev1:
It should, I'll test it now.

lunaradev1:
Yeah, it works off nightly. It's broken on nightly though, but that's probably their code not being fit for nightly.

lunaradev1:
[Here's a recording of me building it with Nightly Rust.](https://asciinema.org/a/9zntLTB8GiuoM28tA3J8tqpUc)

lunaradev1:
Probably not a big deal though, most people use stable, including me.

PR #1707, opened 2025-12-29
fix(pack): docker-language-server to dockerls

jay-babu:
Edit from @uzaaft:
No longer relevant after v6

### DIFF
--- a/lua/astrocommunity/pack/docker/init.lua
+++ b/lua/astrocommunity/pack/docker/init.lua
@@ -20,7 +20,7 @@ return {
     optional = true,
     opts = function(_, opts)
       opts.ensure_installed =
-        require("astrocore").list_insert_unique(opts.ensure_installed, { "docker_language_server" })
+        require("astrocore").list_insert_unique(opts.ensure_installed, { "dockerls" })
     end,
   },
   {
--- a/lua/astrocommunity/pack/docker/init.lua
+++ b/lua/astrocommunity/pack/docker/init.lua
@@ -19,8 +19,7 @@ return {
     "williamboman/mason-lspconfig.nvim",
     optional = true,
     opts = function(_, opts)
-      opts.ensure_installed =
-        require("astrocore").list_insert_unique(opts.ensure_installed, { "dockerls" })
+      opts.ensure_installed = require("astrocore").list_insert_unique(opts.ensure_installed, { "dockerls" })
     end,
   },
   {
--- a/lua/astrocommunity/pack/docker/init.lua
+++ b/lua/astrocommunity/pack/docker/init.lua
@@ -20,7 +20,7 @@ return {
     optional = true,
     opts = function(_, opts)
       opts.ensure_installed =
-        require("astrocore").list_insert_unique(opts.ensure_installed, { "docker-language-server" })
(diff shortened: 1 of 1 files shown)


jay-babu:
```lua
  {
    opts = function(_, opts)
      opts.ensure_installed = opts.ensure_installed or {}
      -- filter out the kotlin_language_server if it is already installed
      opts.ensure_installed = vim.tbl_filter(
        function(server) return server ~= "kotlin_language_server" end,
        opts.ensure_installed
      )
      opts.ensure_installed = vim.tbl_filter(function(server) return server ~= "sqls" end, opts.ensure_installed)
      table.insert(opts.ensure_installed, "tsp_server")
    end,
    config = function(_, opts)
      require("mason-lspconfig").setup(opts)
    end,
  },
```

I really don't know why but my mason-lspconfig setup function is not running till I manually added the config function. This is likely why no one has seen this error for 7 months as this is when this was committed. As soon as I did it started running again.

wert2all:
> I really don't know why but my mason-lspconfig setup function is not running till I manually added the config function. This is likely why no one has seen this error for 7 months as this is when this was committed. As soon as I did it started running again.

Maybe I'm doing wrong, but on a clean install of nvim,  I got `docker-compose-language-service` and `dockerfile-language-server` LSPs. Steps for reproduce:

1. Change `docker-language-server` to `dockerls`
2. ` rm -rf ~/.local/share/nvim && rm -rf ~/.local/state/nvim && rm -rf ~/.cache/nvim`
3. start nvim
4. `docker-compose-language-service` and `dockerfile-language-server` installed.

On my clean config, without you fix, I got `docker-language-server` only. Maybe `docker-language-server` installs other pack, which I am using.

azdanov:
I think it's my fault here.

Since astronvim v5 uses mason-lspconfig v1, which doesn't have `docker-language-server`.
v1: https://github.com/mason-org/mason-lspconfig.nvim/blob/1a31f824b9cd5bc6f342fc29e9a53b60d74af245/lua/mason-lspconfig/mappings/server.lua#L48 and https://github.com/mason-org/mason-lspconfig.nvim/blob/1a31f824b9cd5bc6f342fc29e9a53b60d74af245/lua/mason-lspconfig/mappings/filetype.lua#L48

But v2 has https://github.com/mason-org/mason-lspconfig.nvim/blob/3d89e7c92fbd96c5e10e0298fc2b006f21cf9428/lua/mason-lspconfig/filetype_mappings.lua#L50

So, need to probably hotpatch mason-lspconfig to recognize the new `docker-language-server` lsp to make it work.

There is some confusion here in the thread between `docker-language-server`, `dockerfile-language-server` and `docker-compose-language-service` which are 3 different LSPs. The previous update that I did was to replace the community LSPs (dockerfile and compose) with official https://github.com/docker/docker-language-server.

azdanov:
Here's an example of hotpatching (thanks to mehalter for providing this technique): https://github.com/AstroNvim/astrocommunity/blob/9ee60b5058cfc77d4f0ff6d4a75234a2f4998ee0/lua/astrocommunity/pack/oxlint/init.lua#L23-L42

Uzaaft:
Sorry for the force pushes. I was too deep until I realized that this isnt needed for v6.

PR #1709, opened 2025-12-29
fix(full-dadbod): add mysql and plsql filetypes for completion

taiwbi:
Dadbod completion only worked for `sql` filetype, but when opening queries in dadbod-ui for MySQL or PL/SQL databases, the buffer filetype is set to `mysql` or `plsql`. This caused completion to not work in those buffers.

This makes the dadbod completion configuration to include `mysql` and `plsql` filetypes


rami3l:
@taiwbi Just being curious, I don't use `.mysql` or `.plsql` extensions myself, but since you are on `blink.cmp` too, didn't you hit the issue described in https://github.com/AstroNvim/astrocommunity/pull/1712 as well and, maybe there is an interaction between those issues?

taiwbi:
@rami3l Yes, You're right. I haven't noticed it.

PR #1713, opened 2025-12-31
fix(vector-code-nvim): install or upgrade `vectorcode-cli` based on cli existence

cristobalgvera:
## 📑 Description

This pull request updates the build logic for the `VectorCode` plugin to improve how dependencies are checked and managed. The main change is that it now conditionally installs or upgrades the `vectorcode` tool based on its presence, and uses `vim.system` for better process handling.

Dependency management improvements:

* The build function now checks if `uv` is installed by verifying that `vim.fn.executable "uv"` returns 0, and throws an error if not.
* The script now determines whether to install or upgrade `vectorcode` by checking if it is already executable, and then calls `uv tool install vectorcode` or `uv tool upgrade vectorcode` accordingly using `vim.system` for process execution.

### DIFF
--- a/lua/astrocommunity/editing-support/vector-code-nvim/init.lua
+++ b/lua/astrocommunity/editing-support/vector-code-nvim/init.lua
@@ -1,20 +1,24 @@
 return {
   "Davidyz/VectorCode",
   build = function()
-    local EXECUTABLE_EXISTS = 1
-
-    if vim.fn.executable "uv" == EXECUTABLE_EXISTS then
-      local action = vim.fn.executable "vectorcode" == 0 and "install" or "upgrade"
-      vim.system { "uv", "tool", action, "vectorcode" }
-      return
+    if vim.fn.executable "uv" == 1 then
+      assert(
+        vim.system({
+          "uv",
+          "tool",
+          vim.fn.executable "vectorcode" == 1 and "upgrade" or "install",
+          "vectorcode",
+        }):wait().code == 0,
+        "Failed to install or upgrade vectorcode with uv"
+      )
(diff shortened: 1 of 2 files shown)


lua/astrocommunity/editing-support/vector-code-nvim/init.lua
```diff
@@ -1,8 +1,10 @@
 return {
   "Davidyz/VectorCode",
   build = function()
-    if not vim.fn.executable "uv" then error "The VectorCode pack requires uv installed" end
-    vim.cmd "uv tool install vectorcode"
+    if vim.fn.executable "uv" == 0 then error "The VectorCode pack requires `uv` installed" end
+
+    local action = vim.fn.executable "vectorcode" == 0 and "install" or "upgrade"
+    vim.system { "uv", "tool", action, "vectorcode" }
```
Uzaaft: we should support uv and pipx.
If uv is installed, use uv no matter what. fallback to pipx, and none of them are there, notify the end-user. Also update the docs to reflect this.

Uzaaft (changes requested):
Sorry for the delay in reviewing this.

lua/astrocommunity/editing-support/vector-code-nvim/init.lua
cristobalgvera: Done!

Thanks for the advice.

Uzaaft (changes requested):
1. There is a typo in the url.
2. The variable is unnessecary.
I'll fix this.

Uzaaft (approved):

Issue #1698, opened 2025-12-14
add tsgo support


jay-babu:
### Is your feature related to a problem?

https://github.com/microsoft/typescript-go

https://github.com/sohanemon/tsgo.nvim - tsc.nvim replacement

### Describe the new feature

tsgo has been available for awhile and has its own lsp. it has its own lsp and is much faster than the other options out there like tsserver, vtsls, etc. The ask is to add support for this. Unfortunately, I tried a lot setting this up myself but likely am doing something wrong hence wasn't able to do so.

Uzaaft:
I've been working on this, this wont land until lspconfig in astronvim updates lspconfig.

Uzaaft:
@jay-babu are you using vtsls? 
Also that tsgo.nvim plugin wont be added, its stupid. moving to tsgo in tsc.nvim is fairly easy and what the creator of `tsc.nvim` did.

jay-babu:
@Uzaaft yes currently I am using vtsls. I disabled it when trying to set this up. Yes, wasn't saying use tsgo.nvim, was giving it as a reference

Uzaaft:
@jay-babu  You can pass this to vtsls:
```
settings = { 
typescript = {
exerpimental = {
        useTsgo = true,
}
}
}
```

jay-babu:
Yes, I tried this. Did not work for me

Uzaaft:
> Yes, I tried this. Did not work for me
> […](#)
> 
> On Jan 8, 2026 at 12:52 PM -0500, Uzair Aftab ***@***.***>, wrote:
>  Uzaaft left a comment ([AstroNvim/astrocommunity#1698](https://github.com/AstroNvim/astrocommunity/issues/1698))
>  [@jay-babu](https://github.com/jay-babu) You can pass this to vtsls:
>  settings = {
>  typescript = {
>  exerpimental = {
>         useTsgo = true,
>  }
>  }
>  }
>  —
>  Reply to this email directly, view it on GitHub, or unsubscribe.
>  You are receiving this because you were mentioned.Message ID: ***@***.***>

Make sure your vtsls is updated btw. Its working perfectly fine at my end. :)

PR #1730, opened 2026-01-27
feat!: astronvim v6 compatible 

Uzaaft:
Update astrocommunity in preparation for the v6 release of astrocommunity.

## Closes
Clsoes #1698

### DIFF
--- a/lua/astrocommunity/note-taking/zk-nvim/init.lua
+++ b/lua/astrocommunity/note-taking/zk-nvim/init.lua
@@ -34,7 +34,7 @@ return {
     },
   },
   opts = function()
-    local lsp_config = require("astrolsp").lsp_opts "zk"
+    local lsp_config = vim.lsp.config["zk"] or {}
     return {
       lsp = {
         config = lsp_config,
--- a/lua/astrocommunity/note-taking/zk-nvim/init.lua
+++ b/lua/astrocommunity/note-taking/zk-nvim/init.lua
@@ -19,7 +19,7 @@ return {
   dependencies = { "AstroNvim/astrolsp", opts = { handlers = { zk = function() end } } },
   specs = {
     {
-      "williamboman/mason-lspconfig.nvim",
+      "mason-org/mason-lspconfig.nvim",
       optional = true,
       opts = function(_, opts)
         opts.ensure_installed = require("astrocore").list_insert_unique(opts.ensure_installed, { "zk" })
--- a/lua/astrocommunity/pack/cs/init.lua
+++ b/lua/astrocommunity/pack/cs/init.lua
@@ -28,25 +28,14 @@ return {
     dependencies = {
       {
         "AstroNvim/astrolsp",
-        opts = vim.fn.has "nvim-0.11" == 1
-            and {
-              handlers = {
-                csharp_ls = function(server, opts)
-                  require("lspconfig")[server].setup(opts)
-                  require("csharpls_extended").buf_read_cmd_bind()
-                end,
-              },
-            }
-          or { -- TODO: drop when dropping support for Neovim v0.10
-            config = {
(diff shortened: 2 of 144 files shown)


lua/astrocommunity/pack/rust/init.lua
```diff
@@ -80,7 +80,7 @@ return {
   },
   {
     "mrcjkb/rustaceanvim",
-    version = vim.fn.has "nvim-0.11" == 1 and "^6" or "^5",
+    version = "^6",
```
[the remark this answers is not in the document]
rami3l: @Uzaaft That's a bit surprising to me. May I ask why (and what would be your path forward)?

lua/astrocommunity/note-taking/zk-nvim/init.lua
```diff
@@ -34,7 +34,7 @@ return {
     },
   },
   opts = function()
-    local lsp_config = require("astrolsp").lsp_opts "zk"
+    local lsp_config = vim.lsp.config["zk"] or {}
```
mehalter: This doesn't need the `or {}` because indexing into `vim.lsp.config` will always return a table no matter what

lua/astrocommunity/pack/rust/init.lua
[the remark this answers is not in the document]
rami3l: @Uzaaft I'm not super sure about that... I remember having contributed to both projects and rustaceanvim's code look much more maintainable than the other you've mentioned, but I do remember that rustaceanvim tries to bypass some commonly-used config mechanisms to make way for some particular features, so indeed it's harder to hack it to work in astronvim 🤔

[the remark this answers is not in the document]
Uzaaft: Thats why its my opinion 😂 Cuz it has never worked for me.

lua/astrocommunity/pack/cs/init.lua
```diff
@@ -28,25 +28,14 @@ return {
...
-                handlers = {
-                  ["textDocument/definition"] = function(...) require("csharpls_extended").handler(...) end,
-                  ["textDocument/typeDefinition"] = function(...) require("csharpls_extended").handler(...) end,
-                },
-              },
-            },
+        opts = {
+          handlers = {
+            csharp_ls = function(server, opts)
+              require("lspconfig")[server].setup(opts)
```
rami3l: Just trying it out and got a ~~deprecation warning~~ runtime error.

This should really be `vim.lsp.enable(server)` as `opts` is already in `vim.lsp.config`, or do we need to set it up here?

lua/astrocommunity/pack/rust/init.lua
```diff
@@ -113,8 +113,7 @@ return {
         adapter = cfg.get_codelldb_adapter()
       end
 
-      local astrolsp_avail, astrolsp = pcall(require, "astrolsp")
-      local astrolsp_opts = (astrolsp_avail and astrolsp.lsp_opts "rust_analyzer") or {}
+      local astrolsp_opts = vim.lsp.config["rust_analyzer"] or {}
```
Mic92, line 115: Does this replace?

https://github.com/AstroNvim/astrocommunity/pull/1747

rami3l: @Mic92 No it doesn't, the original implementation meant to extract everything from the `rust_analyzer` config. The `root_dir` callback incompatibility is related but essentially something else. They are meant to be working together though.

Uzaaft:
r? @mehalter

I've lost track/overview of the breaking changes in v6 FYI 😂

mehalter (approved):
made some fixes. Looks good to me. We can let the community debug any issues. But we should get this out.

PR #1751, opened 2026-04-07
fix(trouble-nvim): fix crash when lspkind is unaccessible

jmxyyy:
## 📑 Description

Trouble's LSP-related commands crash when lspkind.nvim is unavailable, because kinds is passed a boolean while a table is expected.

### DIFF
--- a/lua/astrocommunity/diagnostics/trouble-nvim/init.lua
+++ b/lua/astrocommunity/diagnostics/trouble-nvim/init.lua
@@ -37,7 +37,7 @@ return {
         },
         folder_closed = get_icon "FolderClosed",
         folder_open = get_icon "FolderOpen",
-        kinds = lspkind_avail and lspkind.symbol_map,
+        kinds = lspkind_avail and lspkind.symbol_map or nil,
       },
     }
   end,


Uzaaft:
Wait Im not following whats happening here.
if lspkind_avail is unaccessible and isnt enabled, how are we passing a boolean?

jmxyyy:
@Uzaaft  apologize if my previous description was unclear, as English is not my native language.

```lua
:lua local lspkind_avail, lspkind = pcall(require, "lspkind"); local kinds = lspkind_avail and lspkind.symbol_map; vim.print(kinds)
```

When the lspkind plugin is not installed, the expression `kinds = lspkind_avail and lspkind.symbol_map` evaluates to the boolean value false. This false value is then passed into trouble/format.lua at line 147:

```Lua
local icon = ctx.opts.icons.kinds[ctx.item.kind]
```

Here, a table is expected, so passing a boolean causes a crash.

This issue can be easily reproduced with a minimal configuration:

     { import = "astrocommunity.diagnostics.trouble-nvim" },

    Execute the command: :Trouble lsp_document_symbols

Uzaaft (approved):

PR #1755, opened 2026-04-11
feat(rust): bump `rustaceanvim` to v8+

rami3l:
## 📑 Description

Based on #1753 and #1754, as promised :)

This PR bumps `rustaceanvim` to v9 on Neovim v0.12+, and to v8 on Neovim v0.11.

Tested locally with first [this](https://github.com/rami3l/nvim-config/blob/8c46ecc06b6d28cd354941260f26fbefedca1f45/lua/plugins/lang/rust.lua#L9-L29) override (daily-driving) and then a local `dir` override on `astrocommunity`.

## 📖 Additional Information

One major difference from the current `v6` here is that the sharing of project-local configs with VSCode has been outsourced to an external plugin `codesettings.nvim`.

Considering the fact that it might not be relevant to everyone I'm leaving the config under a `pcall` guard.

However, it might still be interesting to include it in the pack here (as an optional dependency or something?), or it can be made into its own recipe later on.

cc @Uzaaft

### DIFF
--- a/lua/astrocommunity/pack/rust/init.lua
+++ b/lua/astrocommunity/pack/rust/init.lua
@@ -79,7 +79,7 @@ return {
   },
   {
     "mrcjkb/rustaceanvim",
-    version = "^6",
+    version = vim.fn.has "nvim-0.12" == 1 and "^9" or "^8",
     ft = "rust",
     specs = {
       {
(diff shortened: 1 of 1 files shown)


Uzaaft (approved):

rami3l:
@Uzaaft Thanks for the review! I think we can merge this now?

Uzaaft:
@rami3l Sorry github ui sucks and it was stuck waiting on the CI to pass so I moved back to doing actual work lol.

PR #1756, opened 2026-04-21
fix(refactoring-nvim): fix dependencies and API usage for latest version

rami3l:
## 📑 Description

The last update of this bundle is from more than 1 year ago, and both the dependencies and the APIs have changed quite a bit since then.

## 📖 Additional Information

See https://github.com/ThePrimeagen/refactoring.nvim#configuration for the example keymapping config.

Personally I would actually prefer having the users configure the keymap themselves for this plugin with rather rich functionalities... @Uzaaft what do you think?


Uzaaft:
I just looked at the rewrite. 
Holy fuck is that a sloppy rewrite. 

I think we should keep the keymaps in here. astrocommunity is meant to cover both power users and people who just want a OOTB experience. 

@rami3l

rami3l:
> I think we should keep the keymaps in here. astrocommunity is meant to cover both power users and people who just want a OOTB experience. 

@Uzaaft Sure!

> I just looked at the rewrite. 
> 
> is that a sloppy rewrite. 

You mean the recent rewrite of the refactoring plugin in question? Yup I was quite surprised when it requires `expr = true` among other stuff :|

Uzaaft (approved):

PR #1757, opened 2026-04-22
fix(refactoring-nvim): add async.nvim dependency

iamralch:
This PR adds [lewis6991/async.nvim](http://github.com/lewis6991/async.nvim) as a dependency for refactoring.nvim.

According to the official installation [instructions](https://github.com/ThePrimeagen/refactoring.nvim#installation), async.nvim is required for proper async functionality, but it was not previously included in the AstroCommunity plugin spec. This could lead to runtime issues or incomplete behavior when using refactoring features.


iamralch:
Closing due to #1756

PR #1759, opened 2026-04-26
perf(trouble-nvim): avoid eagerly loading trouble when configuring snacks integration

ALVAROPING1:
## 📑 Description

The previous method eagerly required trouble while configuring snacks to get trouble's picker actions. New method only requires it when indexing the actions table. Sadly snacks still loads trouble the first time any picker is opened rather than when the action is first triggered, but it's better than loading it at startup

### DIFF
--- a/lua/astrocommunity/diagnostics/trouble-nvim/init.lua
+++ b/lua/astrocommunity/diagnostics/trouble-nvim/init.lua
@@ -57,7 +57,13 @@ return {
       opts = function(_, opts)
         return vim.tbl_deep_extend("force", opts or {}, {
           picker = {
-            actions = require("trouble.sources.snacks").actions,
+            actions = setmetatable({}, {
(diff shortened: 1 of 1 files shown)


Uzaaft:
ngl this looks hedious as fuck to me and not something I'm particularly eager to maintain when its not clear what it does on a first look. 

wdyt @azdanov

azdanov:
Hey. I don't think this is a part of a "community" plugin. If something breaks or becomes incompatible in the future then no one will be able to understand what was the issue. It's more of a personal dotfile thing.

Uzaaft:
Gonna agree with @azdanov. Unless you come up with a version of this that's easier to maintain. This is not good enough as is.

ALVAROPING1:
This is the simplest way i could come up with, so i will keep it in my personal config 👍

PR #1761, opened 2026-04-27
fix(typst)!: remove the outdated .vim plugin; configure treesitter

monoid-kitty:
the outdated plugin (kaarmu/typst.vim) bricks some formatting provided by the LSP (tinymist)

## 📑 Description

Not only is there no need to use workaround code highlighting provided by the said plugin (there is Treesitter now!) and other its features (because a proper LSP does most of it), but also the old plugin somehow broke formatting provided by the LSP for me.

### DIFF
--- a/lua/astrocommunity/pack/typst/init.lua
+++ b/lua/astrocommunity/pack/typst/init.lua
@@ -1,5 +1,12 @@
 return {
-  { "kaarmu/typst.vim", ft = { "typst" } },
+  {
+    "AstroNvim/astrocore",
+    optional = true,
+    ---@type AstroCoreOpts
+    opts = {
+      treesitter = { ensure_installed = { "typst" } },
+    },
+  },
   {
     "mason-org/mason-lspconfig.nvim",
     opts = function(_, opts)


monoid-kitty:
if anyone want to see what behavior it fixes:

# repro.lua
```lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
for _, name in ipairs { "config", "data", "state", "runtime", "cache" } do
end

-- bootstrap lazy
  -- stylua: ignore
end

-- install plugins
local plugins = {
  { "AstroNvim/AstroNvim", version = "^5", import = "astronvim.plugins" },
  { "AstroNvim/astrocommunity", version = "^20", { import = "astrocommunity.pack.typst" } },

}
  root = root .. "/plugins",
})

```

# how to reproduce
1. `nvim -u repro.lua main.typ`
2. type `#let project() = {` (the closing bracket will be auto-completed, and press enter, so that it looks like this:

```rust
#let project() = {
         // cursor here [9 space chars]
}
```

# expected
by default (and I've tried to set this manually) the LSP (tinymist) uses 2 spaces for indending. but neovim forces me to use 9

after the said changes it becomes:
```rust
#let project() = {
  // cursor here [2 space chars]
}
```

lua/astrocommunity/pack/typst/init.lua
```diff
@@ -1,5 +1,12 @@
 return {
-  { "kaarmu/typst.vim", ft = { "typst" } },
+  {
+    "AstroNvim/astrocore",
+    optional = true,
```
azdanov, line 4:
```suggestion
```
Not needed since I assume astrocore would be installed.

azdanov:
I made it a breaking change since that plugin did offer some [extra functionality](https://github.com/kaarmu/typst.vim#features) that might "break" people's workflow.

lua/astrocommunity/pack/typst/init.lua
monoid-kitty: I honestly just looked up how other community plugins do it, and copied it.
See [community.pack.cpp](https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/pack/cpp/init.lua#L20), [community.pack.go](https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/pack/go/init.lua#L60), [community.pack.lua](https://github.com/AstroNvim/astrocommunity/blob/main/lua/astrocommunity/pack/lua/init.lua#L28) etc

azdanov: Oh well. It's just not needed.

azdanov (approved):

PR #1762, opened 2026-04-28
fix(dev-tools-nvim): capture new dependency on async

hakamadare:
## 📑 Description

Per ThePrimeagen/refactoring.nvim#519, which announces a major rewrite,
`plenary.nvim` is no longer a dependency (which is good because that
library is apparently no longer maintained?), but now there's a
dependency on lewis6991/async.nvim.

## 📖 Additional Information

The community recipe in `editing-support` has captured this update, but
since I had configured this recipe but not the `editing-support` recipe,
I got a storm of tracebacks complaining about a missing `async()`
method.


hakamadare:
trying again with updated PR title, it's not clear to me how to make CI happy

PR #1764, opened 2026-05-09
docs: remove old nvim version requirements

ALameLlama:
## 📑 Description

I noticed that a few plugins still have nvim version that we either don't support or are now the minimum required for AstroNvim v6 so I think it's save to remove them


Uzaaft (approved):
`-14` diff me likey :shipit:

Uzaaft:
@mehalter need a senpai merge on this.

mehalter (approved):
:shipit:

PR #1765, opened 2026-05-25
feat(just):  Added autocmd for replacing make with just, if find justfile in workdir

TakiMoysha:
## 📑 Description

Vim has `make+quickfix`, which is very convenient. Currently, `just` can only be called via `:!just` and doesn't pass the output to quickfix.

I've added an autocmd that replaces `make` calls in `make` with `just` if it finds a `justfile`.

- [x] added autodmd `just_as_make`;
- [x] updated readme for pack/just;

## 📖 Additional Information

Maybe it's worth using `VimEnter` event instead, but I think `BufReadPost` works better.


Uzaaft:
This doesn't make sense. 
I'd be open for a `:just` command(if its not provided by the plugins already). Closing as is.

PR #1770, opened 2026-06-12
docs(python): clarify base pack import

### LINKED ISSUE #1748: [pack/python] README example fix
With the recent split of python pack, the README gives an example to import only `pack.python` for base pack, but this is incorrect and doing so willl import `pack/python/init.lua` which adds all the subpacks. The correct way to import base pack should be `pack.python.base` to import `pack/python/base.lua` only. Same with the other examples.

Uzaaft:
Closes #1748.

Clarifies that the default Python pack imports the base pack plus selected tooling subpacks, and that users should import `astrocommunity.pack.python.base` directly for base-only functionality.


Uzaaft:
R? @azdanov

azdanov (approved):
Coolio

PR #1790, opened 2026-07-25
fix(mcphub-nvim): disable make_vars to prevent codecompanion crash

### LINKED ISSUE #1777: mcphub-nvim: make_vars causes error in codecompanion due to load order
### Checklist

### Neovim version (nvim -v)

NVIM v0.12.2

### Operating system/version

Linux

### Terminal/GUI

bash

### Describe the bug

When using the mcphub.nvim plugin from astrocommunity together with codecompanion.nvim, a "bad argument #1 to 'pairs' (table expected, got nil)" error occurs when starting Neovim. The error occurs when the mcphub extension attempts to register variables (make_vars = true) with codecompanion.

### Steps to Reproduce

1. Install the mcphub.nvim and codecompanion.nvim plugins via astrocommunity.
2. Restart NeoVim
3. open any file
4. see error in bottom panel

### Expected behavior

vim.schedule callback: ...b.nvim/lua/mcphub/extensions/codecompanion/variables.lua:20: bad argument #1 to 'pairs' (table expected, got nil)                                                                                                   
stack traceback:                                                                                                                                                                                                                              
        [C]: in function 'pairs'                                                                                                                                                                                                              
        ...b.nvim/lua/mcphub/extensions/codecompanion/variables.lua:20: in function 'register'                                                                                                                                                
        ...b.nvim/lua/mcphub/extensions/codecompanion/variables.lua:92: in function ''                                                                                                                                                        
        vim/_core/editor.lua: in function <vim/_core/editor.lua:0>    

### Screenshots

_No response_

### Additional Context

_No response_

### Minimal configuration

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
end

-- bootstrap lazy
  -- stylua: ignore
end

-- install plugins
local plugins = {

}
  root = root .. "/plugins",
})

```

syltharion:
Closes #1777

Sets `make_vars = false` in the mcphub codecompanion extension config. When mcphub loads before codecompanion is fully initialized, `make_vars = true` triggers variable registration that crashes with 'bad argument #1 to pairs (table expected, got nil)'.

### DIFF
--- a/lua/astrocommunity/editing-support/mcphub-nvim/init.lua
+++ b/lua/astrocommunity/editing-support/mcphub-nvim/init.lua
@@ -16,7 +16,7 @@ return {
             callback = "mcphub.extensions.codecompanion",
             opts = {
               show_result_in_chat = true,
-              make_vars = true,
+              make_vars = false,
               make_slash_commands = true,
             },
           },


Uzaaft:
This seems like a hack.
Instead you should make one load the other if defined or something... Disabling isn't the way.

PR #1792, opened 2026-07-28
fix(opencode-nvim): Update api call for toggle function in opencode

josh-dix:
## 📑 Description

Due to opencode plugin api being modified this requires updating.

https://github.com/nickjvandyke/opencode.nvim/blob/7749a034db61258ece828df70a89ff31bb27ff47/CHANGELOG.md?plain=1#L70


lua/astrocommunity/ai/opencode-nvim/init.lua
```diff
@@ -15,8 +15,8 @@ return {
         local prefix = "<Leader>O"
         maps.n[prefix] = { desc = require("astroui").get_icon("OpenCode", 1, true) .. "OpenCode" }
         maps.n[prefix .. "t"] = {
-          function() require("opencode").toggle() end,
-          desc = "Toggle embedded",
+          function() require("opencode").start() end,
+          desc = "Start & Toggle embedded",
```
Uzaaft, line 19: Shouldn't this be just Start embedded?

josh-dix: Calling start again toggles it, so I thought most people would understand the toggle word despite the api call being start, and including both would be more clear to users.

I don't mind what it is called really but I do think it is better to include both. If I had to choose one I would stick with toggle since that is the user-friendly word for the more important aspect of the action.

Uzaaft: @josh-dix the start() command was removed fro the public api before your PR was opened: https://github.com/nickjvandyke/opencode.nvim/commit/49b34182ad6ba870bf19173b64ea0662b2fddc00

